Issue #95, opened 2014-11-13
Crash in RequestSerializer


danielrhammond:
Experiencing the same crash as observed in #59 in version 2.0.12 (Via cocoapods)

````
SIGSEGV SEGV_ACCERR
Thread: Unknown Name (Crashed)
0  libobjc.A.dylib  0x0000000193f400b4 objc_retain + 20
1  [App Name] 0x000000010023f968 -[AWSXMLRequestSerializer serializeRequest:headers:parameters:] (AWSURLRequestSerialization.m:181)
2  [App Name] 0x000000010024657c __41-[AWSURLSessionManager taskWithDelegate:]_block_invoke164 (AWSURLSessionManager.m:204)
3  [App Name] 0x0000000100256370 __48-[BFTask continueWithExecutor:withSuccessBlock:]_block_invoke (BFTask.m:335)
4  [App Name] 0x0000000100255dc8 __41-[BFTask continueWithExecutor:withBlock:]_block_invoke_2 (BFTask.m:287)
5  [App Name] 0x0000000100253af4 __29+[BFExecutor defaultExecutor]_block_invoke_2 (BFExecutor.m:43)
6  [App Name] 0x0000000100253fdc -[BFExecutor execute:] (BFExecutor.m:107)
7  [App Name] 0x0000000100255d80 __41-[BFTask continueWithExecutor:withBlock:]_block_invoke (BFTask.m:284)
8  [App Name] 0x0000000100255bf8 -[BFTask continueWithExecutor:withBlock:] (BFTask.m:319)
9  [App Name] 0x0000000100256268 -[BFTask continueWithExecutor:withSuccessBlock:] (BFTask.m:331)
10 [App Name] 0x00000001002563fc -[BFTask continueWithSuccessBlock:] (BFTask.m:341)
11 [App Name] 0x0000000100245eb8 -[AWSURLSessionManager taskWithDelegate:] (AWSURLSessionManager.m:150)
12 [App Name] 0x0000000100248550 __61-[AWSURLSessionManager URLSession:task:didCompleteWithError:]_block_invoke (AWSURLSessionManager.m:403)
13 [App Name] 0x0000000100256370 __48-[BFTask continueWithExecutor:withSuccessBlock:]_block_invoke (BFTask.m:335)
14 [App Name] 0x0000000100255dc8 __41-[BFTask continueWithExecutor:withBlock:]_block_invoke_2 (BFTask.m:287)
[...]

PR #98, opened 2014-11-21
Fix crash in AWSXMLRequestSerializer

danielrhammond:
On certain requests (retries for S3 uploads) in `constructURIandHeadersAndBody:rules:uriSchema:error` if `checkResourceIsReachableAndReturnError:` returns an error it will crash

This is a fix for https://github.com/aws/aws-sdk-ios/issues/95 and can be verified with this [sample project](https://github.com/danielrhammond/AWSSDK-Crash-Sample-Project)

### DIFF
--- a/AWSCore/Serialization/AWSURLRequestSerialization.m
+++ b/AWSCore/Serialization/AWSURLRequestSerialization.m
@@ -229,6 +229,7 @@
 
     rules = rules[@"members"] ? rules[@"members"] : @{};
 
+    __block NSError *serializerError;
     __block NSString *rawURI = uriSchema;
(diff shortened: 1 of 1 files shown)


jodyhagins:
You should explicitly initialize serializerError...

    __block NSError *serializerError = nil;

Then later, make sure the caller didn't pass NULL as the error pointer...

    if (serializerError) {
        if (error) {
            *error = serializerError;
        }
        return NO;
    }

danielrhammond:
No longer necessary as #95 has been resolved in 2.0.14

yosuke-matsuda:
Thank you for the pull request. It helped us identify and fix the issue.

PR #100, opened 2014-11-25
Implementation of AWSSTSCredentialsProvider

ilkerdagli:
I am working on a project that server-side requests STS credentials from AWS, and send this credentials
to clients (Token Vending Machine). I couldn't find any credentials provider which allows to use the credentials that came from my token vending machine(accessKey, secretKey, sessionKey, expiration). So, I have implemented AWSSTSCredentialsProvider class.


aidanlister:
@ilkerdagli did you get this working with 2.4?

Issue #108, opened 2014-12-06
Govcloud region support


drmarkpowell:
I can't use AWS SDK for iOS V2 because the AWSServiceConfiguration doesn't accept a Govcloud region type. This is a showstopper for my application since operating in Govcloud is a requirement. If I have missed something, please advise how I can use the V2 API with its AWSServiceConfiguration, AWSS3TransferManager, and DynamoDBObjectMapper in the Govcloud region.

drmarkpowell:
BTW I just integrated SDK v2.2.7 with my GovCloud ios app and it's working great!

PR #110, opened 2014-12-07
adding support for Govcloud region

drmarkpowell:
This is in regard to issue #108 which I filed. This path adds support for the Govcloud region as an endpoint for AWS services.


yosuke-matsuda:
We've released the AWS Mobile SDK for iOS 2.2.3 with the AWS GovCloud (US) Region support. Thank you.

PR #181, opened 2015-04-21
Fixed Complier Warning: "Auto property synthesis will not synthesize pro...

sean-howard:
When using Xcode 6.3 AWSS3TransferManager.h generates a compiler warning as follows:

`Auto property synthesis will not synthesize property 'body'; it will be implemented by its superclass, use @dynamic to acknowledge intention`

I've made a pull request which seems to take care of the issue by adding the specified @dynamic call to the @implementation.

Thanks, Sean.


yosuke-matsuda:
Thanks. We'll address it with the next release.

jpike88:
When is the next released due to come out?

PR #186, opened 2015-05-02
Support KMS server-side encryption.

bonebox:
Add additional enum for AWSServerSideEncryption to enable KMS server-side encryption.


yosuke-matsuda:
We've added the KMS server-side encryption support. Thanks.

PR #199, opened 2015-06-09
Add simple support for secondary index

rromanchuk:


yyzhu:
Thanks for the pulling request. We will add this support in future release

PR #200, opened 2015-06-12
[DynamoDB] Add call  AWSDynamoDBModeling.ignoreAttributes

hmuronaka:
[DynamoDB] Added call AWSDynamoDBModeling.ignoreAttributes.

I implemented this method for ignore some attributes of AWSDynamoDBModel when putting AWSDnamoDBModels to DynamoDB, but this method wasn't called from anywhere. So I added it.

Please check it.


yyzhu:
Thanks for the pull request. We will add it in future release.

PR #267, opened 2015-10-24
use partition_key as delete key in submitAllRecords

gcmcnutt:
(sorry for the editor whitespace.  not sure how to set Xcode to the exact same code style)

Anyway, this change is a proposed fix for what is described here http://blog.accelero.com/2015/10/amazons-ios-sdk-kinesisrecorder-bug.html -- a problem with concurrent saveRecord and submitAllRecords.

Please consider for a future version of KinesisRecorder.


yosuke-matsuda:
Thank you for bringing the issue to our attention. We will evaluate the changes. Thanks.

yosuke-matsuda:
We eliminated the use of `rowid` in the version 2.3.2. Thank you!

PR #269, opened 2015-10-28
TARGET_IPHONE_SIMULATOR has been deprecated

objectiveSee:
Update to use new build flag for detecting simulator. May need to include TargetConditionals.h if used in Swift code, which is not the case here.


yosuke-matsuda:
Thank you for the pull request. We will update the third party libraries with the latest versions in the future release to address this. Thanks.

yosuke-matsuda:
We updated the version of `UICKeyChainStore` and eliminated the use of `TARGET_IPHONE_SIMULATOR`. Thank you.

PR #285, opened 2015-12-02
Added correct URL for GovCloud STS endpoint

msullivan247:
Without this change, calls STS requests with the `AWSRegionUSGovWest1` region get assigned an endpoint with the `https://sts.amazonaws.com` URL, causing authentication requests with that region to fail. This change assigns the correct URL for the GovCloud STS endpoint (`https://sts.us-gov-west-1.amazonaws.com`).


yosuke-matsuda:
Thank you for the pull request!

PR #295, opened 2015-12-15
Fix AWSCore static declarations

hoppenichu:


yosuke-matsuda:
Yes, these variables are not exposed in the header file, so we should put the `static` keyword. Thank you!

hoppenichu:
:beer:

PR #300, opened 2016-01-06
Adding a way to import pkcs12 data into keychain for IoT use

anozaki:
Needed a way to import cert/key that was provisioned outside of this SDK, so here is an implementation that will import a pkcs12 data into the keychain. 

Following command will package the private key and cert into a p12 file that can be imported into the keychain using the following code.
openssl pkcs12 -export -in cert.pem -inkey key.pem -out cloud-key.p12

### DIFF
--- a/AWSIoT/AWSIoTManager.h
+++ b/AWSIoT/AWSIoTManager.h
@@ -185,6 +185,20 @@
  */
 - (void)createKeysAndCertificateFromCsr:(NSDictionary *)csrDictionary callback:(void (^)(AWSIoTCreateCertificateResponse *mainResponse))callback;
 
+
+// openssl pkcs12 -export -in cert.pem -inkey key.pem -CAFile root-ca.crt -out cloud-key.p12
+/**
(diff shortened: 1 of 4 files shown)


gkwicker:
Can you tell us a little more about your use case?  AWS IoT will only authenticate client certificates which were created using the AWS IoT API.

anozaki:
We are provisioning through the AWS IoT API but not on the iPad. We don't want to give the users (iPad) the access privilege to create certificate.

gkwicker:
Thanks for filling us in on the use case.  It's really helpful for us to know how customers are using the SDKs, what issues they're running into, and if they're able to work around them or directly address them like you have here.  For enhancement or new feature pull requests like this one, our procedure is to evaluate them and possibly bring them (or an alternate implementation which provides the same functionality) into our internal development, test, and release procedure.  In the meantime, we're glad you were able to find a solution that works and please let us know about any other issues/feature requests/etc...

PR #323, opened 2016-02-25
Allow setting of allowsCellularAccess via AWSNetworkingConfiguration

humblehacker:


humblehacker:
Great, I'll make that change as well.

humblehacker:
@yosuke-matsuda Thanks for your review.  I updated `AWSURLSessionManager` as requested, and squashed the commit to include it.

yosuke-matsuda:
Thank you so much. We will update the podspec and frameworks with the next release.

PR #325, opened 2016-03-01
Fixed timer schedule bug

alistairg:
There's a bug, certainly against iOS 9.3, that causes the keepalive timer to not actually run in WSS mode.

This causes the MQTT connection to drop, and reconnect, every 60s on an idle connection. This change fixes that scheduling issue.


gkwicker:
Hi Alistair, is this also true when running the [WebSocket example program](https://github.com/awslabs/aws-sdk-ios-samples/tree/master/IoTTemperatureControl-Sample/Swift)?  If you see the problem when running this application, can you please send the console output after deleting  the query parameter `X-Amz-Credential` and its value?

PR #327, opened 2016-03-01
avoid infinite loop on network failure

igor-makarov:
When `[AWSKinesisRecorder -submitRecordsForStream:records:partitionKeys:putPartitionKeys:retryPartitionKeys:]` fails due to network error, it fails to report this error to `AWSAbstractKinesisRecorder`. As a result, it attempts to resend them again immediately, without updating the `retry_count`, causing it to loop until network connectivity is restored, while consuming intensive CPU resources. In my opinion, the retry count shouldn't be updated in this case since this is not a server error, but rather the entire submit should be aborted and retried by the calling application at a later time.

### DIFF
--- a/AWSKinesis/AWSAbstractKinesisRecorder.m
+++ b/AWSKinesis/AWSAbstractKinesisRecorder.m
@@ -271,31 +271,37 @@ NSString *const AWSKinesisAbstractClientRecorderDatabasePathPrefix = @"com/amazo
 
                     NSString *streamName = temporaryRecords[0][@"stream_name"];
 
-                    [[self.recorderHelper submitRecordsForStream:streamName
-                                                         records:temporaryRecords
-                                                   partitionKeys:partitionKeys
-                                                putPartitionKeys:putPartitionKeys
-                                              retryPartitionKeys:retryPartitionKeys] waitUntilFinished];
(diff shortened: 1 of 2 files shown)


yosuke-matsuda:
It depends on the kind of exceptions. For 5xx exceptions, it makes sense to keep retrying since they are service exceptions, and the service should eventually come back online. For 4xx, there are mainly three possibilities:
- Throttling exception: This should be retried later time.
- Invalid/expired AWS temporary credentials: We need to refresh the credentials and retry. However, we should not keep retrying when there is an issue with the permissions/credentials generation.
- Malformed/invalid request: We should not retry since they will always fail.

For 5xx and throttling exceptions, it makes sense to make the retry count configurable.

igor-makarov:
This is probably the most correct way, however it raises several questions,
which seem to be beyond the scope of time I can currently contribute to
this project.

My pull request does, however, fix a painful bug, and does so without
breaking existing functionality.

I suggest you merge it so that other users don't encounter the bug (which
impacts not data, but the end user battery as well) and I'll transfer the
rest of this discussion over to a new issue.

igor-makarov:
Sorry to bug you, but a couple of weeks have passed with no reply.
How about a merge?

yosuke-matsuda:
We've released the AWS Mobile SDK for iOS 2.4.0 to avoid the retry loop when the device is offline. Thanks.

PR #329, opened 2016-03-02
Add ability to invalidate an AWSS3TransferUtility

humblehacker:
Background: It's not possible to change the configuration of an already created `NSURLSession` - it must instead be invalidated and re-created with the updated configuration.  `AWSS3TransferUtility` inherits this limitation, so it should also expose the invalidation functionality.

This is necessary, for example, if one wants to change the `allowsCellularAccess` configuration property on an already running `NSURLSession`.

Note: I opted for parameterized cancelation to keep the API small, but if this isn't desired, I could split into two different methods (much like the underlying `invalidateAndCancel` and `finishTasksAndInvalidate` methods on `NSURLSession`)

### DIFF
--- a/AWSS3/AWSS3TransferUtility.h
+++ b/AWSS3/AWSS3TransferUtility.h
@@ -26,6 +26,13 @@ NS_ASSUME_NONNULL_BEGIN
 @class AWSS3TransferUtilityDownloadExpression;
 
 /**
+ The invalidation completion handler.
+
+ @param error Returns the error that caused invalidation, or nil if the invalidation was explicit.
+ */
+typedef void (^AWSS3TransferUtilityInvalidationCompletionHandlerBlock) (NSError * __nullable error);
+
+/**
  The upload completion handler.
 
  @param task  The upload task object.
@@ -302,6 +309,15 @@ handleEventsForBackgroundURLSession:(NSString *)identifier
                                 downloadTask:(nullable void (^)(AWSS3TransferUtilityDownloadTask *downloadTask,
                                                                 AWSS3TransferUtilityDownloadProgressBlock * __nullable downloadProgressBlockReference,
                                                                 AWSS3TransferUtilityDownloadCompletionHandlerBlock * __nullable completionHandlerReference))downloadBlocksAssigner;
+
+/**
+  Invalidates the underlying NSURLSession.
+
+  @param cancel Cancel all tasks before invalidating
+  @param completionHandler Completion handler to be called after all outstanding tasks have been canceled or have finished.
+ */
+- (void)invalidateAndCancel:(BOOL)cancel withCompletionHandler:(AWSS3TransferUtilityInvalidationCompletionHandlerBlock)completionHandler;
+
 #pragma clang diagnostic pop
 
 /**
--- a/AWSS3/AWSS3TransferUtility.m
+++ b/AWSS3/AWSS3TransferUtility.m
@@ -32,6 +32,7 @@ NSString *const AWSS3TransferUtilityUserAgent = @"transfer-utility";
 @property (strong, nonatomic) NSString *temporaryDirectoryPath;
 @property (strong, nonatomic) AWSSynchronizedMutableDictionary *taskDictionary;
 @property (copy, nonatomic) void (^backgroundURLSessionCompletionHandler)();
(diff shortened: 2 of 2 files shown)


yosuke-matsuda:
Thank you so much for your contributions :)

In general, if you need transfer utility instances with different configurations, it may be easier to use `+ registerS3TransferUtilityWithConfiguration:forKey:` to create separate instances and keep them until your app does not need them anymore.

`+ removeS3TransferUtilityForKey:` should call `- invalidateAndCancel`. I suppose each upload/download request receives a cancellation, which is exposed to the developer, so we can simply log the error from `- didBecomeInvalidWithError:` if there is any. I do not think it is really actionable anyways.

I need to think about it a little more, but what do you think about it?

humblehacker:
That approach could be workable, and would prevent possible reuse of an invalid session, since this PR doesn't address removal of the associated transfer utility.  But one thing I require is some indication that the previous session has successfully been invalidated before I start a new one.  I wouldn't want to cause issues by trying to create new upload tasks for files that already have upload tasks in progress.  

I suppose we could add a completion block to `removeS3TransferUtilityForKey:`, but it seems a bit unclean, and it doesn't provide the option to choose cancelation over allowing tasks to finish (though I'm not sure how big an issue that is - I only require cancelation).

yosuke-matsuda:
Could you describe your use case in more details so that we can understand your needs? Especially, I'm interested in what requirement derives this:
> some indication that the previous session has successfully been invalidated before I start a new one

humblehacker:
Sure, we want to allow our users to enable/disable uploads over cellular connections through a switch in our in-app settings page.

First, when the user toggles this switch while uploads are already in-progress, we need to `invalidateAndCancel` the session, then wait for all the tasks to be canceled and for the `NSURLSessionDelegate` `URLSession:didBecomeInvalidWithError:` method to be called - in order to properly clean-up.

Second, we create a new session with the `allowsCellularAccess` property changed, and re-add all the files that were previously in-progress.

We don't want these two steps to overlap.

humblehacker:
@yosuke-matsuda Any thoughts on the above?

yosuke-matsuda:
Does the following make sense with your use case?

1. The end user disables the cellular transfer.
2. Create a new instance of transfer utility with the wifi only configuration.
3. Call `+ removeS3TransferUtilityForKey:` for the transfer utility with the cellular configuration. (Internally, the invalidate and cancel method is called.)
4. Every time you receive a cancellation error, add the transfer to the new TU.
5. Eventually, all transfers are moved to the new TU.
6. The old TU is deallocated (the SDK should release the strong reference to TU in `- URLSession:didBecomeInvalidWithError:`)

Alternatively, you can have an external module that keeps track of the progress of each transfer.

humblehacker:
Thanks for considering this and taking the time to give this some thought.  

Your solution could work, but I'm afraid at the expense of increased complexity on my end.  Managing simultaneous transfer utilities as described would be difficult to manage and write tests for. It's really quite straightforward to depend on the `URLSession:didBecomeInvalidWithError:` callback - it seems a shame not to expose it in some manner.

Could we instead add a `invalidationCompletionHandler` property to `AWSS3TransferUtility` that gets called from `URLSession:didBecomeInvalidWithError:`and call `invalidateAndCancel` from `+ removeS3TransferUtilityForKey:`?

yosuke-matsuda:
We can consider sending out a notification using `NSNotificationCenter` when `- URLSession:didBecomeInvalidWithError:` is called internally. In this way, we can keep the API simple while meeting your needs.

> re-add all the files that were previously in-progress.

By the way, how are you getting a list of in-progress transfers? I am not sure if there is a reliable way to get this list without some external module that keeps track of all transfers. If you have this manager module, it may not be too difficult to manage two instances of TUs.

humblehacker:
Just getting back to this.  Thanks for your suggestion.  The invalidation notification would serve the purpose.  I'll submit another PR with those changes.

As to your questions, I am already managing the list of in-progress transfers - necessary to implement background uploads.  I _could_ modify it to manage two TUs as you suggest, but I already have a working implementation dependent on `URLSession:didBecomeInvalidWithError:` using the code in this PR, and it doesn't seem to be an improvement to add additional complexity solely to avoid the dependency on `URLSession:didBecomeInvalidWithError:`.

humblehacker:
@yosuke-matsuda I've opened new PR #334.  Closing this one.

PR #334, opened 2016-03-16
Add ability to invalidate an AWSS3TransferUtility

humblehacker:
Background: It's not possible to change the configuration of an already created NSURLSession - it must instead be invalidated and re-created with the updated configuration.  AWSS3TransferUtility inherits this limitation, so it should also expose the invalidation functionality.

### DIFF
--- a/AWSS3/AWSS3TransferUtility.h
+++ b/AWSS3/AWSS3TransferUtility.h
@@ -16,7 +16,7 @@
 #import <UIKit/UIKit.h>
 #import <AWSCore/AWSCore.h>
 
-FOUNDATION_EXPORT NSString *const AWSS3SessionInvalidatedNotification;
+FOUNDATION_EXPORT NSString *const AWSS3TransferUtilityURLSessionDidBecomeInvalidNotification;
 
 NS_ASSUME_NONNULL_BEGIN
 
@@ -199,6 +199,10 @@ typedef void (^AWSS3TransferUtilityDownloadProgressBlock) (AWSS3TransferUtilityD
 
 /**
  Removes the service client associated with the key and release it.
+
+ The underlying NSURLSession is invalidated, and after the invalidation has completed the transfer is utility removed.
+
+ Observe the AWSS3TransferUtilityURLSessionDidBecomeInvalidNotification to be informed when this has occurred.
 
  @warning Before calling this method, make sure no method is running on this client.
 
--- a/AWSS3/AWSS3TransferUtility.h
+++ b/AWSS3/AWSS3TransferUtility.h
@@ -16,6 +16,8 @@
 #import <UIKit/UIKit.h>
 #import <AWSCore/AWSCore.h>
 
+FOUNDATION_EXPORT NSString *const AWSS3SessionInvalidatedNotification;
+
 NS_ASSUME_NONNULL_BEGIN
 
 @class AWSS3TransferUtilityTask;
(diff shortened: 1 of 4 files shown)


AWSS3/AWSS3TransferUtility.m
```diff
@@ -125,6 +126,9 @@ + (instancetype)S3TransferUtilityForKey:(NSString *)key {
 }
 
 + (void)removeS3TransferUtilityForKey:(NSString *)key {
+    AWSS3TransferUtility *transferUtility = [self S3TransferUtilityForKey:key];
+    if (transferUtility)
+        [transferUtility.session invalidateAndCancel];
     [_serviceClients removeObjectForKey:key];
```
yosuke-matsuda: This line should to be called in `- URLSession:didBecomeInvalidWithError:`.

AWSS3/AWSS3TransferUtility.m
```diff
@@ -125,6 +126,9 @@ + (instancetype)S3TransferUtilityForKey:(NSString *)key {
 }
 
 + (void)removeS3TransferUtilityForKey:(NSString *)key {
+    AWSS3TransferUtility *transferUtility = [self S3TransferUtilityForKey:key];
+    if (transferUtility)
```
yosuke-matsuda: This is nitpicking, but please do not omit `{}`.

AWSS3/AWSS3TransferUtility.m
humblehacker: Yes, that makes sense.

AWSS3/AWSS3TransferUtility.m
humblehacker: Nitpicking is fine. :)

AWSS3/AWSS3TransferUtility.h
```diff
@@ -16,6 +16,8 @@
 #import <UIKit/UIKit.h>
 #import <AWSCore/AWSCore.h>
 
+FOUNDATION_EXPORT NSString *const AWSS3SessionInvalidatedNotification;
```
yosuke-matsuda: How about `AWSS3TransferUtilityURLSessionDidBecomeInvalidNotification`? It's pretty verbose but more descriptive and following the delegate method name.

AWSS3/AWSS3TransferUtility.m
humblehacker: I won't have the key in the delegate method, so I'll have to either have to:

1. search the dictionary for the object 
2. extract the key from the `_sessionIdentifier`
3. add the key as a property on `AWSS3TransferUtility` 

Options 2 and 3 are pretty ugly, so I'd go with option 1.  Your thoughts?

yosuke-matsuda: Option 1 sounds good to me!

humblehacker:
@yosuke-matsuda Thanks for all of your helpful feedback.  Hopefully these latest changes meet your approval. Let me know if you prefer I squash these changes into a single commit prior to merging.

AWSS3/AWSS3TransferUtility.h
```diff
@@ -16,7 +16,7 @@
 #import <UIKit/UIKit.h>
 #import <AWSCore/AWSCore.h>
 
-FOUNDATION_EXPORT NSString *const AWSS3SessionInvalidatedNotification;
+FOUNDATION_EXPORT NSString *const AWSS3TransferUtilityURLSessionDidBecomeInvalidNotification;
```
yosuke-matsuda: One very small change recommendation. This line should go below `NS_ASSUME_NONNULL_BEGIN` so that Xcode won't complain about the nullability annotation. Thank you so much for the pull request. We'll merge it in the next release. Thanks.

PR #353, opened 2016-04-13
Fixed race condition in -getIdentityId in AWSAbstractCognitoIdentityProvider

JChanceHud:
The semaphore was being signaled prior to the new identityId being set which was resulting in continuation blocks on other threads being executed and encountering a nil identityId.


yosuke-matsuda:
Thank you for the pull request. We will merge and release the fix once we complete the evaluation. Thank you!

PR #355, opened 2016-04-15
Fixed memory issue when assigning data to AWSS3TransferUtilityDownloadTask

dabrut:
If you start downloading a large file and the app crashes or is terminated by the system and then relaunched by the user the download will continue but AWSS3TransferUtilityDownloadTask will lose the location information. 

So when the completion delegate gets called
```
- (void)URLSession:(NSURLSession *)session
      downloadTask:(NSURLSessionDownloadTask *)downloadTask
didFinishDownloadingToURL:(NSURL *)location 
```
it tries to load the entire object in memory which crashes the app if downloading large files.
The file should be mapped into virtual memory instead.

### DIFF
--- a/AWSS3/AWSS3TransferUtility.m
+++ b/AWSS3/AWSS3TransferUtility.m
@@ -658,7 +658,11 @@ didFinishDownloadingToURL:(NSURL *)location {
             }
         }
     } else {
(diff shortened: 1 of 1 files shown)


yosuke-matsuda:
This is definitely a better way to handle the data. Thank you for the pull request. We'll merge it once we complete the evaluation. Thanks.

PR #367, opened 2016-05-02
Fixes for sending correct SSE-C parameters in Request Headers

sugampandey:
Include x-amz-server-side-encryption-customer-algorithm in X-Amz-SignedHeaders in 2.3.6 version of SDK


yosuke-matsuda:
A more generic way to handle this is to add `requestHeaders` to `AWSS3GetPreSignedURLRequest` and `AWSS3TransferUtilityExpression` just like `requestParameters`. `contentType` and `contentMD5` should be updated to simply create an entry to this new dictionary. In this way, you can sign any headers while addressing this specific use case.

karthiksaligrama:
merged with 2.4.2

PR #368, opened 2016-05-02
Improved error handling for saveRecord

joseandro:
This change intends to make the method ```saveRecord``` easier to read and understand

* Error returnings are now standardized throughout the method. 
* The former implementation was using the variable ```error``` everywhere. 
* Standard ```[AWSTaks taskWithError:error]``` returns where it was appropriate.
* Improved error logging

### DIFF
--- a/AWSKinesis/AWSAbstractKinesisRecorder.m
+++ b/AWSKinesis/AWSAbstractKinesisRecorder.m
@@ -158,8 +158,6 @@ NSString *const AWSKinesisAbstractClientRecorderDatabasePathPrefix = @"com/amazo
                                               @"retry_count" : @0
                                               }
                            ];
-            
-            
             if (!result) {
                 AWSLogError(@"SQLite error. [%@]", db.lastError);
                 dbError = db.lastError;
@@ -186,6 +184,7 @@ NSString *const AWSKinesisAbstractClientRecorderDatabasePathPrefix = @"com/amazo
         __block NSError *fileError = nil;
         NSDictionary *attributes = [[NSFileManager defaultManager] attributesOfItemAtPath:databasePath
                                                                                     error:&fileError];
+        __block NSError *attError = nil;
         if (attributes) {
             NSUInteger fileSize = (NSUInteger)[attributes fileSize];
             [self.recorderHelper checkByteThresholdForNotification:notificationByteThreshold
@@ -205,8 +204,7 @@ NSString *const AWSKinesisAbstractClientRecorderDatabasePathPrefix = @"com/amazo
                                    ];
                     if (!result) {
                         AWSLogError(@"SQLite error. [%@]", db.lastError);
-                        dbError = db.lastError;
-                        return [AWSTask taskWithError:dbError];
+                        attError = db.lastError;
                     }
                 }];
             }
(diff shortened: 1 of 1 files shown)


AWSKinesis/AWSAbstractKinesisRecorder.m
```diff
@@ -208,17 +205,20 @@ - (AWSTask *)saveRecord:(NSData *)data
                                    ];
                     if (!result) {
                         AWSLogError(@"SQLite error. [%@]", db.lastError);
-                        error = db.lastError;
-                        return;
+                        dbError = db.lastError;
+                        return [AWSTask taskWithError:dbError];
```
yosuke-matsuda: Looks like this is a bug. You cannot return an instance of `AWSTask` here.

joseandro: Shouldn't this method always return ```AWSTask```?
I thought this was a miss in the previous code.
Thanks for looking into it.

AWSKinesis/AWSAbstractKinesisRecorder.m
```diff
@@ -178,17 +178,14 @@ - (AWSTask *)saveRecord:(NSData *)data
                                ];
                 if (!result) {
                     AWSLogError(@"SQLite error. [%@]", db.lastError);
-                    error = db.lastError;
+                    dbError = db.lastError;
                 }
             }];
         }
-
-        if (error) {
```
yosuke-matsuda: This changes the logical flow of the method. Probably, we should fail fast when there are any issues with SQLite.

AWSKinesis/AWSAbstractKinesisRecorder.m
```diff
@@ -132,17 +132,17 @@ - (AWSTask *)saveRecord:(NSData *)data
     if ([data length] > 256 * 1024) {
         return [AWSTask taskWithError:[self.recorderHelper dataTooLargeError]];
     }
-
+
```
yosuke-matsuda: This is totally nitpicking, but please don't add unnecessary white spaces.

AWSKinesis/AWSAbstractKinesisRecorder.m
joseandro: Shall we really return here? 
I thought the code bellow shouldn't be locked by the errors from the code above

AWSKinesis/AWSAbstractKinesisRecorder.m
yosuke-matsuda: I totally see it's a little confusing since the blocks are nested. Here is the definition of `- inDatabase:`

    - (void)inDatabase:(void (^)(AWSFMDatabase *db))block;

This is different than the `AWSTask` related blocks. This particular block needs to return `void`.

joseandro: Good catch! Code fix is up

AWSKinesis/AWSAbstractKinesisRecorder.m
joseandro: I understand and agree. Unnecessary white spaces removed.

AWSKinesis/AWSAbstractKinesisRecorder.m
yosuke-matsuda: Since `AWSTask` can expose only one error message, whenever we encounter an issue, we should return it right away. Otherwise, the SDK potentially encounters multiple errors, but returns only one of them. (There are ways to wrap around an error inside of an error, but it's not a common practice in Cocoa Framework.)

joseandro: That indeed makes sense, @yosuke-matsuda 
PR has been updated to reflect this strategy

joseandro:
Since there was no interest I am closing this PR

PR #388, opened 2016-05-23
Fix wrong error codes.

rance-attack:
should use "/" instead of "%"


yosuke-matsuda:
Thank you for the pull request. We'll merge it with the next release. Thanks.

karthiksaligrama:
Merged with 2.4.3

PR #410, opened 2016-06-23
Fix misplaced version info

hubertwang:
The value of "app_version_name" should be app version. And for "app_version_code", it should be app build.


karthiksaligrama:
merged with 2.4.6

Issue #416, opened 2016-07-06
Why kinesis retry only 3times


JMASystems:
In AWSAbstractKinesisRecorder.m,
submitAllRecords method retry only 3times,and failure records is deleted.
This specification will cause a problem at the time of the off-line uses.

My application used in the off-line for long-term,then many records were deleted.
Will this specification can not be abolished,or is there any solution for it?

Thank you.

fosterzhang:
@JMASystems This is to prevent bogus records from being retried over and over again. It's better to check your network before calling `submitAllRecords`. Meanwhile, we'll see how to improve error handling in the SDK. Stay tuned.

karthiksaligrama:
@JMASystems can you just not call `submitAllRecords` when you don't have network reachability?

JMASystems:
Thank you for reply.

The communication is very unstable in our environment.
So even if we check reachability in advance,api call might be failed sometime.

Can we set a custom retry count property?

nexus49:
+1 for @JMASystems comment

Also it would be very beneficial to be able to know when a message was dropped, so that the client could re enter that record depending on the error. We are also trying to build an app that supports a lot of on/offline and bad connectivity scenarios. Its very hard to debug if kinesis drops messages or if there could be other errors in play.

nexus49:
@JMASystems where you able to find a solution to this issue? I hacked the retry of the sdk but still seem to be missing messages in certain offline scenario's.

JMASystems:
At ver 2.4.8.
It seems to be implemented the KinesisRetryHandler.
https://github.com/aws/aws-sdk-ios/blob/d0154d3f1e1cc20fad8c7339208ee208f1d8c516/AWSKinesis/AWSKinesisRequestRetryHandler.m
So,it might be realized if we modify KinesisConfigration so that we can pass your own handler.
@nexus49 How about that.

PR #522, opened 2016-11-14
Add retryHandler configuration

JMASystems:
Fix to adopt retryHandler if it was set.
In order to set the custom retryHandler.
It is related to #416.


karthiksaligrama:
I don't think its a good idea to expose the entire retry handler itself. I would rather expose a retry template that the handler can consume and only implement bits of retry logic.

JMASystems:
Is it mean following process is needed?

1.Add the retry template protocol header
2.Implements some template class.For example "EternalRetryTemplate","NumLimitRetryTemplate","ConditionalRetryTemplate"
3.Add initializer like initWithTemplate:(id<TemplateProtocol>)template

kvasukib:
Closing this PR due to inactivity. Please re-open or open a new one when the comment is addressed.

PR #563, opened 2017-01-09
Allow for future expired/unauthed token calls to properly refresh the aws token.

donileo:
This pull request allows for each (expired token/unauthed) call via aws to reset the identityId, call to refresh the developer aws token and then attempt to get valid aws credentials. If the second (nested) call fails then an error is returned stating that resetting the identityId was not enough to make the call. Subsequent calls go through the same process.

### DIFF
--- a/AWSCore/Authentication/AWSCredentialsProvider.m
+++ b/AWSCore/Authentication/AWSCredentialsProvider.m
@@ -271,7 +271,6 @@ static NSString *const AWSCredentialsProviderKeychainIdentityId = @"identityId";
 @property (nonatomic, strong) AWSCredentials *internalCredentials;
 @property (atomic, assign, getter=isRefreshingCredentials) BOOL refreshingCredentials;
 @property (nonatomic, strong) NSDictionary<NSString *, NSString *> *cachedLogins;
-@property (atomic, assign) BOOL hasClearedIdentityId;
 
 @end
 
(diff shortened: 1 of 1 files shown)


rmartell:
Is there any idea of when this patch might get added?  I'm having the same issue.  In my case, I'm using the APIGateway generated code, and executing multiple API calls.  Whenever one fails (due to timeout), they all fail, even if the first one is in the middle of updating the api key.

I'm using a Cognito User Pool.  The only possible issues I can think of with this patch are:

1) If you're calling several routines simultaneously, this might cause more traffic than otherwise required (because each one would be refreshing), and
2) Because of 1, if there are any variables that are updated in the AWS stack, they would need to be synchronized if they aren't already.

As it currently stands (as of 2.5.3) , the code is setting the `hasClearedIdentityId` variable, which is never getting reset anywhere, which is obviously incorrect.

behrooziAWS:
@rmartell I've looked at the patch and it looks promising.  There exists some synchronization logic to serialize requests to getCredentials so I don't believe additional precautions are necessary.  I'll review this with the SDK team for inclusion.

donileo:
Pull request updated to resolve previous conflicts.

markuswinkler:
@behrooziAWS is there an estimate for including this patch?

markuswinkler:
@rohandubal My tokens still expire without being refreshed on 2.5.7. Has the fix above been implemented?

donileo:
The above patch won't necessarily refresh your tokens in all cases. It will attempt to refresh your token but if it fails on the subsequent attempt it will stop. The patch above was mostly made to allow refreshing the token to take place even if previous calls also failed to refresh the token.

PR #578, opened 2017-02-03
Added delegate method for completion of an intent

ccreonopoulos:
Proposition of a new delegate method for The AWSLexInteractionKitDelegate to notify the client of the completion of an Intent.

Method is similar to `interactionKit:onDialogReadyForFulfillmentForIntent:slots:` but gets called when the whole intent is complete as a separate method instead of having to check again for the dialogeState in the body of the`interactionKit:switchModeInput:completionSource` delegate method.

### DIFF
--- a/AWSLex/AWSLexInteractionKit.h
+++ b/AWSLex/AWSLexInteractionKit.h
@@ -100,6 +100,11 @@ typedef NS_ENUM(NSInteger, AWSLexSpeechEncoding) {
 - (void)interactionKit:(AWSLexInteractionKit *)interactionKit onDialogReadyForFulfillmentForIntent:(NSString *)intentName slots:(NSDictionary *)slots;
 
 /*
+ * Sent to delegate when the all the required slots are filled, the intent is fullfiled and the dialog is complete.
+ */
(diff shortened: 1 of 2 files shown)


karthiksaligrama:
We are still figuring out if it makes sense to have multiple API's which do the same thing. Today we have it for `readyForFullfillment` because that is a particular case in lex where the user has to probably do something different, but it can be still accomplished with the `interactionKit:switchModeInput:completionSource` api. 

I would like to hear your use case more before i proceed on this PR.

ccreonopoulos:
The reason for having it in its own separate method is to follow continuous interaction delegate method standards more closely. Similarly to Apples' NSURLSessionTaskDelegate, where `URLSession:task:didSendBodyData:totalBytesSent:totalBytesExpectedToSend:`
 gets called periodically but `URLSession:task:didCompleteWithError:` is called once when a download is complete. 
This way,  not only redundant if checks like `switchModeInput.dialogState == AWSLexDialogStateFulfilled` don't need to be made if the user does not explicitly need to do something different between interactions.

A use case I ran into was where there were more than a few nested if statements I wrote, in which I had to check each new filled slot for the returned value and **then** check if it was the last one. Lots of redundant ifs and a new method ( a helper to avoid duplicate code) is being called in each of those checks. 

I believe that if this method is added , neither a helper method nor this level of nesting ifs would be necessary.

PR #589, opened 2017-02-15
Fix AWSLex.podspec

yaakaito:


minbi:
Can you explain why you have changed to double quotes?

yaakaito:
'#{s.name}' is not expand in single quotes.

minbi:
Thank you. We will merge this with the next release.

minbi:
The entry here seems to not affect the operation of Bluefront in the project. We have chosen to simply remove the entire search path. Please re-open if there is an issue with this.

Issue #565, opened 2017-01-12
Shadow Operation Still In Progress! BUG!


cocotutch:
Hi,

I'm trying (!!!) to use the AWS IoT iOS SDK with a device shadow, and while I can subscribe and get a shadow, I cannot seem to update it more than once.

I send off an update JSON string, which works, but then all subsequent tries results in an error "Operation still in progress on shadow"

As per usual, the AWS documentation is extraordinarily unclear. Hence the Github issue.

Would appreciate it if someone could shed some light on this ridiculous issue. Thanks!

cocotutch:
It seems that `- (void) shadowOperationTimeoutOnTimer:(NSTimer *)timer {` is not being called whatsoever.

gkwicker:
Hi @cocotutch, we're sorry you're having this issue.  Have you had a look at the [IotTemperatureControl-Sample](https://github.com/awslabs/aws-sdk-ios-samples/blob/master/IoTTemperatureControl-Sample/Swift/IoTTemperatureControlSample/ViewController.swift) example app?  It demonstrates use of Device Shadows.  The main things to note in the sample app are:

1. [Registering with the device shadow](https://github.com/awslabs/aws-sdk-ios-samples/blob/master/IoTTemperatureControl-Sample/Swift/IoTTemperatureControlSample/ViewController.swift#L200)
2. [Handling callbacks from the device shadow](https://github.com/awslabs/aws-sdk-ios-samples/blob/master/IoTTemperatureControl-Sample/Swift/IoTTemperatureControlSample/ViewController.swift#L163)
3. [Performing updates](https://github.com/awslabs/aws-sdk-ios-samples/blob/master/IoTTemperatureControl-Sample/Swift/IoTTemperatureControlSample/ViewController.swift#L74)

Another thing to note is that you need to make sure that you've received a response from any update/get/delete Device Shadow operation before attempting another one.  In the above sample app, you'll see this [being handled using a boolean variable](https://github.com/awslabs/aws-sdk-ios-samples/blob/master/IoTTemperatureControl-Sample/Swift/IoTTemperatureControlSample/ViewController.swift#L70).

Let us know if that information helps, and thanks for using AWS IoT.

cocotutch:
Hi, I get a success Boolean back from the updateShadow() function, however it never lets me perform another request.

I've had a look at the IoT sample app, and I have the same setup.

The shadow timer never times out regardless of whether the operation is a success or not.

[...]

PR #601, opened 2017-02-24
Fix issue reported in Shadow Operation Still In Progress! BUG! #565

ramki982:
In my understanding - The main reason for this is that shadow timer is neither using the scheduledTimerWithTimeInterval nor adding it to the run loop using [NSRunLoop mainRunLoop] addTimer.

I'm referring to the following lines in the AWSIotDataManager.m
shadow.timer = [NSTimer timerWithTimeInterval:shadow.operationTimeout target:self selector: @selector(shadowOperationTimeoutOnTimer:) userInfo:name repeats:NO];

This is based on my own testing, seeing why how other timers in SDK are coded and also based on what i read in different stack overflow threads.

http://stackoverflow.com/questions/10522928/run-repeating-nstimer-with-gcd

thanks
Ram


fengsongAWS:
Thanks for pointing this issue out. You are right and the timer needs to be triggered. We are aware of this issue and working on to fix it. Please stay tuned for the future release. 
Thanks for your interest in AWS IoT IOS SDK.

PR #631, opened 2017-04-05
Fixes a typo/bug when accessing AWSAPIGatewayResponse.rawResponse

lyricsboy:
The `rawReponse` (misspelled) property would never be non-nil, and the actual `rawResponse` property was inaccessible, particularly in Swift.

I would like to add test coverage for this, but I don't have appropriate values for `credentials.json`. Is there a recommended approach to setting that up?


rohandubal:
Since this will trigger compilation issues for people already consuming the SDK, we will merge this with 2.6.0 release. Adding it to that milestone.

rohandubal:
Fixed in 2.6.0.

PR #637, opened 2017-04-11
Lex SDK - Adding input transcript returned from postContent

alluneediswill:
### Description

For Lex API, transcript of the user audio input is now available in the response header.
http://docs.aws.amazon.com/lex/latest/dg/gs-bp-details-postcontent-flow.html

This pull request makes input transcript available in AWSLexPostContentResponse, then AWSLexSwitchModeInput, so that AWSLexInteractionDelegate can have access to the input transcript.

### Test

Tested out with a Lex preview account on an iOS 10.2 and an iOS 10.3.1 devices.


mlshon:
Hi,
Thanks for your pr.
At this time, we cannot accept this pr but we will release this feature in the next released version.

PR #644, opened 2017-04-21
Change APIGatewayClient to use timeouts provided by configuration

jkennington:
The AWSAPIGatewayClient accepts and requires an AWSServiceConfiguration. It would be nice to be able to use the timeout interval properties on the configuration to set desired timeout intervals on the NSURLSession used by the AWSAPIGatewayClient.

This was discussed in https://github.com/aws/aws-sdk-ios/issues/101 in November 2015.


minbi:
Hi @jkennington ,

Thank you for submitting this pull request. We are evaluating the semantic change.

PR #656, opened 2017-05-03
Fix compilation error when using AWS Mobile Analytics along with AWS Pinpoint in one project

podkovyrin:
When using both analytics libraries in one project there is a compilation error:
```<...>/Pods/Headers/Public/AWSPinpoint/AWSPinpointConfiguration.h:22:15: Typedef redefinition with different types ('void (^)(AWSPinpoint * _Nonnull __strong)' vs 'void (^)(AWSMobileAnalytics *__strong)')```

### DIFF
--- a/AWSPinpoint/AWSPinpointConfiguration.h
+++ b/AWSPinpoint/AWSPinpointConfiguration.h
@@ -19,7 +19,7 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-typedef void(^AWSInitializationCompletionBlock)(AWSPinpoint * _Nonnull pinpoint);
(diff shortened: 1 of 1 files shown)


rohandubal:
Hi, 

Could you explain the use case of using both Mobile Analytics and Pinpoint SDK ? Pinpoint SDK supports logging events for your mobile analytics.

Thanks,
Rohan

podkovyrin:
Hi Rohan,
We use both because it is product requirement. One project consists of two business parts.

rcha:
Please be aware that Amazon Mobile Analytics is now part of Pinpoint and that you should start migrating to use the Pinpoint SDK instead as mentioned here: http://docs.aws.amazon.com/mobile/sdkforios/developerguide/analytics.html 

The fix will most likely be included in the next release.

Issue #679, opened 2017-05-24
Crash in AWSCognitoIdentityUserPool calculateSecretHash


zrzka:
# Description

Several new users started complaining that they're unable to sign in in our iOS application. Application simply crashes. But they are able to sign in via web interface, ... I was scratching my head for a few minutes and then I found that they have accented characters in usernames. Tried this case with new user and cause was identified.

# Steps to reproduce

* Create user with username containing accented characters (_Kaštanek_ for example)
* Try to sign in on iOS
* Crash

# Crash description

* It crashes on this line https://github.com/aws/aws-sdk-ios/blob/master/AWSCognitoIdentityProvider/AWSCognitoIdentityUserPool.m#L254
* Because `cData` contains `null` -> `strlen(cData)` -> `strlen(null)` -> crash https://github.com/aws/aws-sdk-ios/blob/master/AWSCognitoIdentityProvider/AWSCognitoIdentityUserPool.m#L250
* And that's because `cStringUsingEncoding:NSASCIIStringEncoding` returns `null` because it can't be losslessly converted to ASCII encoding.

Documentation quote:

> A C string representation of the receiver using the encoding specified by encoding. **Returns NULL if the receiver cannot be losslessly converted to encoding.**

Documentation [source](https://developer.apple.com/reference/foundation/nsstring/1408489-cstringusingencoding?language=objc).

behrooziAWS:
That is definitely a bug, I'll fix for a later release.  Thanks for reporting.

zrzka:
@behrooziAWS I assume `NSUTF8StringEncoding` is correct encoding? Have to fix it locally and send new version to App Store as soon as possible. But don not want to send it to the App Store until someone confirms it. Thanks.

behrooziAWS:
This was fixed in 2.5.8.  Please reopen if you continue to have issues.

PR #680, opened 2017-05-24
use UTF8 encoding to support username w/ accents

zrzka:
Did test #679 with this change and it works. Tested username `ěščřžýáíé`.


behrooziAWS:
Thanks for posting this @zrzka, 2.5.8 contains this fix.  Unfortunately I independently made the same fix before noticing this pull request, so I can't merge it, but happy to give you the credit.

zrzka:
I don't need credit, happy it's fixed in upstream :)

PR #698, opened 2017-06-23
[Kinesis] Ensure submitAllRecords returns an error on failure to submit

jgavris:
Errors when submitting to kinesis are being swallowed. Added fixes and regression tests.

### DIFF
--- a/AWSKinesis/AWSAbstractKinesisRecorder.m
+++ b/AWSKinesis/AWSAbstractKinesisRecorder.m
@@ -279,12 +279,19 @@ NSString *const AWSKinesisAbstractClientRecorderDatabasePathPrefix = @"com/amazo
 
                     NSString *streamName = temporaryRecords[0][@"stream_name"];
 
-                    [[self.recorderHelper submitRecordsForStream:streamName
-                                                         records:temporaryRecords
-                                                   partitionKeys:partitionKeys
-                                                putPartitionKeys:putPartitionKeys
-                                              retryPartitionKeys:retryPartitionKeys
-                                                            stop:&stop] waitUntilFinished];
+                    AWSTask *submitTask = \
+                        [self.recorderHelper submitRecordsForStream:streamName
+                                                            records:temporaryRecords
+                                                      partitionKeys:partitionKeys
+                                                   putPartitionKeys:putPartitionKeys
+                                                 retryPartitionKeys:retryPartitionKeys
+                                                               stop:&stop];
+
+                    [submitTask waitUntilFinished];
+
+                    if (submitTask.error) {
+                        error = submitTask.error;
+                    }
 
                     for (NSString *partitionKey in putPartitionKeys) {
                         BOOL result = [db executeUpdate:@"DELETE FROM record WHERE partition_key = :partition_key"
--- a/AWSKinesis/AWSFirehoseRecorder.m
+++ b/AWSKinesis/AWSFirehoseRecorder.m
@@ -177,8 +177,11 @@ static AWSSynchronizedMutableDictionary *_serviceClients = nil;
     return [[self.firehose putRecordBatch:putRecordBatchInput] continueWithBlock:^id(AWSTask *task) {
         if (task.error) {
             AWSDDLogError(@"Error: [%@]", task.error);
-            if ([task.error.domain isEqualToString:NSURLErrorDomain]) {
+            const NSArray *stopErrorDomains = @[NSURLErrorDomain, AWSCognitoIdentityErrorDomain];
+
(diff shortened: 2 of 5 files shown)


AWSKinesis/AWSAbstractKinesisRecorder.m
```diff
@@ -273,12 +273,19 @@ - (AWSTask *)submitAllRecords {
 
                     NSString *streamName = temporaryRecords[0][@"stream_name"];
 
-                    [[self.recorderHelper submitRecordsForStream:streamName
-                                                         records:temporaryRecords
-                                                   partitionKeys:partitionKeys
-                                                putPartitionKeys:putPartitionKeys
-                                              retryPartitionKeys:retryPartitionKeys
-                                                            stop:&stop] waitUntilFinished];
```
jgavris: We had an invalid pool identity set on the configuration, and the error was being swallowed here.

AWSKinesis/AWSKinesisRecorder.m
```diff
@@ -181,6 +181,8 @@ - (AWSTask *)submitRecordsForStream:(NSString *)streamName
             if ([task.error.domain isEqualToString:NSURLErrorDomain]) {
                 *stop = YES;
             }
+
+            return [AWSTask taskWithError:task.error];
```
jgavris: Likewise, this was always returning a successful `nil` response.

lspectorbutterfly:
This is awesome. 👍

jgavris:
@behrooziAWS can you suggest a reviewer for this?

AWSKinesis/AWSFirehoseRecorder.m
```diff
@@ -177,8 +177,9 @@ - (AWSTask *)submitRecordsForStream:(NSString *)streamName
     return [[self.firehose putRecordBatch:putRecordBatchInput] continueWithBlock:^id(AWSTask *task) {
         if (task.error) {
             AWSDDLogError(@"Error: [%@]", task.error);
-            if ([task.error.domain isEqualToString:NSURLErrorDomain]) {
```
minbi: Is there a reason you removed this check?

jgavris: I think we'd like to surface any errors, not just request related errors.

jgavris:
@minbi Are you ok with this now?

jgavris:
Hi @minbi 👋  ping?

jgavris:
I've rebased this on `master`, could you have a look?

jgavris:
@rohandubal would you mind reviewing this?

AWSKinesis/AWSFirehoseRecorder.m
```diff
@@ -177,8 +177,9 @@ - (AWSTask *)submitRecordsForStream:(NSString *)streamName
...
+            if (task.error) {
```
rohandubal: The check for NSURLErrorDomain was due to a bug where the call went to an infinite loop. We should be setting the stop variable only when the client goes offline. Removing the check would make the process stop on first error (even if its transient and should not be stopped)

jgavris: The error that is thrown in my test is actually `AWSCognitoIdentityErrorDomain`, so we might need to at least also catch that and stop the task early, because that will not recover when online again.

jgavris:
@rohandubal I've updated as per your feedback. Added one more error domain so that the test passes.

AWSKinesisTests/AWSFirehoseRecorderTests.m
```diff
@@ -191,23 +191,35 @@ - (void)testDiskAgeLimit {
...
-    for (int32_t i = 0; i < 1234; i++) {
-        [tasks addObject:[firehoseRecorder saveRecord:[[NSString stringWithFormat:@"TestString-%02d\n", i] dataUsingEncoding:NSUTF8StringEncoding]
-                                           streamName:AWSFirehoseRecorderTestStream]];
-    }
-    
-    [[[[AWSTask taskForCompletionOfAllTasks:tasks] continueWithSuccessBlock:^id(AWSTask *task) {
-        sleep(10);
-        return [firehoseRecorder submitAllRecords];
-    }] continueWithBlock:^id(AWSTask *task) {
-        XCTAssertNil(task.error);
```
jgavris: I'm testing `saveRecord` in the below test, and because the default firehose recorder has an invalid pool id, this correctly throws an error.

rohandubal:
Reviewing the changes...

jgavris:
Thanks for the review!

PR #733, opened 2017-09-20
Add Carthage doco for AWSUserPoolsSignIn

sarah-j-smith:
Carthage support does not work currently for AWSUserPoolsSignIn because the app crashes with an error that `SignIn.storyboard` was not found.  

Carthage has this policy:

_Carthage builds framework binaries using xcodebuild, but leaves the responsibility of integrating them up to the user. CocoaPods’ approach is easier to use, while Carthage’s is flexible and unintrusive._

So it's reasonable to expect manual steps.  The doc I've added describes these.


kvasukib:
Hi @sarah-j-smith,

Thanks for submitting a pull request. We are looking into it.

README.md
```diff
@@ -129,6 +129,21 @@ You should use one of these three ways to import the AWS Mobile SDK but not mult
         Input Files: Empty
         Output Files: Empty
 
+1. If using `AWSUserPoolsSignIn` from Carthage add:
+
```
kvasukib: Can you be more specific one where to add the following three frameworks?

kvasukib: Can you be more specific one where to add the following three frameworks?

PR #766, opened 2017-11-19
Distinguish keychain for different credential providers

mark2b:
Distinguish keychain place for different credential providers by deriving from AWSCognitoCredentialsProvider

Using different Cognito credential providers  (authenticated and unauthenticated) for same identity pool causes to interference. All providers shares same keychain in order to cache credentials.


rohandubal:
Since we are using IdentityId as a identifier, shouldn't it be store in different place for auth/ un-auth identities? Since auth and un-auth identities are not same?

PR #776, opened 2017-12-08
Set AWSMobileClient deployment target to 10.3 to support armv7

### LINKED ISSUE #775: AWSMobileClient target is not compatible with armv7
Archiving the AWS SDK 2.6.8 with a deployment target < iOS 11.0 results a linker error:

```
Undefined symbols for architecture armv7:
"_OBJC_CLASS_$_AWSMobileClient", referenced from:
...
ld: symbol(s) not found for architecture armv7
```

This is because the AWSMobileClient target has set it's deployment target to iOS 11.0 which only supports arm64 architecture.

FlorianMielke:
Fix #775


FlorianMielke:
Fixed in 2.6.9 (https://github.com/aws/aws-sdk-ios/releases/tag/2.6.9)

PR #792, opened 2017-12-28
Kinesis: avoid relying on partition_key as row id in AWSAbstractKinesisRecorder

igor-makarov:
I found a bug in AWSAbstractKinesisRecorder where it uses `partition_key` as the key when deleting successfully transferred records. 

However, due to the ability to specify arbitrary partition keys per record, they are now not guaranteed to be random UUIDs. This, paired with multiple concurrent writes, results in records being deleted before sending.

I solved the problem by using `rowid` which is a built-in SQLite column guaranteed to be unique.

### DIFF
--- a/AWSKinesis/AWSAbstractKinesisRecorder.m
+++ b/AWSKinesis/AWSAbstractKinesisRecorder.m
@@ -238,7 +238,7 @@ NSString *const AWSKinesisAbstractClientRecorderDatabasePathPrefix = @"com/amazo
 
         do {
             [databaseQueue inTransaction:^(AWSFMDatabase *db, BOOL *rollback) {
-                NSMutableArray *partitionKeys = nil;
+                NSMutableArray *rowIds = nil;
 
                 AWSFMResultSet *rs = [db executeQuery:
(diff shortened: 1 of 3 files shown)


AWSKinesis/AWSAbstractKinesisRecorder.m
```diff
@@ -263,7 +263,7 @@ - (AWSTask *)submitAllRecords {
                                                   @"stream_name": [rs stringForColumn:@"stream_name"],
                                                   }];
 
-                    [partitionKeys addObject:[rs stringForColumn:@"partition_key"]];
+                    [partitionKeys addObject:[rs stringForColumn:@"rowid"]];
```
rohandubal: The changes look great. Before we can accept this merge, could you re-name the object appropriately so that the flow makes more sense? `partionKeys` should change to `rowIds` ideally.

igor-makarov:
All done! @rohandubal

igor-makarov:
Hi, can we merge this? I've made the requested changes.

igor-makarov:
Can we merge this? The changes were made.

kvasukib:
Merging the changes. The next revision of the SDK will include these changes.

Issue #490, opened 2016-10-07
AWSS3TransferUtility crashes when task.response is nil


ScottRobbins:
### We are receiving crashes on line 672 of `AWSS3TransferUtility.m`.

Stack trace:

```
* thread #40: tid = 0x13466, 0x235bc156 libsystem_c.dylib`__abort + 134, queue = 'NSOperationQueue 0x1b11bbf0 :: NSOperation 0x160e6130 (QOS: LEGACY)', stop reason = EXC_BREAKPOINT (code=EXC_ARM_BREAKPOINT, subcode=0xdefe)
    frame #0: 0x235bc156 libsystem_c.dylib`__abort + 134
    frame #1: 0x235bc0d0 libsystem_c.dylib`abort + 120
    frame #2: 0x231327dc libc++abi.dylib`abort_message + 108
    frame #3: 0x2314b6a0 libc++abi.dylib`default_terminate_handler() + 268
    frame #4: 0x23157098 libobjc.A.dylib`_objc_terminate() + 192
    frame #5: 0x23148e16 libc++abi.dylib`std::__terminate(void (*)()) + 78
    frame #6: 0x231485f4 libc++abi.dylib`__cxa_throw + 112
    frame #7: 0x23156ed2 libobjc.A.dylib`objc_exception_throw + 250
    frame #8: 0x23999a50 CoreFoundation`+[NSException raise:format:] + 112
  * frame #9: 0x014fc83c AWSS3`-[AWSS3TransferUtility URLSession:task:didCompleteWithError:](self=0x1b1a6410, _cmd="URLSession:task:didCompleteWithError:", session=0x1b1465e0, task=0x1b3dffb0, error=0x00000000) + 348 at AWSS3TransferUtility.m:672
    frame #10: 0x23f866fc CFNetwork`__51-[NSURLSession delegate_task:didCompleteWithError:]_block_invoke170 + 76
    frame #11: 0x241ae84c Foundation`__NSBLOCKOPERATION_IS_CALLING_OUT_TO_A_BLOCK__ + 8
    frame #12: 0x2410f87e Foundation`-[NSBlockOperation main] + 146
    frame #13: 0x24101f8e Foundation`-[__NSOperationInternal _start:] + 774
    frame #14: 0x241b0b0c Foundation`__NSOQSchedule_f + 192
    frame #15: 0x01db04aa libdispatch.dylib`_dispatch_queue_drain + 2014
[...]

PR #799, opened 2018-01-09
Avoid throwing an exception in production

jksk:
This replaces the exception with an assertion and assigns an error. The assertion will also raise an exception, but can be disabled in build settings. This enables us to maintain existing behaviour during development, but avoid exceptions in production.

This is regarding issue #490.


kvasukib:
Hi @jksk, Thanks for submitting a PR. We are looking into it.

PR #827, opened 2018-02-09
Declare framework dependencies in podspecs

jgavris:
Users may want to build the AWS SDK with Clang modules disabled `-fno-modules` because they are using `ccache` which doesn't support `-fmodules`. When modules are disabled, auto-linking of framework modules is also disabled, so these dependencies need to be expressed explicitly.

### DIFF
--- a/AWSCore.podspec
+++ b/AWSCore.podspec
@@ -12,7 +12,7 @@ Pod::Spec.new do |s|
   s.platform     = :ios, '8.0'
   s.source       = { :git => 'https://github.com/aws/aws-sdk-ios.git',
                      :tag => s.version}
-  s.frameworks   = 'UIKit', 'Foundation', 'SystemConfiguration'
+  s.frameworks   = 'CoreGraphics', 'UIKit', 'Foundation', 'SystemConfiguration', 'Security'
   s.libraries    = 'z', 'sqlite3'
   s.requires_arc = true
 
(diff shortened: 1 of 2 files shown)


AWSCore.podspec
```diff
@@ -12,7 +12,7 @@ Pod::Spec.new do |s|
   s.platform     = :ios, '8.0'
   s.source       = { :git => 'https://github.com/aws/aws-sdk-ios.git',
                      :tag => s.version}
-  s.frameworks   = 'UIKit', 'Foundation', 'SystemConfiguration'
+  s.frameworks   = 'CoreGraphics', 'UIKit', 'Foundation', 'SystemConfiguration', 'Security'
```
rohandubal: Could you please explain why do we need CoreGraphics in here?

jgavris: There is some (very old) code in `AWSDDTTYLogger.m` that calls some CoreGraphics methods. UIKit includes the declarations, but not the definitions, so we need to link CoreGraphics as well.

https://github.com/aws/aws-sdk-ios/blob/a50e487cc71fc9345bdd55b8bdfe0eb08c7250ac/AWSCore/Logging/AWSDDTTYLogger.m#L696

jgavris: If you'd like to try this out for yourself, in this project, turn off 'Enable Modules' build setting and try to run the unit tests. You'll need to add all these frameworks in the link build phase.

jgavris: Is there anything else I can do to get this over the line?

jgavris: Hi, friendly bump here

rohandubal:
Thanks for the clarifications! I will merge this with the next release!

jgavris:
@kvasukib could this get into the next release please?

rohandubal:
Hello @jgavris,

This will go out with next release. I am merging the changes.

Thanks,
Rohan

PR #867, opened 2018-04-04
Fix AWSS3 warnings with Xcode 9.3

Coeur:
Warnings are of the type:
> /Users/Coeur/Development/Example/Pods/AWSS3/AWSS3/AWSS3TransferUtility.m:527:95: Block implicitly retains 'self'; explicitly mention 'self' to indicate this is intended behavior


Coeur:
Superseded by #868

PR #887, opened 2018-04-11
Fix analyzer warning "Converting a pointer value of type 'NSNumber *' to a primitive boolean value"

Coeur:


AWSPinpoint/AWSPinpointEndpointProfile.m
```diff
@@ -410,8 +410,8 @@ - (NSString*) description {
             "\"City\" : %@,"
             "\"Region\" : %@,"
             "\"Country\" : %@}",
-            (self.latitude)?self.latitude:@"0",
-            (self.longitude)?self.longitude:@"0",
+            self.latitude?:@0,
```
rohandubal: I believe the latitude and longitude are intended to be strings and not numbers?

Coeur: They are NSNumber according to AWSPinpointEndpointProfile.h, so it's best to keep the same type. (Not that it matters much here, we are in a `description` function.)

rohandubal: Makes sense. Thanks for the clarification!

PR #888, opened 2018-04-11
fix analyzer memory leak

Coeur:


AWSIoT/AWSIoTManager.m
```diff
@@ -224,6 +224,8 @@ + (BOOL)importIdentityFromPKCS12Data:(NSData *)pkcs12Data passPhrase:(NSString *
 
     if (![AWSIoTKeychain addPrivateKeyRef:privateKey tag:privateTag])
     {
+        if (publicKey)
```
rohandubal: Could you please explain why is the public key here? should we release privateKey here?

Coeur: The revealed leak by Xcode clang analyzer is that *publicKey* is not released before we return at this point.

I assume that Private key ownership is given to `[AWSIoTKeychain addPrivateKeyRef:tag:]`, so it would be that method that need to release it if needed.

Coeur:
@rohandubal what about this: I remove the code change I did on AWSIoTManager.m, and I only keep the fix for AWSSignature.m? This way we can at least approve one fix out of two?

PR #903, opened 2018-04-21
Ensure that only one instance of getSession runs at a time

johnflanagan-spok:
Calls to getSession are not thread safe. Specifically they share
getSessionBlock which can cause the same completion block to be
called multiple times.

This change ensures that only once call to getSession runs at a time.


behrooziAWS:
Thanks! I mistakenly thought NSOperationQueue was serial by default.

PR #915, opened 2018-05-01
Fixes 2 incorrect clang pragmas

allenhumphreys:
I believe theses pragmas were meant to suppress the warnings being generated but instead disabled warning generation for diagnostics that weren't actually producing warnings. This updates fixes the warnings produced when using AWSCore with cocoapods.


rohandubal:
Thanks for the PR! I will merge this with next release!

PR #917, opened 2018-05-03
Updated sign up UI to accept and format different phone number formats

hatboyzero:
Replaced logic that simply checked to see if the phone number was in E164 format with implementation that should support most standard phone number inputs.


kvasukib:
Thank you @hatboyzero for submitting a pull request. We are looking into it.

kvasukib:
@hatboyzero Can you give us some numbers on the increase in the size of the artifact of the library or the app when you added this dependency?

vccabral:
@hatboyzero & @kaustavghosh06 this is a feature I would like as well. Any progress?

I also noticed the forgot username and sign up nav controller doesn't have a back button when I was testing. I have a few other concerns about usability that I need to integrate into my app which are making me move away from the pre-built auth ui you all provide. Any chance that there are a few standard ui components you all would be interested in implementing.

 * phone number custom validation (with a string)
 * email validation with a regex or other
 * custom validators of other types
 * customizable back buttons.

rohandubal:
Hi, thanks for the PR! Currently we will not be able to merge this PR. But I am referencing this as part of #1158 so that we can include it as part of planning in our next set of tasks.

PR #929, opened 2018-05-16
Convenience function to force new auth tokens.

Abeansits:
**Why?**
We are building an iOS app that allows a user to gain more access by making a purchase. We have architected this using Cognito User Pools. 
When a user signs up and makes a purchase we transfer them from one group top another. But in order to gain access to the endpoints we need to refresh the auth tokens. Fetching and refreshing the tokens are done calling: `getSession` in `AWSCognitoIdentityUser`.
There is logic in `getSession` that will only allow for the tokens to be refreshed if they have expired or are about to expire.

**What?**
I've added another function that will allow for forcing the refresh of the auth tokens even if they are valid.

### DIFF
--- a/AWSCognitoIdentityProvider/AWSCognitoIdentityUser.h
+++ b/AWSCognitoIdentityProvider/AWSCognitoIdentityUser.h
@@ -107,6 +107,11 @@ NS_ASSUME_NONNULL_BEGIN
 - (AWSTask<AWSCognitoIdentityUserSession *> *)getSession;
 
 /**
+ Get a session with id, access and refresh tokens. Force the refreshing of authentication tokens by passing in the `forceRefresh` flag.
+ */
+-(AWSTask<AWSCognitoIdentityUserSession*> *) getSessionAndForceTokenRefresh: (BOOL)forceRefresh;
+
+/**
  Get a session with the following username and password
  */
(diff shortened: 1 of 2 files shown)


cbommas:
hello @Abeansits 

Thank you for submitting the PR. We will review internally and failing any other questions, will merge into our next release.

behrooziAWS:
Hi Abeansits, I've looked over the PR.  We need to do something similar if a user updates attributes from the sdk, so I agree this is a valid use case.  I think a better way of accomplishing this is to have a method that clears the id and access token from the keychain called clearSession so it could be used as part of update attributes too.  With this approach, do you think having two calls, one to clearSession and then a call to getSession is ok for your use case?

Abeansits:
Hi @behrooziAWS that sounds like that would work.

Abeansits:
Thank you my friend!

Abeansits:
Any news on this guy?

behrooziAWS:
Should be included in the next release, sorry for the long wait here.  There will be a method called clearSession on the user and updateAttributes will automatically call clearSession to force a token refresh.

behrooziAWS:
Ok, it's released.  https://github.com/aws/aws-sdk-ios/blob/master/CHANGELOG.md#2622

Abeansits:
Thank you!

PR #964, opened 2018-06-19
Correct dictionary key for bucket name value

OdNairy:
There is outdated key for dictionary values for bucket name.
Rename from `bucket` to `bucket_name`


kvasukib:
Thank you @OdNairy for submitting a PR. We will merge it in the upcoming release.

PR #986, opened 2018-07-08
Feature/add sign in uri parameter

sria-shoji:
Thank you for providing a great SDK.
The purpose of this change is to make Cognito's authentication page customizable.
In the current source code, the URL of the authentication page is generated based on the parameter "CognitoAuthWebDomain" in the plist file.
I would like to create an authentication page in a domain different from "CognitoAuthWebDomain" and display it in the in-app browser. For that purpose, we added a new parameter "CognitoAuthSignInUri" to the plist file. By setting the URL of the authentication page to this parameter, the custom authentication page is displayed.
Since I need this change, I would appreciate it if you merge the pull request. Thank you for your consideration.

### DIFF
--- a/AWSCognitoAuth/AWSCognitoAuth.h
+++ b/AWSCognitoAuth/AWSCognitoAuth.h
@@ -205,7 +205,6 @@ typedef void (^AWSCognitoAuthSignOutBlock)(NSError * _Nullable error);
  @param scopes Set of scopes to obtain
  @param signInRedirectUri uri to redirect on sign in.  Must be configured as a uri scheme in your info.plist
  @param signOutRedirectUri uri to redirect on sign out.  Must be configured as a uri scheme in your info.plist
- @param signInUri uri on sign in.  Must be configured as a uri scheme in your info.plist
  @param webDomain The FQDN of your Cognito endpoint, something like https://mydomain.region.auth.amazoncognito.com
  */
 - (instancetype)initWithAppClientId:(NSString *) appClientId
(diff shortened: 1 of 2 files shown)


minbi:
Hi @sria-shoji ,

Thanks for the pull request. The change as it is cannot be accepted because it would break existing customers by changing the method signature.

If I understand correctly, this change will support custom domains that do not follow the `<webdomain>/oauth2/authorize` which means `<webdomain>/otherpath` will also work after hte change?

You will need to create a separate init method without the `webDomain` variable to support this case.

sria-shoji:
Hi @minbi,
Thank you for confirming the changes and comments.

As you say, given the impact on other customers, changes to the init method were incorrect.

> If I understand correctly, this change will support custom domains that do not follow the <webdomain> / oauth 2 / authorize which means <webdomain> / otherpath will also work after hte change?

I think so.
However, as before, <webdomain> will be used for getting the token.
The purpose is to display a page of custom URL which is not "<webdomain> / oauth2 / authorize" when signing in.

I fixed the following points in the source code:
 1. I replaced the signatures of the two init methods that affect others. However, these two methods call the newly added init method.
 2. I added a new init method so that it accepts arguments of <signInUri>.

This will work as before if there is no <CognitoAuthSignInUri> key in info.plist, and if there is a key, use the value of this key to display the ViewController for signing in.

Could you please accept this?

sria-shoji:
How about this matter?
Thank you for your answer.

PR #992, opened 2018-07-12
AWSIoTMQTTClient: fix connection age timer not scheduled

mattczech:
Currently, the `connectionAgeTimer` in `AWSIoTMQTTClient` gets initialized, but never scheduled. This results in the exponential backoff timer increasing for each disconnect and never being reset to the base reconnect time.


cbommas:
@mattczech 

Thank you for submitting this PR. I'll review/run tests and get back to you.

PR #1000, opened 2018-07-19
User completion code called after SFSafariViewController dismissed

### LINKED ISSUE #998: AWSCognitoAuth sign-out followed by sign-in broken - SFSafariViewController remains
* What service are you using?
AWSCognitoAuth
* In what version of SDK are you facing the problem?
2.6.24
* Is the issue limited to Simulators / Actual Devices?
No
* Can your problem be resolved if you bump to a higher version of SDK?
No
* Is this problem related to specific iOS version?
No
* How are you consuming the SDK? CocoaPods / Carthage / Prebuilt frameworks?
CocoaPods
* Can you give us steps to reproduce with a minimal, complete, and verifiable example? Please include any specific network conditions that might be required to reproduce the problem.

        self.auth.signOut { error in
            if error == nil {
                self.auth.getSession { (session, error) in
                    if error == nil {
                        print("Success!")
                    } else {
                        print("Sign-In failure: " + ((error! as NSError).userInfo["error"] as! String))
                    }
                }
            } else {
                print("Sign-Out failure: " + ((error! as NSError).userInfo["error"] as! String))
            }
        }

Suggestion: Reimplement the [signOutTapped method](https://github.com/awslabs/aws-sdk-ios-samples/blob/master/CognitoAuth-Sample/Swift/CognitoAuthSample/ViewController.swift#L117) with the above snippet for a quick repro.

Prompting the user to sign-in immediately after sign-out is a [common pattern](https://docs.aws.amazon.com/aws-mobile/latest/developerguide/how-to-user-sign-in-sign-out.html). However, the code above fails with `Sign-In failure: Loading page failed`. Subsequently, the user is able to sign-in, but the sign-in modal does not dismiss itself as it should.

Suggestion: Place breakpoints on the following lines:
1. launchSignInVC - [`self.svc = [[SFSafariViewController alloc] initWithURL:[NSURL URLWithString:url] entersReaderIfAvailable:NO];`](https://github.com/aws/aws-sdk-ios/blob/2.6.24/AWSCognitoAuth/AWSCognitoAuth.m#L226)
2. completeSignOut - [`self.signOutBlock(error);`](https://github.com/aws/aws-sdk-ios/blob/2.6.24/AWSCognitoAuth/AWSCognitoAuth.m#L336)
3. signOutInternal - [`self.svc = [[SFSafariViewController alloc] initWithURL:[NSURL URLWithString:url] entersReaderIfAvailable:NO];`](https://github.com/aws/aws-sdk-ios/blob/2.6.24/AWSCognitoAuth/AWSCognitoAuth.m#L367)
4. dismissSafariVC - [`if(self.svc){`](https://github.com/aws/aws-sdk-ios/blob/2.6.24/AWSCognitoAuth/AWSCognitoAuth.m#L588)
5. dismissSafariVC - [`self.svc = nil;`](https://github.com/aws/aws-sdk-ios/blob/2.6.24/AWSCognitoAuth/AWSCognitoAuth.m#L591)

Running the above code with the given breakpoints results in the following order:
1. signOutInternal (set svc)
2. completeSignOut (call user's signOutBlock)
3. dismissSafariVC (check svc)
4. launchSignInVC (set svc)
5. print "Sign-In failure"
6. dismissSafariVC (check svc)
7. dismissSafariVC (unset svc)

As you can see, the svc variable gets corrupted at step 4. Also, the modal isn't dismissed completely until step 7, well after the user's code is executed in step 2. So, here is the order that would prevent corruption and would avoid the annoying presence of a modal/SFSafariViewController when the user's completeSignOut code is executed:
1. signOutInternal (set svc)
2. dismissSafariVC (check svc) [previously step 3]
3. dismissSafariVC (unset svc) [previously step 7]
4. completeSignOut (call user's signOutBlock) [previously step 2]
5. launchSignInVC (set svc) [previously step 4]
6. dismissSafariVC (check svc) [new]
7. dismissSafariVC (unset svc) [new]
8. completeGetSession (call user's getSessionBlock) [new]
9. print "Success" [new]

Also, the implementation seems a little more complicated than necessary: I believe `dismissSafariVC` is only ever called from the main thread, so its asynchronous main thread call (i.e. `dispatch_async(dispatch_get_main_queue()`) might be unnecessary.

EthanLozano:
Fixes #998
Resets the user's application view and all AWSCognitoAuth internal state
before calling the user's completion block during signOut and getSession.
As a result, the SFSafariViewController won't be displayed and internal
variables will have null values when the user's code executes.

### DIFF
--- a/AWSCognitoAuth/AWSCognitoAuth.m
+++ b/AWSCognitoAuth/AWSCognitoAuth.m
@@ -313,14 +313,10 @@ static NSString * AWSCognitoAuthAsfDeviceId = @"asf.device.id";
         self.getSessionError = error;
         [self.getSessionQueue cancelAllOperations];
     }
-    if(self.svc){
-        [self.svc dismissViewControllerAnimated:NO completion:^{
-            self.svc = nil;
-            [self cleanUpAndCallGetSessionBlock:userSession error:error];
-        }];
-    }else {
+    
+    [self dismissSafariVC: ^{
         [self cleanUpAndCallGetSessionBlock:userSession error:error];
-    }
+    }];
 }
 
 - (void) cleanUpAndCallGetSessionBlock: (nullable AWSCognitoAuthUserSession *) userSession  error:(nullable NSError *) error {
@@ -341,14 +337,10 @@ static NSString * AWSCognitoAuthAsfDeviceId = @"asf.device.id";
         self.signOutError = error;
         [self.signOutQueue cancelAllOperations];
     }
-    if(self.svc){
-        [self.svc dismissViewControllerAnimated:NO completion:^{
-            self.svc = nil;
-            [self cleanUpAndCallSignOutBlock:error];
-        }];
-    } else {
+    
+    [self dismissSafariVC: ^{
         [self cleanUpAndCallSignOutBlock:error];
-    }
+    }];
 }
 
 - (void) cleanUpAndCallSignOutBlock:(nullable NSError *) error {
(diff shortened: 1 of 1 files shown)


EthanLozano:
@kvasukib have you had a chance to look this over? Do you want me to change anything to get this approved? I'd like to get this fix since the AWSCognitoAuthDelegate and non-dismissed modals are giving me trouble.

EthanLozano:
ping

behrooziAWS:
@EthanLozano Hi, I will review this, thanks for the contribution.

PR #1007, opened 2018-07-26
Optional method added to the AWSCognitoAuthDelegate protocol and the AWSCognitoAuthClientErrorExpiredRefreshToken error case.

lucianboboc:
…AuthDelegate. It will let the user decide if `launchSignInVC` is called when the refresh token is expired or not.


lucianboboc:
Any chance to get an answer on this?

behrooziAWS:
Is there any reason that you would want to change this during program execution or it yes or no based on your app.  If it is yes or no and you know it at application startup, it would make more sense as a property in AWSCognitoAuthConfiguration.  Let me know what you think.

lucianboboc:
Thank you for your reply.
It is during the app execution. I think it is a bad experience for the user (that’s logged in) at some point to just present a SafariVC to ask him to log back in with google.
Being optional it doesn’t affect the current framework behavior, we will just have the option to decide what to do.

behrooziAWS:
Ok, code looks good then, I'll let the maintainer know to merge.  Thanks!

lucianboboc:
Good news, thank you 😃

PR #1037, opened 2018-09-11
Fix memory leaks for IoT

jkennington:
After removing an `AWSIoTDataManager`, the `AWSIoTDataManager`, `AWSIoTMQTTClient`, `AWSURLSessionManager` and other related entities were leaked.

This fixes a retain cycle between the `AWSIoTMQTTClient` where it held a strong reference to the `associatedObject` the `AWSIoTDataManager` which was holding a strong reference to the `mqttClient`. This makes the `associatedObject` reference weak.

If an `AWSIoTDataManager` was disconnected before the `connectionAgeInSeconds` exceeded the `minimumConnectionTime` the `connectionAgeTimer` was never invalidated and held a strong reference to the `AWSIoTMQTTClient` (e.g. `self` passed to the timer on creation of the timer). This fixes the issue by issuing an `invalidate` to the timer on the correct thread in the event the user requests a disconnect before reaching the `minimumConnectionTime`.

### DIFF
--- a/AWSIoT/Internal/AWSIoTMQTTClient.m
+++ b/AWSIoT/Internal/AWSIoTMQTTClient.m
@@ -639,6 +639,16 @@ static const NSString *SDK_VERSION = @"2.6.19";
     [defaultRunLoopTimer invalidate];
     
     if (!self.runLoopShouldContinue ) {
+        // Add a block to invalidated the connectionAgeTimer in order to release the strong reference to self
+        // and run the run loop another cycle to give the block a chance the execute.
+        [runLoopForStreamsThread performBlock:^{
+            if (self.connectionAgeTimer != nil ) {
+                [self.connectionAgeTimer invalidate];
+                self.connectionAgeTimer = nil;
+            }
+        }];
+        [runLoopForStreamsThread runMode:NSDefaultRunLoopMode beforeDate:[NSDate dateWithTimeIntervalSinceNow:10]];
+        
         [self.session close];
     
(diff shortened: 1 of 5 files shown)


AWSIoT/Internal/AWSIoTMQTTClient.m
```diff
@@ -639,6 +639,16 @@ - (void)openStreams:(id)sender
     [defaultRunLoopTimer invalidate];
     
     if (!self.runLoopShouldContinue ) {
```
cbommas: Good catch. The connectionAgeTimer will self terminate once the max age has been reached, but if a disconnect is issued before then, it will continue to run and hold on to the object reference. 

I think we can simplify this by just doing this after line 639. Thoughts?

         if (self.connectionAgeTimer != nil ) {
                [self.connectionAgeTimer invalidate];
                self.connectionAgeTimer = nil;
            }

jkennington: I believe we need to move the `invalidate` call because of the documentation below. 

From https://developer.apple.com/documentation/foundation/nstimer/1415405-invalidate?language=objc 

> Special Considerations
> You must send this message from the thread on which the timer was installed. If you send this message from another thread, the input source associated with the timer may not be removed from its run loop, which could prevent the thread from exiting properly.

jkennington:
Updated PR at https://github.com/aws-amplify/aws-sdk-ios/pull/1175

PR #1044, opened 2018-09-19
New initializer to create user pool client with 'Share Access Group'  for iOS Extensions (Share, Today)

qaisershehzad:
…roup' and 'Service' to handle extension usecase where same session can be used in app extension.

New initializer to create user pool client with 'Share Access Group' and 'Service' to handle extension use case where same session can be used in app extension. 
Currently if we use CognitoUserPoolAuth it will ask for login again as session is not shared to extension  
with Share Access Group for keychain.

This is how we can initialize User Pool

// initialize user pool client
        AWSCognitoIdentityUserPool.register(with:serviceConfiguration, userPoolConfiguration: poolConfiguration, forKey: AWSCognitoUserPoolsSignInProviderKey, forService: "<Add your Service Configuration Id>", forAccessGroup: "<Add your accessGroup>")

Above code is added in reference to the following issues : 
https://github.com/awslabs/aws-sdk-ios-samples/issues/301
https://github.com/aws-amplify/aws-sdk-ios/issues/1106

### DIFF
--- a/AWSCognitoIdentityProvider/AWSCognitoIdentityUserPool.h
+++ b/AWSCognitoIdentityProvider/AWSCognitoIdentityUserPool.h
@@ -66,6 +66,12 @@ NS_ASSUME_NONNULL_BEGIN
                                    userPoolConfiguration:(AWSCognitoIdentityUserPoolConfiguration *)userPoolConfiguration
                                                   forKey:(NSString *)key;
 
++ (void)registerCognitoIdentityUserPoolWithConfiguration:(nullable AWSServiceConfiguration *)configuration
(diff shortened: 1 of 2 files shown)


palpatim:
@nikhil-dabhade Can you review this use case and see if it's something we want to support in the SDK? If so, we'll review this PR more carefully, but otherwise, we should close it so it's not just hanging out.

Grigs-b:
👍for supporting. I'm currently using this on a project and it would be nice for it to be in an official release.

koxon:
Hello. Can we get an update on this extension issue ? We have to update the Pod code to make this work. This PR is one year old and dying here ... 
AWS can we expect something on this front ?

Thank you

PR #1068, opened 2018-10-23
JWT token claims values not only strings

MichK1:
When trying to get JWT values which are not Strings application written in Swift crashes.

### DIFF
--- a/AWSCognitoIdentityProvider/AWSCognitoIdentityUser.h
+++ b/AWSCognitoIdentityProvider/AWSCognitoIdentityUser.h
@@ -272,7 +272,7 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  A NSDictionary of claims in this token.
  */
-@property (nonatomic, readonly) NSDictionary<NSString *, NSString*> * claims;
+@property (nonatomic, readonly) NSDictionary<NSString *, id> * claims;
 
 @end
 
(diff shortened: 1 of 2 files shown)


lucianboboc:
I had the same crash today for accessing the `claims` from swift and if-let binding to `[String:Any]` is not working. Please accept this fix asap 🙏

palpatim:
The change makes sense on the face of it, but I'm going to dive a bit deeper to make sure there's no implications I'm missing. Assuming the existing tests pass, I'd also like to add some Swift unit tests as well to ensure the correct deserialization behavior. I'm working on one locally--if I'm able to validate this change locally we can merge it into `master` and follow up on the next official release with expanded test coverage.

palpatim:
Hi @MichK1 and @lucianboboc,

Thanks for bringing this to our attention. Unfortunately, the proposed fix is a breaking change--callers who are currently relying on the property to be a `[String : String]` would have to add an explicit cast. Instead, I added your fix onto a new property, `tokenClaims` and deprecated `claims`. At some point after that, we can re-institute `claims` with the correct signature and retire the unfortunately-named `tokenClaims`. :)

We'll have that pushed in an upcoming release.

palpatim:
This is fixed by using the `tokenClaims` property introduced in 2.7.1.

MichK1:
Thanks guys for addressing this. :+1: 🚀

PR #1099, opened 2018-11-11
Change the xcode version for release carthage job

sunchunqiang:
Change the xcode version from 10.0.0 to 9.2.0 for release carthage job to fix the carthage build timeout issue

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ jobs:
   release_carthage:
     # Specify the Xcode version to use
     macos:
-      xcode: "10.0.0"
+      xcode: "9.2.0"
 
     steps: 
       - checkout
--- a/CreateCarthageArchive.sh
+++ b/CreateCarthageArchive.sh
@@ -1,6 +1,6 @@
-# carthage archive | tee "carthageout.txt"
-frameworkfilename=$(carthage archive |  grep -o 'Created .*\.zip$'  | grep -o   '\S*\.zip$')
-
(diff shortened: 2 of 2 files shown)


CreateCarthageArchive.sh
```diff
@@ -1,6 +1,6 @@
-# carthage archive | tee "carthageout.txt"
-frameworkfilename=$(carthage archive |  grep -o 'Created .*\.zip$'  | grep -o   '\S*\.zip$')
-
+carthage archive | tee "carthageout.txt"
+ls *.zip
```
muellerfr: why are you running "ls" here?

.circleci/config.yml
```diff
@@ -8,7 +8,7 @@ jobs:
   release_carthage:
     # Specify the Xcode version to use
     macos:
-      xcode: "10.0.0"
+      xcode: "9.2.0"
```
muellerfr: How does changing the xcode version help?

palpatim: Xcode 9.3 [isn't packaged with `docsetutil`](https://github.com/tomaz/appledoc/issues/628), which we use for generating appledoc. We decided to force-downgrade the Xcode version we're building with.

rohandubal: @palpatim We already made the change for that for our docs workflow, but this one would also use the older xcode for our Carthage builds.

CreateCarthageArchive.sh
rohandubal: I believe this is for debugging purposes.

Issue #857, opened 2018-03-23
AWSCognitoAuth signIn error


serddmit:
### To help us solve your problem better, please answer the following list of questions.

* What service are you using?
  AWSCognitoAuth

* In what version of SDK are you facing the problem?
  SDK 11.2

* Is the issue limited to Simulators / Actual Devices?
  NO

* Can your problem be resolved if you bump to a higher version of SDK?
  NO

* Is this problem related to specific iOS version?
  NO

* How are you consuming the SDK? CocoaPods / Carthage / Prebuilt frameworks?
  Cocoapods

* Can you give us steps to reproduce with a minimal, complete, and verifiable example? Please include any specific network conditions that might be required to reproduce the problem.

SignIn with Google provider. SignOut. Sign in with Google provide -> Error occurs.

When I had authorized previously and made sign out after, my next sign In will display "Loading page failed". This happens every second login-logout pair.

This happens only for Google provider.
The error is generated in `- (void)safariViewController:(SFSafariViewController *)controller didCompleteInitialLoad:(BOOL)didLoadSuccessfully` with `didLoadSuccessfully = NO`.

So this will generate error and not wait for NSURLSession callback.

Does anyone has such problem?Thanks

kainoffke:
I can confirm this issue. I am currently facing the same problems. As a workaround I created an extension for `AWSCognitoAuth` like so:

```
extension AWSCognitoAuth : SFSafariViewControllerDelegate {
   public func safariViewController(_ controller: SFSafariViewController, didCompleteInitialLoad didLoadSuccessfully: Bool) {
        // I was just implemented do stop AWSCognitoAuth calling its own delegate implementation.
    }
}
```

The problem seems to be, that the redirects in the SafariViewController are happening too fast for it to actually load the page initially.

serddmit:
I end up by removing this method at all:
`public func safariViewController(_ controller: SFSafariViewController, didCompleteInitialLoad didLoadSuccessfully: Bool) {
        // I was just implemented do stop AWSCognitoAuth calling its own delegate implementation.
    }`
This will not impact functionality, but leave user to make done/cancel by itself.

dluquedavid:
problem only happen when yo set "identity_provider=google", the redirect fail.
We fixed it with @kainoffke solution.

PR #1110, opened 2018-11-14
Fix SafariViewController compatibility issue.

zhentan:
safariViewController:didCompleteInitialLoad: will always return NO for didLoadSuccessfully when the webUI immediately calls the redirect url. (ie Facebook does this when user is already logged in.) This fix prevents the SFSafarVC error handling blocks getting called which will prevent with the oauth2/token success callback being called.

This behavior has been reported in the past by other when Google is used as Identity Provider: https://github.com/aws-amplify/aws-sdk-ios/issues/857

### DIFF
--- a/AWSCognitoAuth/AWSCognitoAuth.m
+++ b/AWSCognitoAuth/AWSCognitoAuth.m
@@ -40,6 +40,7 @@ NSString *const AWSCognitoAuthErrorDomain = @"com.amazon.cognito.AWSCognitoAuthE
 
 @property (nonatomic) BOOL useSFAuthenticationSession;
 @property (nonatomic) BOOL sfAuthenticationSessionAvailable;
+@property (nonatomic) BOOL isHandlingRedirection;
 
 @end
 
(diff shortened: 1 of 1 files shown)


AWSCognitoAuth/AWSCognitoAuth.m
```diff
@@ -589,6 +590,7 @@ - (NSString *) urlEncode: (NSString *) stringToEncode {
 }
     
 - (BOOL)application:(UIApplication *)app openURL:(NSURL *)url options:(NSDictionary<UIApplicationOpenURLOptionsKey, id> *)options {
+    self.isHandlingRedirection = YES;
```
palpatim: I think this is a bit too high-level a place to be setting this flag: in cases where the `processURL:` method never reaches the point where it establishes an NSURLConnection, we would never clear the flag.

I'm going to accept the PR, and then I'll make a change for the next release to set this flag only if we reach the point where we actually establish the NSURLConnection to handle the auth token. I'll also verify my belief that `safariViewController:didCompleteInitialLoad:` will always be invoked *after* `application:openURL:options:` in redirect scenarios.

zhentan: Sounds good. I initially was going to put this flag where the NSURLConnection for the oauth2/token call. I'm pretty indifferent about it so feel free to move it there or however you see fit. The only real reason I put it where it is now is because the variable name is `isHandlingRedirection` and technically the `application:openURL:` is where that begins in this class.

PR #1114, opened 2018-11-17
Fixed S3 compatible endpoint signature issues

jlalvarez18:
Now properly checking service type when generating authorization header.

I've tested support for DigitalOcean Spaces S3 compatible API.


OkiRules:
what is the code to uplaod to digitalocean in swift?

PR #1013, opened 2018-08-08
Fix issue where the SignIn storyboard is not found

PaulHorenberger:
The previous code was looking for resources in a non-existent AWSAuthUI.bundle nested in the AWSAuthUI framework. This was causing an exception when the `[AWSSignInViewController getUIStoryboardFromBundle]` method was called. This change removes the call to append the AWSAuthUI.bundle to the path of the framework to load the storyboard and image resources.

PR #1122, opened 2018-11-24
Fix for 'Could not find a storyboard named 'AWSUserPools'

peterfennema:
After applying patch #1013 successfully I encountered a similar issue for another storyboard:

When tapping on the 'Create new account' link or the 'Forgot your password?' link in the Sign In UI my test app crashes because the AWSUserPools storyboard was not found.

This patch solves the issue, the applied solution is similar to te patch #1013


peterfennema:
@rohandubal I use Carthage

rohandubal:
Thanks for letting me know. I will validate if these changes work with both cocoapods / carthage and reply back here.

rohandubal:
Hello @peterfennema 

While this fix works for Carthage/ Frameworks, it does not for Cocoapods. I am working on a solution to support all dependencies and it will be released in a future version. Closing this PR as the changes would break Cocoapods users.

Thanks,
Rohan

rohandubal:
Hello,

This issue is resolved in `2.7.4` SDK release. Please let us know if you continue to face issues.

Thanks,
Rohan

peterfennema:
@rohandubal Tested with `2.7.4` using Carthage. Works well. Thanks a lot.

PR #1130, opened 2018-11-28
[AWSCore] Make XMLDictionary header public

keith:
This header is imported by AWSS3 even though it's private


cbommas:
@keith 

Thank you for this suggestion. This is however, by design. While the XMLDictionary is used internally from other modules, we do not want to make it publicly accessible.

PR #1132, opened 2018-11-28
[AWSS3] Remove duplicate Validation category

keith:
This fixes a warning because the Validation category was defined twice.


cbommas:
@keith 
Changes look good. This will be included in the next release of the SDK.
Thank you for your contribution.

PR #1133, opened 2018-11-28
[AWSS3] Change imports to include from module

keith:
This makes it more clear that these imports are coming from the
framework dependency.


cbommas:
Changes look good. This will be included in the next rev of the SDK.

PR #1139, opened 2018-12-07
Build AllTests target for master branch to validate builds.

rohandubal:
Currently master branch is never built on CircleCI, this will make sure all frameworks are compiled on master branch.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -116,7 +116,8 @@ workflows:
       - build:
           filters:
             branches:
-              ignore: /gh-.*/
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -97,6 +113,10 @@ workflows:
   version: 2
   build_and_test:
     jobs: 
+      - build:
+          filters:
+            branches:
+              ignore: /gh-.*/
```
minbi:
```suggestion
              only:
                  - master
```

PR #1155, opened 2018-12-29
Add ability to set keychain service and access group

larryonoff:
iOS allows to share data between "apps" / "apps and extensions", e.g. keychain data or files in folders.

AWS iOS SDK doesn't allow to share credentials between apps, but it can be easily achieved with keychain access groups.

This PR allows developers to set keychain service and access group to achieve sharing the same session between apps.

### DIFF
--- a/AWSCore/Authentication/AWSCredentialsProvider.h
+++ b/AWSCore/Authentication/AWSCredentialsProvider.h
@@ -203,6 +203,21 @@ typedef NS_ENUM(NSInteger, AWSCognitoCredentialsProviderErrorType) {
            identityProviderManager:(nullable id<AWSIdentityProviderManager>)identityProviderManager;
 
 /**
+ Initializer for credentials provider with enhanced authentication flow. This is the recommended method for first time Amazon Cognito developers. Will create an instance of `AWSEnhancedCognitoIdentityProvider`.
+ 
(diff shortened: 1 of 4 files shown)


rohandubal:
Tagging @nikhil-dabhade to get feedback whether we can integrate this.

james-ff:
+1 for this feature

vladimirz-medopad:
What is happening with this feature?

james-ff:
We could very much do with access to the current idToken of the signed in user within our share extension, does this PR cover that case?

abemusic:
Has anyone figured this out? I'm currently providing the ID token via an app group between my host app and share extension, but the token will eventually expire. This makes for a poor experience to basically tell my users they need to go open the main app and that they will need to do this once per hour if they want it to continue working.

I've also provided the refresh token in the shared app group, but there doesn't seem a way to use it directly to obtain a new set of tokens. Another option would be to store the user's credentials when they sign into the main app the first time and reuse them in the share extension....but that sort of defeats the purpose, right?

Totally out of ideas at this point besides building a login flow into the share extension and have the users sign in twice. More burden on the user though, so I'd prefer not to go this route either.

Anyone have any workarounds they can share?

HSin-Tse:
+1 for this feature
my services app use  today extensions  control many  iot devices .
really need this  feature .

PR #1163, opened 2019-01-04
change release-sdk process

sunchunqiang:
1. change workflow of release-sdk so that only release-carthage and release-cocoapods are finished, can release-appdoc job be started

2. split release-tag job from release-carthage job

3. use pipeline to input release description to github-release command instead of using parameter so that special characters can be supported in release description

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -81,6 +81,22 @@ jobs:
       - store_artifacts:
           path: integration_test_result
 
+  release_tag:
+    # Specify the Xcode version to use
+    macos:
+      xcode: "10.1.0"
+    steps: 
+      - checkout
+      - run:
+          name: install github-release
+          command: brew install github-release                
+      - run:
+          name: release the tag
+          command: |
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -179,14 +189,19 @@ jobs:
...
           command: |
             git config --local user.name "AWS"
             git checkout  gh-pages
-            mkdir docs/reference/AWSMobileClient
-            cp -R docs/api/* docs/reference/AWSMobileClient
+            if [ -d docs/reference/AWSMobileClient ]    
+            then 
+              echo "docs/reference/AWSMobileClient is present"
+            else
+                mkdir docs/reference/AWSMobileClient
```
palpatim: For future, you don't actually need the if check, assuming that the build systems have a version of `mkdir` that supports the `-p` option: `mkdir -p docs/reference/AWSMobileClient`. The command silently fails without error if the directory already exists, but makes the target with all path elements if not. E.g.:

```bash
% mkdir path/that/doesnt/exist
mkdir: path/that/doesnt: No such file or directory
% echo $?
1
% mkdir -p path/that/doesnt/exist
% echo $?
0
% mkdir -p path/that/doesnt/exist
% echo $?
0
```

sunchunqiang: great! 
I already changed according to your comment

sunchunqiang:
merged

PR #1174, opened 2019-01-12
fix brew update error

sunchunqiang:
when brew install an application, brew update will be first called, but recently brew update will fail if the environment variable HOMEBREW_LOGS is not set. In this change, we set the variable before bew update is called to prevent it from failing


palpatim:
Merged offline.

Issue #913, opened 2018-04-26
Crash [AWSURLSessionManager taskWithDelegate:] Fatal Exception: Task created in a session that has been invalidated


jschwab923:
### To help us solve your problem better, please answer the following list of questions.

* What service are you using?
AWSS3 sdk
* In what version of SDK are you facing the problem?
2.6.2-2.6.12
* Is the issue limited to Simulators / Actual Devices?
No
* Can your problem be resolved if you bump to a higher version of SDK?
Have been unable to test most recent 2.6.16 because of other issue.
* Is this problem related to specific iOS version?
Probably not
* How are you consuming the SDK? CocoaPods / Carthage / Prebuilt frameworks?
Cocoapod

I believe this happens when my app enters the foreground and has begun to use the AWSS3TransferUtility to fetch images. Unfortunately I've been unable to track down precisely what triggers it. However, I'm hoping the crash may point to something not being handled correctly in the sdk? 

It appears that the sdk is retrying a task multiple times and then eventually trying to start a new task, but on a session that has been invalidated. I have been able to recreate it all the way up to 2.6.12 but am unable to use any higher sdk due to a different issue that arises in the newer versions with multipart upload.

```
#15. Crashed: com.twitter.crashlytics.ios.exception
0  Felt Photos                    0x1026bafd4 CLSProcessRecordAllThreads + 4309282772
1  Felt Photos                    0x1026bafd4 CLSProcessRecordAllThreads + 4309282772
2  Felt Photos                    0x1026bae90 CLSProcessRecordAllThreads + 4309282448
3  Felt Photos                    0x1026aafac CLSHandler + 4309217196
4  Felt Photos                    0x1026b8fdc __CLSExceptionRecord_block_invoke + 4309274588
5  libdispatch.dylib              0x1848f0ae4 _dispatch_client_callout + 16
6  libdispatch.dylib              0x18492deb4 _dispatch_queue_barrier_sync_invoke_and_complete + 56
7  Felt Photos                    0x1026b8a70 CLSExceptionRecord + 4309273200
8  Felt Photos                    0x1026b8898 CLSExceptionRecordNSException + 4309272728
9  Felt Photos                    0x1026b84ac CLSTerminateHandler() + 4309271724
10 libc++abi.dylib                0x1841a937c std::__terminate(void (*)()) + 16
11 libc++abi.dylib                0x1841a9400 std::terminate() + 84
[...]

PR #1175, opened 2019-01-14
Fixes memory leaks in IoT

jkennington:
Similar to earlier PR with another memory leak fixed and target most recent SDK version.

After removing an `AWSIoTDataManager`, the `AWSIoTDataManager`, `AWSIoTMQTTClient`, `AWSURLSessionManager` and other related entities were leaked.

This fixes a retain cycle between the `AWSIoTMQTTClient` where it held a strong reference to the `associatedObject` the `AWSIoTDataManager` which was holding a strong reference to the `mqttClient`. This makes the `associatedObject` reference weak.

If an `AWSIoTDataManager` was disconnected before the `connectionAgeInSeconds` exceeded the `minimumConnectionTime` the `connectionAgeTimer` was never invalidated and held a strong reference to the `AWSIoTMQTTClient` (e.g. `self` passed to the timer on creation of the timer). This fixes the issue by issuing an `invalidate` to the timer on the correct thread in the event the user requests a disconnect before reaching the `minimumConnectionTime`.

The `reconnectTimer` was invalidated on whatever thread `disconnect` was called on rather than the `reconnectThread` causing a memory leak; this is fixed by having the `reconnectTimer` invalidated on the correct thread.

### DIFF
--- a/AWSIoT/Internal/AWSIoTMQTTClient.m
+++ b/AWSIoT/Internal/AWSIoTMQTTClient.m
@@ -513,7 +513,7 @@ static const NSString *SDK_VERSION = @"2.6.19";
     }
     
     //Invalidate the reconnect timer so that there are no reconnect attempts.
-    [self.reconnectTimer invalidate];
+    [self cleanupReconnectTimer];
     
     //Set the userDisconnect flag to true to indicate that the user has initiated the disconnect.
     self.userDidIssueDisconnect = YES;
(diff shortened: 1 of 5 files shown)


AWSIoT/Internal/AWSIoTMQTTClient.m
```diff
@@ -639,6 +672,12 @@ - (void)openStreams:(id)sender
     [defaultRunLoopTimer invalidate];
     
     if (!self.runLoopShouldContinue ) {
+        
+        if (self.connectionAgeTimer != nil ) {
```
jkennington: @cbommas, your earlier suggestion to simplify this was correct as this method will be called already on the correct thread.

AWSIoT/Internal/AWSIoTMQTTClient.m
```diff
@@ -547,6 +576,10 @@ - (void)reconnectToSession {
     
     self.connectionAgeInSeconds = 0;
     if (self.connectionAgeTimer != nil ) {
+        // NOTE: This does not work as intended. The reconnection attempt is always done on the reconnect thread
+        // but the timer needs to be invalidated on the streams thread to ensure memory is released.
+        // However, as currently implemented a switch to the streams thread here to clean this up is always beaten to the
+        // invalidation by other events which are are already being processed on the streams thread.
         [self.connectionAgeTimer invalidate];
```
jkennington: As noted this doesn't really work but other code avoids a memory leak. Ideally, this would have code to switch threads to clean up correctly but I left this alone for now as by the time a thread switch got control another event had already done the cleanup. I can update PR to have the thread switch if desired.

cbommas:
@jkennington 

Can we break this PR up into two separate ones? There are changes to the IOT code and the Core code. 

The challenge I have with AWSCore changes code is that we had to make the change to not invalidate the session due to crashes ( see #913). We will have to dive deeper on this. 

I will be testing the changes to the IOT code and prepping it for release pending the tests.

AWSIoT/Internal/AWSIoTMQTTClient.m
cbommas: For this change, rather than doing a thread switch, lets just remove this code, as it is handled in one of the other events.

AWSIoT/Internal/AWSIoTMQTTClient.m
cbommas: 👍

jkennington:
Have split into #1202 and #1203. Closing this PR.

PR #1196, opened 2019-01-24
Update version of ios sample applications and ios amplify doc

sunchunqiang:


sunchunqiang:
this is covered by pr 1198

PR #1203, opened 2019-01-29
Fix memory leaks from missing invalidations

jkennington:
Core framework side of fixes needed to avoid IoT memory leaks.

This will still need work to avoid problem due to existing bugs in other consumption of the Core framework.

From @cbommas 
> The challenge I have with AWSCore code changes is that we had to make the change to not invalidate the session due to crashes ( see #913). We will have to dive deeper on this.

These changes, in conjunction with #1202, fix leaking IoT objects.


palpatim:
Looking at this in relation to #913, I'm concerned that we may our retry logic may attempt to reissue a request after the session has been invalidated. I'm exploring this more to see if I can prove or disprove this.

palpatim:
@jkennington 

I have submitted PR #1728 based on this original PR. In it, I made a number of changes to your original approach that I think will make this a bit safer and mitigate the risk of accidentally reintroducing the crash reported in #913, which was what led to the removal in the first place.

I am closing this PR in favor of that one, but thank you for the initial work getting this underway.

Issue #1230, opened 2019-02-13
[S3TransferUtility] `uploadUsingMultiPart` is upload fails at credential update timing 


kimiyash:
**Describe the bug**
I'm using a multipart upload with cognito authentication.
But sometimes multipart uploading failed with refresh credential timing.

I have investigated a above behavior with created a uses S3 TransferUtility multipart upload simple project.

https://github.com/kimiyash/S3TransferUtilityMultipartUploadTest/blob/master/S3TransferUtilityMultipartUploadTest/AppDelegate.swift

NOTE, In order for this to work you need to include the appropriate settings in the Config structure. I have entered settings for cognito's identity pool/identity id/long time alives token and s3 that I can use.

I got a filtered console log by `[debug print]` string with above sample project
```
2019/02/13 13:23:45.890 [debug print] - upload() - start
2019/02/13 13:23:45.929 [debug print] - upload() - uploadDateString:Wednesday, February 13, 2019 at 1:23:45 PM Japan Standard Time
2019/02/13 13:23:46.182 [debug print] - token() - DeveloperCognitoCredentialProvider
2019/02/13 13:23:47.342 [debug print] - upload() - uploadUsingMultiPart - continueWith:<AWSTask: 0x600000e84400; completed = YES; cancelled = NO; faulted = NO; result = <AWSS3TransferUtilityMultiPartUploadTask: 0x600002a08c80>>
2019/02/13 13:23:47.342 [debug print] - upload() - uploadUsingMultiPart - continueWith - task.result.status:1
2019/02/13 13:23:47.830 [debug print] - upload() - completionHandler
...
2019/02/13 13:33:46.681 [debug print] - token() - DeveloperCognitoCredentialProvider
2019/02/13 13:33:46.781 [debug print] - upload() - uploadUsingMultiPart - continueWith:<AWSTask: 0x600000eb4d80; completed = YES; cancelled = NO; faulted = NO; result = <AWSS3TransferUtilityMultiPartUploadTask: 0x600002a1d2c0>>
2019/02/13 13:33:46.781 [debug print] - upload() - uploadUsingMultiPart - continueWith - task.result.status:1
```

I expected a upload processing continue, but the next upload processing did not start.
At this time the content of the log output at [here](https://github.com/aws-amplify/aws-sdk-ios/blob/master/AWSS3/AWSS3TransferUtility.m#L1335) was like this.

```
[...]

PR #1231, opened 2019-02-13
Naive fixing for issue #1230

kimiyash:
Naive fixing for issue https://github.com/aws-amplify/aws-sdk-ios/issues/1230


cbommas:
As discussed in #1230, I am going to go with an approach that does not use `waitUntilFinished`. I will go ahead and reject this PR.

PR #1273, opened 2019-03-09
Propagate content-disposition header, if set in upload request expression

colinhumber:


cbommas:
@colinhumber 
Thank you for the PR. I will target including this in the next rev of the SDK.

Issue #1246, opened 2019-02-23
AWSMobileClient change password API can not give exactly exception when it failed


HayTran94:
**Describe the bug**
When using AWSMobileClient API to change password, if user give incorrect current password, the API return error object is not instance of AWSMobileClientError to cast error. So I can not display error properly to user.

**To Reproduce**
```
 AWSMobileClient.sharedInstance().changePassword(currentPassword: currentPwd, proposedPassword: proposedPwd) { (error) in
              if error != nil {
                    log.info(error?.localizedDescription)
                    if let error = error as? AWSMobileClientError {
                      // Never be called
                    } else {
                       // Always be called
                    }
                } else {
                    // Success
                }
        }

```
**Which AWS service(s) are affected?**
AWSMobileClient

**Expected behavior**
Change password exception object should be instance of AWSMobileClient to cast error easier.

**Environment(please complete the following information):**
 - AWS iOS SDK Version: 2.9.0
 - Dependency Manager: CocoaPod
 - Swift Version : 4.0
 - Device: Iphone 6
 - iOS Version: 12.1.4

rohandubal:
Hello @HayTran94 

Sorry that you are having issues using the SDK.

I have root caused this issue to this line:
https://github.com/aws-amplify/aws-sdk-ios/blob/master/AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClientExtensions.swift#L504

We are not casting the error to appropriate `AWSMobileClientError`. I am working on a fix for this issue and it will be fixed in an upcoming release.

Thanks,
Rohan

muellerfr:
A fix for this issue was released with [version 2.9.4](https://github.com/aws-amplify/aws-sdk-ios/releases/tag/2.9.4).

PR #1301, opened 2019-03-17
Cast error to AWSMobileClientError when using `changePassword` API.

rohandubal:
*Issue #, if available:*
#1246 

*Description of changes:*
- Cast error to AWSMobileClientError for `changePassword` API.
- Update test to check for type.


rohandubal:
This is changing the behavior of the API as it should work. Won't call it a breaking change since it can still be caught as type `Error`(though underlying type will change.)

PR #1315, opened 2019-03-21
[IoT] Fixes username metadata tracking feature

minbi:
Adds the default values back into metadata and preserves previous metadata between adds
Adds unit test for the expected behavior

### DIFF
--- a/AWSIoT/AWSIoTDataManager.m
+++ b/AWSIoT/AWSIoTDataManager.m
@@ -209,12 +209,6 @@ static NSString *const AWSInfoIoTDataManager = @"IoTDataManager";
 
 @implementation AWSIoTDataManager
 
-/*
- This version is for metrics collection for AWS IoT purpose only. It may be different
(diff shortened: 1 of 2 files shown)


minbi:
The `SDK_VERSION` was removed because no one was updating the value when they made a change. The tracker is more reliable with the generic IoT SDK version

AWSIoT/AWSIoTDataManager.m
```diff
@@ -328,7 +322,7 @@ - (instancetype)initWithConfiguration:(AWSServiceConfiguration *)configuration w
         if(_mqttClient == nil){
             AWSDDLogError(@"**** mqttClient is nil. **** ");
         }
-        _mqttClient.userMetaData = [NSString stringWithFormat:@"%@%@", @"?SDK=iOS&Version=", SDK_VERSION];
```
desokroshan: Remove SDK_VERSION if it is not being used

minbi: Please check the latest update; it's already removed.

PR #1370, opened 2019-04-01
models-update-monitoring-2019-04-01-15-55

rohandubal:
models-update-monitoring-2019-04-01-15-55


rohandubal:
@sunchunqiang This seems like a genuine failure case. Can you have a look?

rohandubal:
There is a bug which needs to be fixed while generating code for this model. There are 2 different errors: `ResourceNotFound` and `ResourceNotFoundException` which are both translated to `ResourceNotFound` error enum. 

We do not have any similar cases currently and need to determine a path to address this issue.

PR #1374, opened 2019-04-01
models-update-sts-2019-04-01-15-55

rohandubal:
models-update-sts-2019-04-01-15-55


rohandubal:
@sunchunqiang This is not created in the right folder. This should be part of [AWSCore](https://github.com/aws-amplify/aws-sdk-ios/tree/develop/AWSCore/STS). We also don't publish podspec for `AWSSTS`, so need to disable that as well.

rohandubal:
Closing this PR as it was opened in error.

PR #1381, opened 2019-04-01
models-update-cognito-sync-2019-04-01-16-32

rohandubal:
models-update-cognito-sync-2019-04-01-16-32


sunchunqiang:
close the pr as model update for the package is turn off

PR #1433, opened 2019-04-08
by pass test in master by default

sunchunqiang:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -200,15 +200,17 @@ commands:
     description: >-
       skip the test job is project variabe OVERRIDE_TEST is set with TRUE
     steps:
+
       - run:
-          name: skip test in master
+          name: skip test if the commit is already tested
           command: |
-              notskipflag=$(echo "${NOT_SKIPTESTINMASTER}" |  tr '[:upper:]' '[:lower:]')
-              if [ "${notskipflag}" != "true" ]  &&  [ "${CIRCLE_BRANCH}" == "master" ]
-              then 
-                echo "bypass test job in master"
-                circleci step halt
-              fi          
+            echo "${TESTED_COMMITID}:${CIRCLE_SHA1}"
+            skipflag=$(echo "${SKIPTESTEDCHANGE}" |  tr '[:upper:]' '[:lower:]')
+            if [ "${TESTED_COMMITID}" == "${CIRCLE_SHA1}" ] && [ "${skipflag}" != "true" ] 
+            then
+              echo "skip test as the change ${CIRCLE_SHA1} is already tested"
+              circleci step halt              
+            fi      
       - run:
           name: check_for_test_override 
           command: |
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -214,6 +223,18 @@ commands:
     description: >-
       skip job unless required
     steps:
+      - run:
```
kvasukib: I do not see this step in Android CircleCI job. Is this not relevant in Android?

.circleci/config.yml
```diff
@@ -200,6 +200,15 @@ commands:
     description: >-
       skip the test job is project variabe OVERRIDE_TEST is set with TRUE
```
palpatim: s/variabe/variable/

.circleci/config.yml
```diff
@@ -214,6 +223,18 @@ commands:
     description: >-
       skip job unless required
     steps:
+      - run:
+          name: skip_job_from_enviroment_variable
```
palpatim: s/enviroment/environment/

.circleci/config.yml
```diff
@@ -214,6 +223,18 @@ commands:
...
+          command: |
+              varaible_name=SKIP_JOB_${CIRCLE_JOB}
```
palpatim: s/varaible_name/variable_name/

.circleci/config.yml
sunchunqiang: this feature for android was already checked in  before

.circleci/config.yml
```diff
@@ -214,6 +223,18 @@ commands:
...
+              
+              skipjob_flag=$(eval "echo $"${varaible_name}"")
+              skipjob_flag=$(echo "${skipjob_flag}" |  tr '[:upper:]' '[:lower:]')
+              if [ "${skipjob_flag}" == "true" ]
+              then
+                echo "skip current job as ${varaible_name} is set in project environment variabes"
```
palpatim: s/variabes/variables/

.circleci/config.yml
```diff
@@ -200,6 +200,15 @@ commands:
     description: >-
       skip the test job is project variabe OVERRIDE_TEST is set with TRUE
     steps:
+      - run:
+          name: skip test in master
+          command: |
+              notskipflag=$(echo "${NOT_SKIPTESTINMASTER}" |  tr '[:upper:]' '[:lower:]')
```
palpatim: 1. Where is `NOT_SKIPTESTINMASTER` set?
1. Why are we setting a `NOT` flag instead of defaulting to always running tests and forcing the preceding job (which I assume is a step that merges from the `develop` branch and recognizes that all tests have passed) to set the "skip" flag?

sunchunqiang: NOT_SKIPTESTINMASTER is set in circleci project environment variable
the reason we use NOT flag is to skip the test in master by default, that means when the flag is not set, we will skip the test as we already do the test in develop branch

.circleci/config.yml
sunchunqiang: done

.circleci/config.yml
sunchunqiang: done

.circleci/config.yml
sunchunqiang: done

.circleci/config.yml
```diff
@@ -200,11 +200,22 @@ commands:
     description: >-
       skip the test job is project variabe OVERRIDE_TEST is set with TRUE
     steps:
+
+      - run:
+          name: skip test if the commit is already tested
+          command: |
+            echo "${TESTED_COMMITID}:${CIRCLE_SHA1}"
+            skipflag=$(echo "${SKIPTESTEDCHANGE}" |  tr '[:upper:]' '[:lower:]')
```
palpatim: Where is `SKIPTESTEDCHANGE` be set?

.circleci/config.yml
palpatim: This is still misspelled--should be "environment"

PR #1460, opened 2019-04-14
models-update-polly-2019-04-14-15-00

aws-amplify-ops:
models-update-polly-2019-04-14-15-00


palpatim:
Investigate version bump script to see if it's overwriting the plist format.

Issue #1395, opened 2019-04-03
AWSIoT allow update to UserMetaData


xuaninbox:
addUserMetaData only allow append and not update. update should be allowed as the value may change.

PR #1486, opened 2019-04-18
Allow updates to user meta data field

desokroshan:
*Issue #, if available:*
https://github.com/aws-amplify/aws-sdk-ios/issues/1395

*Description of changes:*

### DIFF
--- a/AWSIoT/AWSIoTDataManager.h
+++ b/AWSIoT/AWSIoTDataManager.h
@@ -425,8 +425,8 @@ NS_ASSUME_NONNULL_BEGIN
  iotDataManager.addUserMetaData(userMetaData)
  @param userMetaData A dictionary of key-value metadata pairs to be appended to the connection username
  */
+- (void)addUserMetaData:(NSDictionary<NSString *, NSString *> *)userMetaData
 DEPRECATED_MSG_ATTRIBUTE("Use `updateUserMetaData` for updating the user meta data");
-- (void)addUserMetaData:(NSDictionary<NSString *, NSString *> *)userMetaData;
 
 /**
  Set user-specified dictionary of the additional values to be passed as components of
--- a/AWSIoT/AWSIoTDataManager.h
+++ b/AWSIoT/AWSIoTDataManager.h
@@ -425,7 +425,7 @@ NS_ASSUME_NONNULL_BEGIN
  iotDataManager.addUserMetaData(userMetaData)
  @param userMetaData A dictionary of key-value metadata pairs to be appended to the connection username
  */
-DEPRECATED_MSG_ATTRIBUTE("Use `updateUserMetaData` for updating the user meta data")
+DEPRECATED_MSG_ATTRIBUTE("Use `updateUserMetaData` for updating the user meta data");
 - (void)addUserMetaData:(NSDictionary<NSString *, NSString *> *)userMetaData;
 
 /**
--- a/AWSIoT/AWSIoTDataManager.h
+++ b/AWSIoT/AWSIoTDataManager.h
@@ -417,6 +417,18 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)enableMetricsCollection:(BOOL)enabled;
 
 /**
+ @deprecated This method is deprecated and will be deleted in the next minor version. Please use `updateUserMetaData` instead.
+ Set user-specified dictionary of the additional values to be passed as components of
+ connection username.
+ *Swift*
+ let userMetaData: [String: String] = ["AFRSDK": "ios", "AFRSDKVersion": "1.0.0", "AFRLibVersion":"1.4.1"]
+ iotDataManager.addUserMetaData(userMetaData)
+ @param userMetaData A dictionary of key-value metadata pairs to be appended to the connection username
+ */
+DEPRECATED_MSG_ATTRIBUTE("Use `updateUserMetaData` for updating the user meta data")
+- (void)addUserMetaData:(NSDictionary<NSString *, NSString *> *)userMetaData;
+
+/**
  Set user-specified dictionary of the additional values to be passed as components of
  connection username.
  *Swift*
--- a/AWSIoT/AWSIoTDataManager.h
+++ b/AWSIoT/AWSIoTDataManager.h
@@ -424,7 +424,7 @@ NS_ASSUME_NONNULL_BEGIN
  iotDataManager.addUserMetaData(userMetaData)
  @param userMetaData A dictionary of key-value metadata pairs to be appended to the connection username
  */
-- (void)addUserMetaData:(NSDictionary<NSString *, NSString *> *)userMetaData;
+- (void)updateUserMetaData:(NSDictionary<NSString *, NSString *> *)userMetaData;
 
 /**
  Initialises the MQTT session and connects to AWS IoT using certificate-based mutual authentication
--- a/AWSIoT/AWSIoTDataManager.m
+++ b/AWSIoT/AWSIoTDataManager.m
@@ -118,6 +118,7 @@ static NSString *const AWSInfoIoTDataManager = @"IoTDataManager";
 @property (nonatomic, strong) AWSIoTMQTTClient *mqttClient;
 @property  BOOL userDidIssueDisconnect;
 @property  BOOL userDidIssueConnect;
+@property (nonatomic, strong) NSMutableDictionary<NSString *, NSString *> *userMetaDataDict;
 @end
 
 @implementation AWSIoTMQTTLastWillAndTestament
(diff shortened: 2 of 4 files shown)


xuaninbox:
https://github.com/aws-amplify/aws-sdk-ios/blob/7cae401ed27a57830965e980705e81faea7a398a/AWSIoT/AWSIoTDataManager.m#L362

gives a warning in Xcode 'Format specifies type 'unsigned int' but the argument has type 'unsigned long'', should be %lu instead?

xuaninbox:
Looks good!

AWSIoT/AWSIoTDataManager.h
```diff
@@ -417,15 +417,27 @@ NS_ASSUME_NONNULL_BEGIN
...
 /**
+ @deprecated This method is deprecated and will be deleted in the next minor version. Please use `updateUserMetaData` instead.
  Set user-specified dictionary of the additional values to be passed as components of
  connection username.
  *Swift*
  let userMetaData: [String: String] = ["AFRSDK": "ios", "AFRSDKVersion": "1.0.0", "AFRLibVersion":"1.4.1"]
  iotDataManager.addUserMetaData(userMetaData)
  @param userMetaData A dictionary of key-value metadata pairs to be appended to the connection username
  */
+DEPRECATED_MSG_ATTRIBUTE("Use `updateUserMetaData` for updating the user meta data");
```
palpatim: I think this syntax is incorrect. `DEPRECATED_MSG_ATTRIBUTE` goes after the method declaration, before the terminating semicolon: `- (void)addUserMetaData:(NSDictionary<NSString *, NSString *> *)userMetaData DEPRECATED_MSG_ATTRIBUTE ("Use `updateUserMetaData` for updating the user meta data");`

Either way, please check this locally and make sure that Xcode properly reports the method as deprecated at the call site.

AWSIoT/AWSIoTDataManager.m
```diff
@@ -342,7 +344,47 @@ - (void)addUserMetaData:(NSDictionary<NSString *, NSString *> *)userMetaDataMap
...
+    for (id key in userMetaDataMap) {
+        [_userMetaDataDict setObject:[userMetaDataMap objectForKey:key] forKey: key];
+    }
+
+    // validate the length of username field
+    NSMutableString *userMetaDataString = [NSMutableString stringWithFormat:@"%@%@", @"?SDK=iOS&Version=", AWSIoTSDKVersion];
+
+    // Append each of the user-specified key-value pair to the connection username
+    if (_userMetaDataDict) {
+        for (id key in _userMetaDataDict) {
```
rohandubal: Can we combine the logic of constructing the `userMetaDataString`?

AWSIoT/AWSIoTDataManager.h
desokroshan: Verified that Xcode flags the deprecation warning upon call to ```addUserMetaData``` method

AWSIoT/AWSIoTDataManager.m
desokroshan: Discussed offline

AWSIoTTests/AWSIoTDataTests.m
```diff
@@ -995,17 +995,25 @@ - (void)testUsernameMetaData {
...
-    [iotDataManager addUserMetaData: metaData2];
-    NSString *expectedUserMetaData2 = [NSString stringWithFormat:@"?SDK=iOS&Version=%@&foo=bar&clazz=2&foo2=bar2", AWSIoTSDKVersion];
+    NSDictionary<NSString *,NSString *> * metaData2 = @{@"foo": @"bar2", @"foo3": @"bar3", @"foo4": @""};
+    [iotDataManager updateUserMetaData: metaData2];
+    NSString *expectedUserMetaData2 = [NSString stringWithFormat:@"?SDK=iOS&Version=%@&foo=bar2&foo4&clazz=2&foo3=bar3", AWSIoTSDKVersion];
     NSString *actualUserMetaData2 = [[iotDataManager mqttClient] userMetaData];
     XCTAssertTrue([actualUserMetaData2 isEqualToString:expectedUserMetaData2]);
+
+    // Check truncation logic
+    NSDictionary<NSString *,NSString *> * metaData3 = @{@"foo": @"bar2", @"foo3": @"bar3", @"foo4": @"", @"unusuallyLongKeynameExpectedToCauseTruncation":@"unusuallyLongValueExpectedToCauseTruncation",
```
palpatim: It would be better to limit the number of assertions in a single test case. Testing each member of a value might make sense in a single test case; testing validation logic doesn't really belong in the existence test.

More importantly, dictionary iteration is not guaranteed to have a deterministic order, so you can't rely on the metadata string always being the same unless there is logic that guarantees the order string. Which, if there is something that guarantees the order, needs its own test.

AWSIoTTests/AWSIoTDataTests.m
```diff
@@ -995,17 +995,31 @@ - (void)testUsernameMetaData {
...
     // Check state after adding additional fields
     NSDictionary<NSString *,NSString *> * metaData = @{@"foo": @"bar", @"clazz": @"2"};
-    [iotDataManager addUserMetaData: metaData];
-    NSString *expectedUserMetaData = [NSString stringWithFormat:@"?SDK=iOS&Version=%@&foo=bar&clazz=2", AWSIoTSDKVersion];
+    [iotDataManager updateUserMetaData: metaData];
+    NSString *defaultUserMetaData = [NSString stringWithFormat:@"?SDK=iOS&Version=%@", AWSIoTSDKVersion];
     NSString *actualUserMetaData = [[iotDataManager mqttClient] userMetaData];
-    XCTAssertTrue([actualUserMetaData isEqualToString:expectedUserMetaData]);
-    
+    XCTAssertTrue([actualUserMetaData hasPrefix:defaultUserMetaData] && [actualUserMetaData containsString:@"foo=bar"] && [actualUserMetaData containsString:@"clazz=2"]);
```
palpatim: Break each of these conditions into its own `XCTAssert` rather than combining them, so it's easier to tell which one fails.

AWSIoTTests/AWSIoTDataTests.m
```diff
@@ -995,17 +995,31 @@ - (void)testUsernameMetaData {
...
+
     // Check state after adding additional fields twice
-    NSDictionary<NSString *,NSString *> * metaData2 = @{@"foo2": @"bar2"};
-    [iotDataManager addUserMetaData: metaData2];
-    NSString *expectedUserMetaData2 = [NSString stringWithFormat:@"?SDK=iOS&Version=%@&foo=bar&clazz=2&foo2=bar2", AWSIoTSDKVersion];
+    NSDictionary<NSString *,NSString *> * metaData2 = @{@"foo": @"bar2", @"foo3": @"bar3", @"foo4": @""};
+    [iotDataManager updateUserMetaData: metaData2];
     NSString *actualUserMetaData2 = [[iotDataManager mqttClient] userMetaData];
-    XCTAssertTrue([actualUserMetaData2 isEqualToString:expectedUserMetaData2]);
+    XCTAssertTrue([actualUserMetaData2 containsString:@"foo=bar2"] && [actualUserMetaData2 containsString:@"foo3=bar3"] && [actualUserMetaData2 containsString:@"&foo4"] && [actualUserMetaData2 hasPrefix:defaultUserMetaData]);
```
palpatim: As above: Separate assertions

AWSIoTTests/AWSIoTDataTests.m
```diff
@@ -995,17 +995,31 @@ - (void)testUsernameMetaData {
...
+    [AWSIoTDataManager registerIoTDataManagerWithConfiguration:serviceConfig forKey:key];
+    __block AWSIoTDataManager *iotDataManager = [AWSIoTDataManager IoTDataManagerForKey:key];
+
+    // Check truncation logic
+    NSDictionary<NSString *,NSString *> * metaData = @{@"foo": @"bar2", @"foo3": @"bar3", @"foo4": @"", @"unusuallyLongKeynameExpectedToCauseTruncation":@"unusuallyLongValueExpectedToCauseTruncation",
+                                                        @"unusuallyLongKeynameExpectedToCauseTruncation2":@"unusuallyLongValueExpectedToCauseTruncation2", @"unusuallyLongKeynameExpectedToCauseTruncation3":@"unusuallyLongValueExpectedToCauseTruncation3"};
+    [iotDataManager updateUserMetaData: metaData];
+    NSString *defaultUserMetaData = [NSString stringWithFormat:@"?SDK=iOS&Version=%@", AWSIoTSDKVersion];
+    NSString *actualUserMetaData = [[iotDataManager mqttClient] userMetaData];
+    XCTAssertTrue([actualUserMetaData hasPrefix: defaultUserMetaData] && [actualUserMetaData length] == 255);
```
palpatim: As above: separate assertions. Good catch asserting that the prefix is still present, to ensure we didn't truncate from the wrong side. :)

AWSIoTTests/AWSIoTDataTests.m
desokroshan: Done

AWSIoTTests/AWSIoTDataTests.m
desokroshan: Addressed

PR #1540, opened 2019-05-10
Fix cross-module include files to properly use framework prefixes

hughescr:
This fixes breakage when each framework is in a separate project file, as when
using the new `generate_multiple_pod_projects` feature in cocoapods 1.7

*Issue #, if available:* NA

*Description of changes:*

This changes a few `#import "something.h"` to be the more-correct `#import <Module/something.h>` which means that if you use the new `generate_multiple_pod_projects` in cocoapods 1.7 you will still be able to build.  See http://blog.cocoapods.org/CocoaPods-1.7.0-beta/ for info


polmum:
Hey,
We are also interested in using generate_multiple_pod_projects, but are blocked by the code from this request being merged.
Any ETA for merging?

ReneLindhorst:
Works for me with `AWSCognitoIdentityProvider` using Cocoapods 1.7.2 and ":generate_multiple_pod_projects => true". 👍

#1611 additionally adds `AWSXMLDictionary` to the public headers of `AWSCore`. Probably a good idea as `AWSS3` is using it.

palpatim:
Thanks for your patience everybody. I've merged this to `develop` and will make sure it gets into our next release.

PR #1541, opened 2019-05-12
models-update-cognito-idp-2019-05-12-15-00

aws-amplify-ops:
models-update-cognito-idp-2019-05-12-15-00


palpatim:
Incorrect license. Will fix in codegen template.

PR #1555, opened 2019-05-17
TransferUtility createUploadSubTask not handling some error cases.

wfan9:
Made a change so that any errors from getPreSignedURL:getPreSignedURLRequest get returned instead of ignored.


palpatim:
@wfan9 Thanks for the PR. We fixed this issue as part of #1627 -- we actually went a bit further by propagating the exception as well. :)

palpatim:
@wfan9 Thanks for the PR. I made one change to remove the assignment to `error`, since we actually did that in #1627, which was created later but merged earlier than yours.

I've approved this for now, and will merge it once I merge a unit test verifying this behavior.

palpatim:
Closing this PR in favor of #1650, which includes your commit but adds unit test

PR #1556, opened 2019-05-17
Fix retain cycle between NSURLSession and AWSURLSessionManager

jaetzold:
*Description of changes:*

`NSURLSession` strongly references its `delegate`. The session releases the delegate only once it is invalidated or the application exits (See https://developer.apple.com/documentation/foundation/nsurlsession/1411530-delegate?language=objc)
`AWSURLSessionManager` strongly references the `NSURLSession` it is itself a delegate to, this way creating a retain cycle.

This change adds a `finishTasksAndInvalidate` method to `AWSURLSessionManager` that is forwarded to the `NSURLSession` so that the retain cycle will eventually be broken. It is called when the `AWSURLSessionManager` will no longer be used by the `AWSNetworking` instance that created it.

---

### DIFF
--- a/AWSCore/Networking/AWSNetworking.m
+++ b/AWSCore/Networking/AWSNetworking.m
@@ -83,6 +83,10 @@ NSString *const AWSNetworkingErrorDomain = @"com.amazonaws.AWSNetworkingErrorDom
     return self;
 }
 
+- (void)dealloc {
+    [self.networkManager finishTasksAndInvalidate];
+}
+
 - (AWSTask *)sendRequest:(AWSNetworkingRequest *)request {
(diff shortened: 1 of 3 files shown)


palpatim:
@jaetzold Thanks for the PR. This is actually a duplicate in spirit, if not in the exact implementation, to #1203 from @jkennington, and subject to the same concerns about accidentally reintroducing the crash that we fixed in #913 by _removing_ the invalidation. :)

I'm closing this as a duplicate, and will continue the evaluation & discussion on #1203.

palpatim:
I ended up with these changes (and @jkennington's from #1203) on PR #1728. Got a bit more work to do to add some tests, but I'm hoping to have this wrapped up shortly.

Issue #1398, opened 2019-04-03
AWSIoT mqtt may fail to decode mqtt message when doing Ota update


xuaninbox:
After mqtt connect and no action for a while, it may fail to publish message. It seems the web socket has timed out but we didn't get a callback of the status change.

cbommas:
@xuaninbox 
The SDK will send a periodic ping message to make sure that the connection is still valid. Can you share your MQTT Configuration settings, the Quality of service setting, code snippets and logs to help debug further?

xuaninbox:
Update - After further debugging, we saw that the network is NOT the issue, the mqtt message did reached the mqtt client however the decoding is not getting called.

Because we are doing OTA, the messages are coming in a burst (about 200 at a time), not sure if this is related to the decoding not getting triggered.

cbommas:
@xuaninbox 
Can you share logs, configuration, QOS settings etc., to help debug this further

xuaninbox:
After debugging with @roskumr we have narrowed it down to 

https://github.com/aws-amplify/aws-sdk-ios/blob/1229b7b89f631b2a578c3b22d464ba18642556b3/AWSIoT/Internal/MQTTSDK/AWSMQTTDecoder.m#L63-L95

where the byte it read from the stream about the header or length is wrong, which in turn caused all following read from the stream are wrong, as there's not clear of the stream, the retry will fail as well.

We are still investigating if its the raw payload from mqtt is wrong (less likely) or the read from web socket stream has some issue.

xuaninbox:
we have also identified that a check should be added to check if packets are corrupted at 

https://github.com/aws-amplify/aws-sdk-ios/blob/1229b7b89f631b2a578c3b22d464ba18642556b3/AWSIoT/Internal/MQTTSDK/AWSMQTTDecoder.m#L93

else will cause a overflow issue

see: https://github.com/novastone-media/MQTT-Client-Framework/issues/483

PR #1595, opened 2019-06-06
Add check for malformed remaining length

desokroshan:
*Issue #, if available:*
This is part of fix for https://github.com/aws-amplify/aws-sdk-ios/issues/1398

*Description of changes:*
Updating the algorithm to decode the remaining length of MQTT packet as per specification
http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/csprd02/mqtt-v3.1.1-csprd02.html#_Toc385349759

### DIFF
--- a/AWSIoT/Internal/MQTTSDK/AWSMQTTDecoder.m
+++ b/AWSIoT/Internal/MQTTSDK/AWSMQTTDecoder.m
@@ -29,6 +29,8 @@
 
 @implementation AWSMQTTDecoder
 
+int maxLengthMultiplier = 128 * 128 * 128;
+
 - (id)initWithStream:(NSInputStream*)aStream
 {
     _status = AWSMQTTDecoderStatusInitializing;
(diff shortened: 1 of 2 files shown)


AWSIoT/Internal/MQTTSDK/AWSMQTTDecoder.m
```diff
@@ -53,7 +53,7 @@ - (void)close {
 
 - (void)stream:(NSStream*)sender handleEvent:(NSStreamEvent)eventCode {
     AWSDDLogVerbose(@"%s [Line %d] EventCode:%lu, stream: %@, Thread: %@", __PRETTY_FUNCTION__, __LINE__, (unsigned long)eventCode, sender, [NSThread currentThread]);
-
+    int maxLengthMultiplier = 128 * 128 * 128;
```
rohandubal: Not the right place to do this. It will still be computed every time there is an event. We should have it at a higher level that it is only computed once ideally.

Issue #1592, opened 2019-06-04
How do I authenticate to Amazon Cognito using AWSMobileClient?


ghost:
Problem: I have two user pools in Cognito. Let's call them Pool A and B. I am able to authenticate a user to "Pool A" successfully.

I need to authenticate a user created by backend dynamically to "Pool B" from the client.

Docs I have read: 
- Amazon Amplify
- Amazon Cognito
- Amazon SDK for iOS
- AWSMobileClient

My setup: 
- AWSMobileClient v2.9.8
- iOS 12.0
- iPhone 7 Plus

I have tried the following solution:

- awsconfiguration.json has information regarding "Pool A"
- I log in a user to "Pool A" and perform business logic (success)
- I log out the user globally and invalidate the tokens. (success)
- I setup configuration for "Pool B" from code, I'm not sure how to specify multiple user pools in awsconfiguration.json. Not sure if it's doable.
- I login the user created by the backend present in "Pool B". (fails with the error message "Could not get the end user to sign in")

AWSMobileClient.sharedInstance().signIn(username: "user_present_in_pool_a", password: "password" ) { result, error in

        debugPrint(result, error)

        // Async sign out

        AWSMobileClient.sharedInstance().signOut(
            options: .init(
                signOutGlobally: true,
                invalidateTokens: true
            ),
            completionHandler: { error in

            debugPrint(error)

            // setup configuration for "Pool B"
            let serviceConfig = AWSServiceConfiguration(
                region: .USWest2,
                credentialsProvider: nil
            )

            let poolConfig = AWSCognitoIdentityUserPoolConfiguration(
                clientId: "pool_b_client_id", clientSecret: nil, poolId: "pool_b_id"
            )

            // init pool client
            AWSCognitoIdentityUserPool.register(
                with: serviceConfig,
                userPoolConfiguration: poolConfig,
                forKey: "some_key"
            )

            AWSMobileClient.sharedInstance().signIn(
                username: "user_present_in_pool_b",
                password: "password"
            ) { result, error in
                debugPrint(result, error)
            }
        })
    }

The aws-sdk-ios documentation is vague and I would really appreciate pointers to appropriate docs or perhaps a solution. Thank you :)

desokroshan:
@aqeelbhat This is currently not supported in AWSMobileClient. I have marked this as a feature request. I will take it to the team for prioritization. At this point I do not have a specific timeline that I can share with you. I will update this thread when I have more information.

[...]

PR #1609, opened 2019-06-14
Added method to set the user pool at runtime

aarons22:
*Issue #, if available:*
 Related to #1592 

*Description of changes:*
As far as I can tell, there is currently no way to dynamically configure a user pool at runtime. Instead of using `awsconfiguration.json`, I'd like to be able to set the user pool during app setup. Our app will be pulling configuration options from a server based on a few parameters, so runtime configuration is a requirement.

An example of use:
```
let serviceConfig = AWSServiceConfiguration(
    region: .USEast1,
    credentialsProvider: nil
)
let poolConfig = AWSCognitoIdentityUserPoolConfiguration(clientId: "client_id_from_server",
                                                         clientSecret: "client_secret_from_server",
                                                         poolId: "pool_id_from_server",
                                                         shouldProvideCognitoValidationData: true,
                                                         pinpointAppId: nil)

AWSCognitoIdentityUserPool.register(with: serviceConfig,
                                    userPoolConfiguration: poolConfig,
                                    forKey: "some_key")
AWSMobileClient.sharedInstance().setUserPool(forKey: "some_key")
```

### DIFF
--- a/AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClientExtensions.swift
+++ b/AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClientExtensions.swift
@@ -246,6 +246,17 @@ extension AWSMobileClient {
     internal var currentUser: AWSCognitoIdentityUser? {
         return self.userPoolClient?.currentUser()
     }
+
+
+    /// Set the current cognito user pool. This assumes you have registerd
+    /// a pool with `AWSCognitoIdentityUserPool`.
+    ///
+    /// - Parameter key: The key used to register the user pool
+    public func setUserPool(forKey key: String) {
+        let client = AWSCognitoIdentityUserPool(forKey: key)
+        self.userpoolOpsHelper.userpoolClient = client
+        client.delegate = self.userpoolOpsHelper
+    }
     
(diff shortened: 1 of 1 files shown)


dimpiax:
@rohandubal When will you accept the tiny but revolutionary PR?

palpatim:
Thanks for the PR, but we have a few concerns. Although the PR is titled "configure user pool at runtime", you're really exposing a feature that lets an app *change* the user pool at runtime, which could cause issues: e.g., what if an app has in-process requests (e.g. S3 background uploads), launches the app with some user pool A (or no UP at all), then changes the pool at runtime? This makes more sense as a configuration option, similar to what JS does https://aws-amplify.github.io/amplify-js/api/classes/amplify.html#configure. That would allow for a broader set of use cases and provide more predictable behavior by requiring configurations at initialization time. We'll take that on as a feature request across both iOS and Android platforms, and make sure the feature is supported with tests.

In the meantime, if this is blocking your use case, I think your only recourse is to fork the repo to add this feature. As the owner of that change, you'd be in control of how you use it and could avoid situations like swapping out the UserPools client while current processes are running.

palpatim:
Tracking the Feature Request in #1649

PR #1611, opened 2019-06-15
Fix compile issues when installing AWSS3 with Cocoapods 1.7+ with multiple project generation enabled

colinhumber:
*Description of changes:*
Installing the AWSS3 SDK with Cocoapods 1.7+ with `generate_multiple_pod_projects` enabled caused the frameworks to fail compilation.

`AWSXMLDictionary` was being imported directly within the `AWSS3` project when it exists in `AWSCore`. Fixed this by exposing `AWSXMLDictionary` as a public header and importing the file directly from the `AWSCore` framework.

### DIFF
--- a/AWSCore.podspec
+++ b/AWSCore.podspec
@@ -17,5 +17,5 @@ Pod::Spec.new do |s|
   s.requires_arc = true
 
   s.source_files = 'AWSCore/*.{h,m}', 'AWSCore/**/*.{h,m}'
-  s.private_header_files = 'AWSCore/XMLDictionary/**/*.h', 'AWSCore/XMLWriter/**/*.h', 'AWSCore/FMDB/AWSFMDatabase+Private.h', 'AWSCore/Fabric/*.h', 'AWSCore/Mantle/extobjc/*.h', 'AWSCore/CognitoIdentity/AWSCognitoIdentity+Fabric.h'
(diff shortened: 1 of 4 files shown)


palpatim:
@colinhumber 

Thanks for the PR. Can you point to a small sample app that demonstrates the problem that this is intended to fix? I'm not familiar with the multiple pod projects feature of CocoaPods and how it might  interact. `pod lib lint` succeeds both with and without your change, so I'm unclear what impact this would have for other SDK users.

colinhumber:
@palpatim You bet. I'll get one prepped right away.

colinhumber:
[AWS-CocoaPods-Sample.zip](https://github.com/aws-amplify/aws-sdk-ios/files/3330948/AWS-CocoaPods-Sample.zip)

Building the project as is will fail as it can't find `AWSXMLDictionary.h` in the `AWSS3` project.

The details about that new pod setting is outlined [here](http://blog.cocoapods.org/CocoaPods-1.7.0-beta/). Basically, it splits each pod into it's own project under an umbrella Pods project instead of all pods being in a single master Pods project. This shouldn't impact any other SDK users, as the change is just being more explicit in the importing of files from different frameworks, in this case, importing `AWSXMLDictionary` which is in `AWSCore` into `AWSS3` using the framework import syntax.

ReneLindhorst:
@palpatim #1540 is related and additionally fixes imports in `AWSCognitoIdentityProvider`. It would be awesome if you could have a look at it.

palpatim:
Thanks for the PR and the sample, @colinhumber. I made one change to remove the AWSXMLDictionary import that @hughescr had already put in.

colinhumber:
@palpatim Sounds good!

PR #1631, opened 2019-07-01
Test cleanup & port to new roles/resources

palpatim:
Fixed a bug where multiple values would be added to the 'host' header when using the V4 signer on a request that already included a 'host' header.

- Manually updated AWSMachineLearningService.m to propagate URLString parameter
  to sendRequest. This will be overwritten by the next automatic model update

Also: Test cleanup & migration to new policies & resources

- APIGateway: error reporting; use new functions, including regenerating AWSLambdaMicroservice client
- Cognito: cleanup
- KinesisVideoStreams: credentials management
- Lambda: use new functions
- Lex: Added AWSLexUnitTests to AWSLex scheme
- Logs: Added AWSLogsUnitTests to AWSLogs scheme
- MobileAnalytics: Removed 'Unknown' context attributes since it is a
  deprecated client and those attribute tests are broken in newer Xcode
  environments
- Rekognition: cleanup
- S3: cleanup, use new buckets
- SageMaker: use Cognito credentials instead of static AWS creds
- SES: cleanup
- SDB: cleanup
- SNS: cleanup
- SQS: cleanup
- Transcribe: cleanup

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/AWSAPIGatewayTests/AWSLambdaMicroserviceClient.m
+++ b/AWSAPIGatewayTests/AWSLambdaMicroserviceClient.m
@@ -14,11 +14,13 @@
  */
  
 
+
 #import "AWSLambdaMicroserviceClient.h"
 #import <AWSCore/AWSCore.h>
 #import <AWSCore/AWSSignature.h>
 #import <AWSCore/AWSSynchronizedMutableDictionary.h>
 
+#import "AWSStringValue.h"
 
 @interface AWSAPIGatewayClient()
 
@@ -42,6 +44,12 @@
 
 @end
 
+@interface AWSLambdaMicroserviceClient()
+
+@property (nonatomic, strong) AWSServiceConfiguration *configuration;
+
+@end
+
 @interface AWSServiceConfiguration()
 
 @property (nonatomic, strong) AWSEndpoint *endpoint;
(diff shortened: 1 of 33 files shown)


palpatim:
Still need to verify the full "AllTests" run locally, and need to validate the AuthSDK tests as well, but this should be a good starting point.

AWSAPIGatewayTests/AWSLambdaMicroserviceClient.m
```diff
@@ -72,16 +86,51 @@ + (instancetype)defaultClient {
...
 - (instancetype)initWithConfiguration:(AWSServiceConfiguration *)configuration {
     if (self = [super init]) {
         _configuration = [configuration copy];
-        NSString *filePath = [[NSBundle bundleForClass:[self class]] pathForResource:@"credentials"
-                                                                              ofType:@"json"];
-        NSDictionary *credentialsJson = [NSJSONSerialization JSONObjectWithData:[NSData dataWithContentsOfFile:filePath]
-                                                                        options:NSJSONReadingMutableContainers
-                                                                          error:nil];
 
-        NSString *URLString = credentialsJson[@"apiEndpoint"];
```
kvasukib: Should this be part of the credentials.json file?

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## 2.9.10
 
+### Bug Fixes
+
```
kvasukib: SDK name is missing.

AWSAuthSDK/Tests/AWSMobileClientTests/AWSMobileClientTests.swift
```diff
@@ -757,7 +790,57 @@ class AWSMobileClientTests: XCTestCase {
             downloadExpectation.fulfill()
         }
         wait(for: [downloadExpectation], timeout: 5)
-
+        
     }
+    
+}
 
+extension AWSMobileClientError {
```
kvasukib: Will this be useful if it's part of the SDK?

palpatim: I thought about it, but I'd prefer not to clutter the API at this point.

CHANGELOG.md
palpatim: Thanks, fixed

AWSAPIGatewayTests/AWSLambdaMicroserviceClient.m
palpatim: This was downloaded verbatim from the API Gateway site. I'd prefer to leave it as-is so we get a better test of what customers are actually dealing with. It's not sensitive information, inasmuch as it simply exposes an echo API that we can protect with appropriate auth methods as needed.

PR #1642, opened 2019-07-09
[AWSMobileClient] Add test for GetIdentityID

kvasukib:
*Issue #, if available:*

*Description of changes:*


AWSAuthSDK/Tests/AWSMobileClientTests/AWSMobileClientTests.swift
```diff
@@ -793,6 +795,37 @@ class AWSMobileClientTests: XCTestCase {
...
+            return nil
+        })
+        wait(for: [identityIdExpectation], timeout: 5)
+
+        guard AWSMobileClient.sharedInstance().identityId != nil else {
+            XCTFail("Could not find identityId.")
+            return
+        }
+
+        let verifyCredentialsExpectation = expectation(description: "Credentials should be retrieved successfully")
```
palpatim: I'm losing track of what you're actually testing in this. Are you asserting in the second half that `getAWSCredentials` doesn't incorrectly _clear_ the identity ID? I think this would be clearer with a separate test. In general let's try to minimize the number of assertions in each of our tests, with 1 being the ideal number.

AWSAuthSDK/Tests/AWSMobileClientTests/AWSMobileClientTests.swift
```diff
@@ -793,6 +795,37 @@ class AWSMobileClientTests: XCTestCase {
         
     }
     
+    func testGetIdentityId() {
+        XCTAssertNil(AWSMobileClient.sharedInstance().identityId, "Identity Id should be nil after initialize.")
+
+        let identityIdExpectation = expectation(description: "IdentityID should be retrieved successfully")
```
palpatim: Nit: you're fulfilling the expectation upon any invocation of the block, not only successful retrieval. Please make the name & description conform to what you're actually using it for.

AWSAuthSDK/Tests/AWSMobileClientTests/AWSMobileClientTests.swift
```diff
@@ -793,6 +795,37 @@ class AWSMobileClientTests: XCTestCase {
...
+
+        let verifyCredentialsExpectation = expectation(description: "Credentials should be retrieved successfully")
+        AWSMobileClient.sharedInstance().getAWSCredentials { (creds, error) in
+            XCTAssertNil(error)
+            XCTAssertNotNil(creds)
+            verifyCredentialsExpectation.fulfill()
+        }
+        wait(for: [verifyCredentialsExpectation], timeout: 5)
+
+        guard AWSMobileClient.sharedInstance().identityId != nil else {
```
palpatim: Generally we want to reserve `guard` for early exits. In this case, you can just `XCTAssertNotNil(AWSMobileClient.sharedInstance().identityId)`

AWSAuthSDK/Tests/AWSMobileClientTests/AWSMobileClientTests.swift
kvasukib: Removed `getAWSCredentials` from the test.

AWSAuthSDK/Tests/AWSMobileClientTests/AWSMobileClientTests.swift
kvasukib: Renamed the expectation.

AWSAuthSDK/Tests/AWSMobileClientTests/AWSMobileClientTests.swift
kvasukib: Changed guard to assert.

Issue #1554, opened 2019-05-17
AWSCredentials of DeveloperAuthenticatedIdentityProvider are not being cached


phanchutoan:
**Describe the bug**
AWSCredentials of DeveloperAuthenticatedIdentityProvider seems not being cached, this make creating multiple pre-signed urls run slow on app second start

**To Reproduce**

Using Developer Authenticated Identity Provider
```
DeveloperAuthenticatedIdentityProvider * devAuth = [[DeveloperAuthenticatedIdentityProvider alloc] initWithRegionType:AWSRegionAPNortheast1
                                                                                                           identityPoolId:@"xxxxxx"
                                                                                                          useEnhancedFlow:YES
                                                                                                  identityProviderManager:nil];
    
AWSCognitoCredentialsProvider *credentialsProvider = [[AWSCognitoCredentialsProvider alloc]
                                                      initWithRegionType:AWSRegionAPNortheast1
                                                      identityProvider:devAuth];


AWSServiceConfiguration *configuration = [[AWSServiceConfiguration alloc] initWithRegion:AWSRegionAPNortheast1 credentialsProvider:credentialsProvider];

AWSServiceManager.defaultServiceManager.defaultServiceConfiguration = configuration;
```

Get pre-signed urls in a loop

```
NSLog(@"i=0: %@", [NSDate date]);
NSString *bucket = @"example.com";
for (int i=0; i < 1000; i++) {
    NSString *key = [NSString stringWithFormat:@"testkey%d", i];
    AWSS3GetPreSignedURLRequest *getPreSignedURLRequest = [AWSS3GetPreSignedURLRequest new];
    getPreSignedURLRequest.bucket = bucket;
    getPreSignedURLRequest.key = key;
    getPreSignedURLRequest.expires = [NSDate dateWithTimeIntervalSinceNow:3600];
    getPreSignedURLRequest.HTTPMethod = AWSHTTPMethodGET;
    [[[AWSS3PreSignedURLBuilder defaultS3PreSignedURLBuilder] getPreSignedURL:getPreSignedURLRequest]
     continueWithBlock:^id(AWSTask *task) {
         if (task.error) {
             NSLog(@"Error: %@",task.error);
         } else {
             NSURL *presignedURL = task.result;
             NSLog(@"presignedURL: \n%@", presignedURL);
         }
         if (i == 999) {
             NSLog(@"i=999: %@", [NSDate date]);
         }
         return nil;
     }];
}
```
[...]

PR #1678, opened 2019-07-23
Initialize in-memory 'internalCredentials' from keychain

palpatim:
Set in-memory 'internalCredentials' to keychain value when retrieving from keychain (#1554)

- Removed empty test file
- Removed inadvertent commit of local file reference

*Issue #, if available:*
- #1554
- #1691

*Description of changes:*

Unfortunately there's not a good way of asserting this behavior, without mocking the keychain and tying closely to the internals of the two credentials providers this change touches. However, running the tests `-[AWSCognitoCredentialsProviderTests testEnhancedProviderKeychain]` and `-[AWSCognitoCredentialsProviderTests testWICProviderKeychain]` with verbose logging would output log messages to the console. You would expect to see the message "Retrieving credentials from keychain" appear four times:

- During static setUp: `+[AWSTestUtility setupCognitoCredentialsProvider]`
- During the creation of provider 1
- During the creation of provider 2 (inside of provider 1's callback)
- During teardown `-[AWSCognitoCredentialsProviderTests tearDown]` clearing the keychain


AWSCore/Authentication/AWSCredentialsProvider.m
```diff
@@ -237,6 +238,8 @@ - (AWSCredentials *)internalCredentials {
                                                                      sessionKey:self.keychain[AWSCredentialsProviderKeychainSessionToken]
                                                                      expiration:expiration];
 
+        _internalCredentials = credentials;
```
royjit: I see that in this class we assign the value to internalCredentials through async calls. [Example](https://github.com/aws-amplify/aws-sdk-ios/blob/master/AWSCore/Authentication/AWSCredentialsProvider.m#L425). 

So if we try to set _internalCredentials value in its getter, while a 'set' is happening in different thread can corrupt the value.

palpatim: Ah, this is a good point. Let me revisit the call flow to see the best way to handle this--that may be why the variable wasn't being set in the first place.

palpatim:
We released this in https://github.com/aws-amplify/aws-sdk-ios/releases/2.11.0

PR #1679, opened 2019-07-23
Fix for signInCallback not called for confirmSignIn

royjit:
*Description of changes:* Since signIn and confirmSignIn API of AWSMobileClient is using the same callback, we were facing a situation where the callback is changed to nil when we stack the calls. This PR try to set the callback ivar to nil before making the callback.


palpatim:
Discussed, withdrawing this PR to work on a more comprehensive fix.

PR #1701, opened 2019-07-29
Add generic test helper function to create and delete bucket for integration test

royjit:
S3 transferutility required to have a predefined bucket in S3 for the integration test to work. This change creates the bucket before test and delete it after.

### DIFF
--- a/AWSS3Tests/AWSS3TestHelper.m
+++ b/AWSS3Tests/AWSS3TestHelper.m
@@ -1,0 +1,141 @@
+//
+// Copyright 2010-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License").
+// You may not use this file except in compliance with the License.
+// A copy of the License is located at
+//
+// http://aws.amazon.com/apache2.0
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+//
+
+#import "AWSS3TestHelper.h"
+#define AWS_TEST_BAH_INSTEAD true
+
+@interface AWSS3CreateBucketConfiguration()
+
++ (NSValueTransformer *)locationConstraintJSONTransformer;
+
+@end
+
+@interface AWSEndpoint()
+
++ (NSString *)regionNameFromType:(AWSRegionType)regionType;
+
+@end
+
+@implementation AWSS3TestHelper
+
++ (BOOL)createBucketWithName:(NSString *)bucketName
+                   andRegion:(AWSRegionType)regionType {
+    
+    AWSS3 *s3 = [AWSS3 defaultS3];
+    
+    AWSS3CreateBucketRequest *createBucketReq = [AWSS3CreateBucketRequest new];
+    createBucketReq.bucket = bucketName;
+    
+    if (regionType != AWSRegionUSEast1) {
+        NSString *regionName = [AWSEndpoint regionNameFromType:regionType];
+        NSValueTransformer *locationConstraintTransformer = [AWSS3CreateBucketConfiguration locationConstraintJSONTransformer];
+        NSNumber *constraint = [locationConstraintTransformer transformedValue:regionName];
+        AWSS3BucketLocationConstraint dupcon = constraint.integerValue;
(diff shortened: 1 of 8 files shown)


AWSS3Tests/AWSS3TestHelper.m
```diff
@@ -0,0 +1,141 @@
...
+// http://aws.amazon.com/apache2.0
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+//
+
+#import "AWSS3TestHelper.h"
+#define AWS_TEST_BAH_INSTEAD true
```
palpatim: Is this still necessary now that this region has launched? (Not just the assignment, but the entire `AWS_TEST_BAH_INSTEAD` flag.)

AWSS3Tests/AWSS3TestHelper.m
```diff
@@ -0,0 +1,141 @@
...
+    
+    [[NSRunLoop currentRunLoop] runUntilDate:[NSDate dateWithTimeIntervalSinceNow:10]];
+    
+    return success;
+}
+
++ (BOOL)deleteBucketWithName:(NSString *)bucketName {
+    AWSS3 *s3 = [AWSS3 defaultS3];
+    
+    AWSS3ListObjectVersionsRequest *listObjectVersionsRequest = [AWSS3ListObjectVersionsRequest new];
```
palpatim: Why can't we just invoke `deleteAllObjects` before deleting the bucket? What is special about the individual deletions in this method?

AWSS3Tests/AWSS3TestHelper.m
```diff
@@ -0,0 +1,141 @@
...
+    [[[s3 listObjects:listObjectsRequest] continueWithBlock:^id(AWSTask *task) {
+        AWSS3ListObjectsOutput *output = task.result;
+        
+        for (AWSS3Object *object in output.contents) {
+            // Delete the object
+            AWSS3DeleteObjectRequest *deleteObjectRequest = [AWSS3DeleteObjectRequest new];
+            deleteObjectRequest.bucket = bucketName;
+            deleteObjectRequest.key = object.key;
+            
+            [[s3 deleteObject:deleteObjectRequest] continueWithBlock:^id(AWSTask *task) {
```
palpatim: Consider a batch delete here instead of individual single deletes: https://docs.aws.amazon.com/AmazonS3/latest/API/multiobjectdeleteapi.html

AWSS3Tests/AWSS3TransferUtilityTests.swift
```diff
@@ -86,12 +88,22 @@ class AWSS3TransferUtilityTests: XCTestCase {
             dataString += dataString
         }
         testData = dataString.data(using: String.Encoding.utf8)!
+        
+        let timeInterval = Date.timeIntervalSinceReferenceDate
+        generalTestBucket = "ios-v2-s3.periods\(timeInterval)"
```
palpatim: It's probably safe to make the bucket name more generic than "ios-v2-s3.periods", unless each created bucket is being created with periods, or unless the existence of periods is being asserted somewhere (which probably belongs in its own test then). Consider something like "ios-integ-tests-\(timeinterval)". Also, consider converting the timeInterval to milliseconds and truncating to an integer, to ensure the format is consistent and doesn't include trailing decimals (TimeInterval is a Double)

royjit: Renamed to s3-integ-test-, s3-integ-transfermanager-test-, s3-integ-transferutil-test- so that we can distinguish between tests.

AWSS3Tests/AWSS3TestHelper.m
royjit: Changed to multi delete.

AWSS3Tests/AWSS3TestHelper.m
royjit: Just deleting the objects will not delete the versions. And without deleting the version delete bucket will fail. For versioned buckets, when you attempt to delete objects, Amazon S3 inserts delete markers for all objects, but doesn't delete the object versions.
           
So we need to individually delete the version if present.

AWSS3Tests/AWSS3TestHelper.m
royjit: Removed

PR #1718, opened 2019-08-06
Fix for S3 integration test

royjit:
S3 test were failing for just created files when we try to download it using mobileClient

### DIFF
--- a/AWSS3Tests/AWSS3TransferUtilityTests.swift
+++ b/AWSS3Tests/AWSS3TransferUtilityTests.swift
@@ -2367,7 +2367,7 @@ class AWSS3TransferUtilityTests: XCTestCase {
                 transferUtility?.uploadData(
                     testData.data(using: String.Encoding.utf8)!,
                     bucket: generalTestBucket,
-                    key: key,q
+                    key: key,
                     contentType: "text/plain",
                     expression: uploadExpression,
                     completionHandler: uploadCompletionHandler
--- a/AWSS3Tests/AWSS3TransferUtilityTests.swift
+++ b/AWSS3Tests/AWSS3TransferUtilityTests.swift
@@ -2324,11 +2324,11 @@ class AWSS3TransferUtilityTests: XCTestCase {
                 if let HTTPResponse = task.response {
                     XCTAssertEqual(HTTPResponse.statusCode, 200)
 
-                    let timeout = DispatchTime.now() + .seconds(2)
+                    let timeout = DispatchTime.now() + .seconds(30)
                     while(!AWSS3TestHelper.checkIfObjectIsPresent(key, bucket: generalTestBucket)) {
                         if (DispatchTime.now() > timeout )  {
                             XCTFail("Could not verify uploaded key \(key)")
-                            break;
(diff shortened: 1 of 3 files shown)


AWSS3Tests/AWSS3TransferUtilityTests.swift
```diff
@@ -2322,7 +2322,8 @@ class AWSS3TransferUtilityTests: XCTestCase {
                 
                 if let HTTPResponse = task.response {
                     XCTAssertEqual(HTTPResponse.statusCode, 200)
-                    
+                    // Sleep for 2 sec so that the url created will be accessible.
+                    sleep(2)
```
palpatim: Instead, make this a loop that does a `headObject` until the metadata comes back/doesn't throw an exception, then do the download.

drochetti: Can we leverage [expectations](https://developer.apple.com/documentation/xctest/asynchronous_tests_and_expectations/testing_asynchronous_operations_with_expectations) in this case? A `waitUntil` utility could fit this case well

royjit: Thanks for the suggestion and I updated the test with a loop.

@drochetti Not sure whether we can use expectation for a pooling call. Maybe we might need to create our own implementation of expectation?

AWSS3Tests/AWSS3TransferUtilityTests.swift
```diff
@@ -2322,7 +2323,16 @@ class AWSS3TransferUtilityTests: XCTestCase {
                 
                 if let HTTPResponse = task.response {
                     XCTAssertEqual(HTTPResponse.statusCode, 200)
-                    
+
+                    let timeout = DispatchTime.now() + .seconds(2)
```
palpatim: We've been using 30s as a minimum network timeout. Since the loop will terminate as soon as the object is found, you won't be adding additional time to success cases by increasing the timeout, and you build in safety for latency & network flakiness.

AWSS3Tests/AWSS3TransferUtilityTests.swift
```diff
@@ -2322,7 +2323,16 @@ class AWSS3TransferUtilityTests: XCTestCase {
...
+                    while(!AWSS3TestHelper.checkIfObjectIsPresent(key, bucket: generalTestBucket)) {
+                        if (DispatchTime.now() > timeout )  {
+                            XCTFail("Could not verify uploaded key \(key)")
+                            break;
```
palpatim: Consider a `return` instead of a `break`, since the rest of the test conditions will fail if the object isn't present.

AWSS3Tests/AWSS3TransferUtilityTests.swift
royjit: Updated to 30

AWSS3Tests/AWSS3TransferUtilityTests.swift
royjit: Fixed

PR #1722, opened 2019-08-08
Release VC after CognitoAuth getSession

palpatim:
*Description of changes:*

Customer reports retain cycle in the view controller they're using as their CognitoAuth delegate. This change sets the delegate to `weak` ownership.

We don't have a good way of integration testing this, because the interaction depends on a view controller to present the Safari login page. However, I have verified the behavior manually by inspecting the object retain graph and noting that CognitoAuth no longer maintains a reference to the delegate after the VC is popped off the stack.

### DIFF
--- a/AWSiOSSDKv2.xcodeproj/project.pbxproj
+++ b/AWSiOSSDKv2.xcodeproj/project.pbxproj
@@ -93,7 +93,6 @@
 /* End PBXAggregateTarget section */
 
 /* Begin PBXBuildFile section */
-		170C2D471EDFAE43008A2E52 /* AWSCognitoAuthTests.m in Sources */ = {isa = PBXBuildFile; fileRef = EFDE85AC1ED203D9008841EC /* AWSCognitoAuthTests.m */; };
 		173641DE1ECBBABC00512239 /* AWSLambdaRequestRetryHandler.h in Headers */ = {isa = PBXBuildFile; fileRef = 173641DC1ECBBABC00512239 /* AWSLambdaRequestRetryHandler.h */; };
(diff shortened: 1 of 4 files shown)


royjit:
Should we add weak [here](https://github.com/aws-amplify/aws-sdk-ios/blob/master/AWSCognitoAuth/AWSCognitoAuth.m#L32) as well? 

The following function set that value.
```
cognitoAuth.getSession(self)  { (session:AWSCognitoAuthUserSession?, error:Error?) 
```

AWSiOSSDKv2.xcodeproj/project.pbxproj
```diff
@@ -12690,7 +12690,7 @@
 			isa = PBXSourcesBuildPhase;
 			buildActionMask = 2147483647;
 			files = (
-				170C2D471EDFAE43008A2E52 /* AWSCognitoAuthTests.m in Sources */,
```
royjit: Why was `AWSCognitoAuthTests` removed?

palpatim: That was a reference that included the file as part of the unit tests project. The file is still present, and linked. Of course, it's also empty, so it doesn't make a practical difference. :)

palpatim:
> Should we add weak [here](https://github.com/aws-amplify/aws-sdk-ios/blob/master/AWSCognitoAuth/AWSCognitoAuth.m#L32) as well?
> 
> The following function set that value.
> 
> ```
> cognitoAuth.getSession(self)  { (session:AWSCognitoAuthUserSession?, error:Error?) 
> ```

I looked at that, but we actually set it to `nil` in [`cleanupSignIn`](https://github.com/aws-amplify/aws-sdk-ios/blob/master/AWSCognitoAuth/AWSCognitoAuth.m#L235). I *think* it would be OK to weaken, but in the interests of minimizing changes I'd prefer to leave it at just the delegate.

Issue #1682, opened 2019-07-24
AWS Cogito Crashes


bfallon:
**Describe the bug**
Reviewing Crashlytics reports, the SDK has a lot of crashes.

**To Reproduce**
There are no steps to reproduce this issue.

**Which AWS service(s) are affected?**
Cognito

**Expected behavior**
We expect the SDK to not crash. 

**Code**
Below please find the classes and line numbers where the app crashes:

AWSTaskCompletionSource.m 
line 56 -[AWSTaskCompletionSource setError:]

AWSTaskCompletionSource.m 
line 49 -[AWSTaskCompletionSource setResult:]

AWSMobileClientExtensions.swift line 121
specialized closure #2 in AWSMobileClient.logins()

**Environment(please complete the following information):**
 - SDK Version: 2.9.6
 - Dependency Manager: CocoaPods
 - Swift Version: 4

**Device Information (please complete the following information):**
 - Device: Multiple iPhone devices
 - iOS Version: iOS 11 and iOS 12
 - Specific to simulators: This is specific to iOS devices.

palpatim:
I believe this is a dupe of #1505

palpatim:
This was released on [2.10.3](https://github.com/aws-amplify/aws-sdk-ios/releases/tag/2.10.3). Please let us know if you continue to experience problems.

Issue #1505, opened 2019-04-29
"Cannot set error on a completed task" exception on signOut()


jamesatfish:
**Describe the bug**
We have an iOS app using Cognito User Pools authentication.

Upon app launch in our App Delegate we initialise the AWS SDK (`AWSMobileClient.sharedInstance().initialize`) and if we get a response of `userState == .signedIn` then we proceed to retrieve the user's details from our app service database.

If the AWS SDK returns a different userState, or an error, we clean up by calling `signOut()` and `invalidateCachedTemporaryCredentials` on the SDK to ensure there's a clean slate (no cached tokens etc) for the user to login manually.

We are seeing Crashlytics / Instabug crash reports from some users where our app is crashing within seconds of launch.  The crash is arising from the AWS SDK:

```
0 | CoreFoundation | _exceptionPreprocess + 228
1 | libobjc.A.dylib | objc_exception_throw + 56
2 | CoreFoundation | init] + 0
3 | AWSCore | -[AWSTaskCompletionSource setError:] (AWSTaskCompletionSource.m:56)
4 | AWSMobileClient | specialized AWSMobileClient.performUserPoolSignOut() (<compiler-generated>:0)
5 | AWSMobileClient | AWSMobileClient.signOut() (AWSMobileClientExtensions.swift:598)

```

We have yet to be able to reproduce the issue in our own testing, but it's becoming a fairly frequent report from our users so there's obviously something not working as expected in the AWS SDK.

**To Reproduce**
Per above, as best we can gather:

1. In AppDelegate `applicationDidBecomeActive`, call `AWSMobileClient.sharedInstance().initialize`

2. When the `userState != .signedIn`, call signOut() and the various cleanup commands to clear any cached tokens / credentials / user details:

```
AWSMobileClient.sharedInstance().signOut()
AWSMobileClient.sharedInstance().invalidateCachedTemporaryCredentials()
AWSMobileClient.sharedInstance().clearCredentials()
AWSMobileClient.sharedInstance().clearKeychain()
```

3. On some occasions (we are unsure of what else is required to trigger these occasions) the app will crash.

**Which AWS service(s) are affected?**
Cognito login with User Pool accounts

**Expected behavior**
We simply expected the app not to crash.  We expected any errors in the signOut() function to be handled safely by the AWS SDK.

**Screenshots**
N/A.

**Environment(please complete the following information):**
 - SDK Version:
From Cocoapods:
```
[...]

PR #1726, opened 2019-08-08
Check task before adding error in signout

royjit:
*Issue #1505 
*Issue #1682 

*Description of changes:*
Multiple occurrence of the below crash:
```
# Cause: Cannot set the error on a completed task.
```
This change check the task status before setting a value.

Steps to reproduce:

1. `getSession()` invokes `interactiveAuth` [here](https://github.com/aws-amplify/aws-sdk-ios/blob/abf63f5bf5ad7dc85a6006a9b59db130aab59153/AWSCognitoIdentityProvider/AWSCognitoIdentityUser.m#L219) 
2. `interactiveAuth` internally calls [`passwordAuthInternal`](https://github.com/aws-amplify/aws-sdk-ios/blob/abf63f5bf5ad7dc85a6006a9b59db130aab59153/AWSCognitoIdentityProvider/AWSCognitoIdentityUser.m#L538) which then creates and set [AWSCompletionSource](https://github.com/aws-amplify/aws-sdk-ios/blob/abf63f5bf5ad7dc85a6006a9b59db130aab59153/AWSCognitoIdentityProvider/AWSCognitoIdentityUser.m#L553)
3. `passwordAuthTaskCompletionSource` is always set to a value [here](https://github.com/aws-amplify/aws-sdk-ios/blob/dd963da9cf146c7eda4987edbb81d1889e83a755/AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClientExtensions.swift#L951) during signedIn or invalid token. But it is already made a completed task [here](https://github.com/aws-amplify/aws-sdk-ios/blob/dd963da9cf146c7eda4987edbb81d1889e83a755/AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClientExtensions.swift#L947)
4. So now if we call `signOut()` result in crash.

Screenshot with the flow that set passwordAuthTaskCompletionSource
Screen Shot 2019-08-12 at 3 28 02 PM


palpatim:
Please add CHANGELOG entry as well.

PR #1728, opened 2019-08-09
Fix memory leaks from missing invalidations

palpatim:
*Issue #, if available:*

Extended #1203, and #1556 with a flag that indicates a session is in the process of being invalidated, and added an error type to return when attempting to add a task to an invalidated session.

*Description of changes:*

### DIFF
--- a/AWSCore/Networking/AWSNetworking.m
+++ b/AWSCore/Networking/AWSNetworking.m
@@ -86,6 +86,13 @@ NSString *const AWSNetworkingErrorDomain = @"com.amazonaws.AWSNetworkingErrorDom
 - (AWSTask *)sendRequest:(AWSNetworkingRequest *)request {
     return [self.networkManager dataTaskWithRequest:request];
 }
+
+- (void)dealloc {
+    // If this is being released, the network manager should be notified so it can invalidate
+    // its NSURLSession to avoid a memory leak.
+    [_networkManager invalidate];
+}
+
 @end
 
 #pragma mark - AWSNetworkingConfiguration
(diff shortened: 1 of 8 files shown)


AWSCore/Networking/AWSURLSessionManager.m
```diff
@@ -186,6 +195,13 @@ - (void)taskWithDelegate:(AWSURLSessionManagerDelegate *)delegate {
 
             [self printHTTPHeadersAndBodyForRequest:delegate.request.task.originalRequest];
 
+            if (!self.session || !self.isSessionValid) {
```
royjit: Should we move this above line 193 ie before adding the delegate to `self.sessionManagerDelegates`

palpatim: Yes, good call, will do on next commit.

palpatim:
Addressed @royjit's comment, and added tests.

palpatim:
Integration tests are passing locally with this change.

AWSCore/Networking/AWSNetworking.m
```diff
@@ -58,11 +58,19 @@ + (instancetype)aws_stringWithHTTPMethod:(AWSHTTPMethod)HTTPMethod {
...
+
+- (void)invalidate;
+
+@end
+
 #pragma mark - AWSNetworking
 
 @interface AWSNetworking()
 
-@property (nonatomic, strong) AWSURLSessionManager *networkManager;
```
royjit: Update here as well ? [Code Ref](https://github.com/aws-amplify/aws-sdk-ios/blob/6ab2982040eaca984eab38e311b98c3b9227b773/AWSCoreTests/AWSNetworkingTests.m#L132)

palpatim: Great catch, thanks!

Issue #1717, opened 2019-08-05
Calling AWSMobileClient.signOut with completion handler never calls the completion handler if default SignOut options are used.


jasonbobier:
**Describe the bug**
Calling AWSMobileClient.signOut with completion handler never calls the completion handler if default SignOut options are used.

**To Reproduce**
Add this call to your code:

AWSMobileClient.sharedInstance().signOut { (error) in
    doSomething()
}

doSomething will never get called

**Expected behavior**
The completion handler will be called when signOut completes.

**Smartphone (please complete the following information):**
 - Device: iPhone
 - OS: iOS 12.4

PODS:
  - Amplitude-iOS (4.6.0)
  - AWSAppSync (2.14.0):
    - AWSCore (~> 2.9.0)
    - ReachabilitySwift (= 4.3.0)
    - SQLite.swift (= 0.11.6)
  - AWSAuthCore (2.9.10):
    - AWSCore (= 2.9.10)
  - AWSCognitoIdentityProvider (2.9.10):
    - AWSCognitoIdentityProviderASF (= 1.0.1)
    - AWSCore (= 2.9.10)
  - AWSCognitoIdentityProviderASF (1.0.1)
  - AWSCore (2.9.10)
  - AWSMobileClient (2.9.10):
    - AWSAuthCore (= 2.9.10)
    - AWSCognitoIdentityProvider (= 2.9.10)
  - ReachabilitySwift (4.3.0)
  - SQLite.swift (0.11.6):
    - SQLite.swift/standard (= 0.11.6)
  - SQLite.swift/standard (0.11.6)

royjit:
PR #1700  for the fix

palpatim:
This was released on [2.10.3](https://github.com/aws-amplify/aws-sdk-ios/releases/tag/2.10.3). Please let us know if you continue to experience problems.

jasonbobier:
It works for me now. Thanks

Jason

PR #1735, opened 2019-08-13
Fix signOut callback

royjit:
signOut with callback was not invoking the callback if we do not pass any options. This change resolves that.

*Issue #1717 

*Description of changes:*


palpatim:
Also, I see that we invoke the completion handler before actually performing the `signOut` [here](https://github.com/aws-amplify/aws-sdk-ios/pull/1735/files#diff-2e9d07f1fb81c2e69b88ea2508050330R570). Is that going to cause any problems?

PR #1741, opened 2019-08-15
Cleaning up AWSMobileClient code

royjit:
*Description of changes:*
- Moved json encode/decode to separate helper class
- Moved error related helper method to AWSMobileClientError extension
- Rearranged methods in AWSMobileClient 

### DIFF
--- a/AWSAuthSDK/Sources/AWSMobileClient/AWSMobileResults.swift
+++ b/AWSAuthSDK/Sources/AWSMobileClient/AWSMobileResults.swift
@@ -118,7 +118,7 @@ extension AWSMobileClientError {
     }
     
     /// Utility class which maps a service error to `AWSMobileClientError`. If no mapping is available, returns the original error.
-    public static func getMobileError(for error: Error) -> Error {
+    public static func makeMobileClientError(from error: Error) -> Error {
         if error._domain == AWSCognitoIdentityProviderErrorDomain {
             if error._code == -3000 {
                 return AWSMobileClientError.deviceNotRemembered(message: "This device does not have an id, either it was never tracked or previously forgotten.")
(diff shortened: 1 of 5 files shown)


AWSAuthSDK/Sources/AWSMobileClient/AWSMobileResults.swift
```diff
@@ -66,9 +66,76 @@ public enum AWSMobileClientError: Error {
...
+        }
+        return error
+    }
     
-    static func CognitoAuthErrorMappingHelper(error: AWSCognitoAuthClientErrorType) -> AWSMobileClientError {
+    public static func getMobileError(cognitoAuthError: AWSCognitoAuthClientErrorType) -> AWSMobileClientError {
+        return AWSMobileClientError.CognitoAuthErrorMappingHelper(error: cognitoAuthError)
+    }
+    
+    public static func CognitoAuthErrorMappingHelper(error: AWSCognitoAuthClientErrorType) -> AWSMobileClientError {
```
palpatim: Now that we're making this public, let's take the opportunity to rename it to something that doesn't sound like a class constructor. Maybe both of these could be:

```swift
public static func makeMobileClientError(from error: AWSCognitoAuthClientErrorType) -> AWSMobileClientError { ... }

public static func makeMobileClientError(fromErrorCode errorCode: String....

```

royjit: Renamed, keeping it as static initializer.  `public static func makeMobileClientError(from error: Error) -> Error` returns an `Error` if not mapping is found so cannot be made into a init method.

AWSAuthSDK/Sources/AWSMobileClient/Internal/JSONHelper.swift
```diff
@@ -0,0 +1,61 @@
...
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+//
+//  JSONHelper.swift
+//  AWSMobileClient
+//
+
+import Foundation
+
+
```
palpatim: nit: remove extra vertical whitespace

royjit: Removed

Issue #1512, opened 2019-04-29
Loosing network during AWSS3 uploads does not properly reset progress reporting.


jeffjvick:
**Describe the bug**
I have noticed that if while you are uploading you disable the network and then reenable it (causing the currently upload parts to restart) the progressBlock stops getting updates until the progress has surpassed the point where it left off.  This is true for both single part and multiple-part uploads.  I have not tested downloads.

**To Reproduce**
This is especially evident if you do a **non**-multipart upload. So for example you are uploading a 50 mb file and at 30% you turn on/of the network, it will restart the upload from he beginning but the progress will just sit at 30% until the restarted upload passes back over 30% From there it will continue on as normal.  Obviously this would be very confusing to the user as to what is happening.

**Which AWS service(s) are affected?**
AWSS3

**Expected behavior**
`transferUtilityUploadTask.progress.completedUnitCount` should adjust appropriately back down after upload parts are restarted.  To reflect the current % completed not just sit at the last % point.

**Screenshots**
NA

**Environment(please complete the following information):**
 - SDK Version: 2.9.7
 - Dependency Manager: Cocoapods
 - Swift Version : 4.0

**Device Information (please complete the following information):**
 - Device: iPhone 7 plus
 - iOS Version: iOS 12.2
 - Specific to simulators:  You should only be testing on real devices and not have them connected to the debugger to simulate real world behavior.

**Additional context**

This is referenced from: https://github.com/aws-amplify/aws-sdk-ios/issues/769
@cbommas 

I looked into the AWS3TransferUtility.m code and see the issue. It is only updating transferUtilityUploadTask.progress.completedUnitCount if it is less than'totalBytesSent but this doesn't take into account if totalBytesSent gets reset when parts (or the whole) upload needs to get restarted. I think the easiest way to fix this would be to just use !=. I'm not sure why you have <. I made these two changes and it is working:

```
if (transferUtilityUploadTask.progress.completedUnitCount != totalBytesSent) {
    transferUtilityUploadTask.progress.completedUnitCount = totalBytesSent;
    
    if (transferUtilityUploadTask.expression.progressBlock) {
        transferUtilityUploadTask.expression.progressBlock(transferUtilityUploadTask, transferUtilityUploadTask.progress);
    }
}
```

```
if (transferUtilityMultiPartUploadTask.progress.completedUnitCount != totalSentSoFar ) {
[...]

PR #1813, opened 2019-08-23
AWSS3TransferUtility Uploads - Send Progress for all values including failed and restarted upload progresses

lawmicha:
*Issue #, if available:*
1512

*Description of changes:*
When an upload starts makes some progress, say 50%, fails due to network issue and restarts at 0%, the didSendBodyData method is called again from the beginning of the amount of bytes sent, which represents the upload has restarted. This change allows the progressBlock to be called whenever the value is different instead of checking only when the last known progress is less than current, so that it is better represents the progress of the work.


lawmicha:
manual verification was done https://github.com/aws-amplify/aws-sdk-ios/issues/1512#issuecomment-524116795

PR #1824, opened 2019-08-27
AWSMobile Client fixes - Balance semaphore's wait and signal calls.

royjit:
*Description of changes:* 

1. GetIdentity inside `AWSIdentityProvider.m` process one request at a time. If a new request comes in while we are processing another, we ask the new request to wait using semaphore `dispatch_semaphore_wait(self.semaphore, dispatch_time(DISPATCH_TIME_NOW, 5 * NSEC_PER_SEC));`.

This semaphore is signaled when we receive a request back for identity `dispatch_semaphore_signal(self.semaphore);`. But in the case where there is only one getIdentityId request, we signal the semaphore without calling a wait and thus increasing the semaphore value. So now when a new concurrent getIdentity is encountered, the semaphore is not decreased to a negative value and it wont wait.

To fix this, I changed made a call to wait semaphore with 0 second for a single request.

2. Call signOut doesnot make federationProvider to .none. This is an intermittent race condition.
Because of this calls to getCredentials goes to the federationProvider getSession and waits for user to enter signIn details. Our integration tests fails because of this situation.

### DIFF
--- a/AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClientExtensions.swift
+++ b/AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClientExtensions.swift
@@ -543,7 +543,6 @@ extension AWSMobileClient {
         self.saveLoginsMapInKeychain()
         self.setLoginProviderMetadataAndSaveInKeychain(provider: provider)
         self.internalCredentialsProvider?.clearCredentials()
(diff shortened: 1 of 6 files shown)


palpatim:
As discussed, also verify that `signOut` is clearing `identityId`.

PR #1826, opened 2019-08-29
Add CHANGELOG entry for S3 progress reporting fix

palpatim:
(Also changed some stray tabs to spaces in older parts of the file)


lawmicha:
This looks like something I missed adding with my PR. Thanks for the fix!

PR #1836, opened 2019-09-04
Changing intValue to longLongValue

benmckillop:
I was having an issue that if the file being uploaded is bigger than 2323693218 bytes, then the intValue call results in an overflow and the comparison fails. The change to longLongValue prevents the overflow.

*Description of changes:*
Changed the length of bytes sent so far in a multipart task to be of type longLongValue rather than intValue that it was previously

### DIFF
--- a/AWSS3/AWSS3TransferUtility.m
+++ b/AWSS3/AWSS3TransferUtility.m
@@ -2230,7 +2230,7 @@ didCompleteWithError:(NSError *)error {
                     totalBytesSent += aSubTask.totalBytesExpectedToSend;
                 }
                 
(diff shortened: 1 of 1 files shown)


royjit:
Thank you @benmckillop for the PR, this is merged to develop branch. Should be available in the next release of the SDK.

lawmicha:
Interesting! Looks like this is only for progress reporting in the multipart upload flow. What about upload and download? I think if you now try to download the large file, it should fail with the same error right?

palpatim:
We released this in https://github.com/aws-amplify/aws-sdk-ios/releases/2.11.0

PR #1855, opened 2019-09-15
Remove duplicate import

thii:


palpatim:
Thank you for this PR. This class is actually automatically generated by an internal process, so we'll have to adjust that code rather than merge this. I'll close this for now, but be sure to reference it when we merge in the generated changes.

palpatim:
Covered by #1925

PR #1856, opened 2019-09-15
[ec2] Update models to latest

awsmobilesdk:
[ec2] Update models to latest


palpatim:
Certain enum values emitted by the service have changed. Shouldn't need a breaking change, since the symbolic values are the same, but worth a callout in the CHANGELOG/release notes for customers who may be using old service clients.

palpatim:
Note: Original push for this PR had an incorrect commit message. Modified locally & force-pushed to fix.

PR #1859, opened 2019-09-17
Update CHANGELOG

palpatim:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 # AWS Mobile SDK for iOS CHANGELOG
 
+## 2.11.1
+
+### Misc. Updates
+
+- Model updates for the following services
+  - Amazon EC2
+    - NOTE: This model update includes a change to the mapping of certain service-emitted enum values. The symbols to which these values map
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
...
 
+## 2.11.1
+
+### Misc. Updates
+
+- Model updates for the following services
+  - Amazon EC2
+    - NOTE: This model update includes a change to the mapping of certain service-emitted enum values. The symbols to which these values map
+      remains the same, but customers using older versions of the AWSEC2 SDK may wish to upgrade. Enum values affected:
+      - "deleted-running" is now "deleted_running"
```
royjit: Shouldnt be this a breaking change?

palpatim: It would be breaking if upgrading to this version of the SDK required app code changes. This is the reverse--an upgrade to this SDK is required to make an app compatible with the service change; no app code change required, because the symbols to which the raw service-emitted enums are mapped have not changed.

PR #1927, opened 2019-09-19
Fixed configuration unit test failures

palpatim:
Some of the AWSCore tests were inappropriately categorized as integration tests, and would therefore be allowed to retry (e.g., https://circleci.com/gh/aws-amplify/aws-sdk-ios/13146). This change moves those tests into UnitTest configs so that they are executed as part of each build, and not allowed to retry on failure.

- Moved some Core tests into unit test target
- Moved config tests into their own test targets since AWSInfo is a singleton,
  and mocking isn't appropriate since we're explicitly testing the singleton's
  initialization behavior.

*Issue #, if available:*

*Description of changes:*


palpatim:
Fixed a few more config errors in an update, but nothing major structurally.

Issue #1994, opened 2019-10-16
Crash: AWSS3TransferUtility.m line 1074


AMORFATI-TRL:
**Describe the bug**
I've had this crash for a while running different versions of the SDK but it's still happening in version **2.11.0**
I noticed it was reported as fixed in #1249 but it's still happening in a different line number. 

Fatal Exception: NSInvalidArgumentException
Cannot read file at file:///var/mobile/Containers/Data/Application/EEF5A5F0-359D-49EE-8548-928378FA82C9/Library/Application%20Support/MeerkatSessionManager/sessions/89ADDE21-8D5B-4761-881F-3278AACBD982.1571162648

**To Reproduce**
Can't easily reproduce on my test devices as it doesn't happen always 

**Which AWS service(s) are affected?**
S3

**Expected behavior**
Not crashing 

**Screenshots**
NA

**Environment(please complete the following information):**
 - SDK Version: 2.11.0
 - Dependency Manager: Cocoapods
 - Swift Version : 5.1

**Device Information (please complete the following information):**
 - Device: iPhone XR
 - iOS Version: iOS 13.1.2

**Additional context**
```
Fatal Exception: NSInvalidArgumentException
0  CoreFoundation                 0x1a315dc30 __exceptionPreprocess
1  libobjc.A.dylib                0x1a2e780c8 objc_exception_throw
2  CFNetwork                      0x1a645ed20 (Missing)
3  CFNetwork                      0x1a645da7c (Missing)
4  CFNetwork                      0x1a645e3d0 (Missing)
5  CFNetwork                      0x1a645eae4 cfnTranslateCFError
6  libdispatch.dylib              0x1a2e0500c _dispatch_client_callout
7  libdispatch.dylib              0x1a2e11804 _dispatch_lane_barrier_sync_invoke_and_complete
8  CFNetwork                      0x1a63a2038 (Missing)
9  CFNetwork                      0x1a645e334 (Missing)
10 AWSS3                          0x10318710c __55-[AWSS3TransferUtility createUploadTask:startTransfer:]_block_invoke + 1074 (AWSS3TransferUtility.m:1074)
11 AWSCore                        0x1030567f8 __56-[AWSTask continueWithExecutor:block:cancellationToken:]_block_invoke + 331 (AWSTask.m:331)
12 AWSCore                        0x103025288 __30+[AWSExecutor defaultExecutor]_block_invoke_2 + 66 (AWSExecutor.m:66)
[...]

PR #2003, opened 2019-10-18
Fix Crash in S3 upload data

royjit:
*Issue : #1994 

### DIFF
--- a/AWSS3/AWSS3TransferUtility.m
+++ b/AWSS3/AWSS3TransferUtility.m
@@ -1038,7 +1038,8 @@ static AWSS3TransferUtility *_defaultS3TransferUtility = nil;
 }
 
 
--(AWSTask<AWSS3TransferUtilityUploadTask *> *) createUploadTask: (AWSS3TransferUtilityUploadTask *) transferUtilityUploadTask startTransfer:(BOOL) startTransfer {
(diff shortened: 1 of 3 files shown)


AWSS3/AWSS3TransferUtility.m
```diff
@@ -1071,9 +1072,29 @@ - (AWSS3TransferUtilityUploadSubTask * ) hydrateMultiPartUploadSubTask:(NSMutabl
...
-                                                                        fromFile:[NSURL fileURLWithPath:transferUtilityUploadTask.file]];
-        transferUtilityUploadTask.sessionTask = uploadTask;
+        NSURLSessionUploadTask *nsURLUploadTask = nil;
+        NSString *exceptionReason = @"";
+        @try {
+            nsURLUploadTask  = [self.session uploadTaskWithRequest:request
+            fromFile:[NSURL fileURLWithPath:transferUtilityUploadTask.file]];
+        }
+        @catch (NSException *exception) {
+            AWSDDLogDebug(@"Exception in upload task %@", exception.debugDescription);
```
palpatim: Log this as error instead of debug?

royjit: Changed to `AWSDDLogWarn` because we are not doing anything inside the exception block other than passing the value to the error.

PR #2031, opened 2019-11-07
[pinpoint] Update models to latest

awsmobilesdk:
[pinpoint] Update models to latest


AWSPinpoint/AWSPinpointTargeting/AWSPinpointTargetingModel.m
```diff
@@ -360,6 +361,46 @@ + (NSValueTransformer *)itemJSONTransformer {
 
 @end
 
+@implementation AWSPinpointTargetingActivity
+
++ (NSDictionary *)JSONKeyPathsByPropertyKey {
+	return @{
+             @"conditionalSplit" : @"ConditionalSplit",
+             @"detail" : @"Description",
+             @"EMAIL" : @"EMAIL",
```
palpatim: This is an unusual case for a property definition. Checking with service team to ensure this was intentional, but it is how the service defined the property: https://docs.aws.amazon.com/pinpoint/latest/apireference/apps-application-id-journeys-journey-id.html#apps-application-id-journeys-journey-id-properties

PR #2052, opened 2019-11-18
Improve error reporting from ec2

### LINKED ISSUE #1062: How to get the raw/full error message in ec2?
**State your question**
Hi,
I'm wondering how to get the full error message as it is in the xml response of a failed ec2 request? E.g. 
`<?xml version="1.0" encoding="UTF-8"?>
<Response><Errors><Error><Code>InvalidParameterValue</Code><Message>Invalid value 'i-011111111111' for instanceId. Instance does not have a volume attached at root (/dev/xvda)</Message></Error></Errors><RequestID>some-id</RequestID></Response>`

The problem is that the error provided in the completionblock (result is nil) locks like this `Error Domain=com.amazonaws.AWSEC2ErrorDomain Code=0 "(null)` but I want the error message as it is in the xml.

**Which AWS Services are you utilizing?**
EC2

**Provide code snippets (if applicable)**

**Environment(please complete the following information):**
 - SDK Version: 2.6.20
 - Dependency Manager: Cocoapods
 - Swift Version : 4.0

**Device Information (please complete the following information):**
 - Device: [e.g. iPhone6, Simulator]
 - iOS Version: [e.g. iOS 11.4]
 - Specific to simulators:
Non specific device/iOS version issue

Any help is appreciated.

Thanks
-Sal

coredumped:
Fixes #1062 

Sets userInfo as delegate.responseObject which contains the full error object returned by the API as found by @ssferrazza


AWSCore/Networking/AWSURLSessionManager.m
```diff
@@ -323,7 +323,12 @@ - (void)URLSession:(NSURLSession *)session task:(NSURLSessionTask *)sessionTask
                                                                                                         data:delegate.responseData
                                                                                                        error:&error];
                     if (error) {
-                        delegate.error = error;
+                        if([delegate.responseObject isKindOfClass:[NSDictionary class]]) {
+                            delegate.error = [NSError errorWithDomain:AWSNetworkingErrorDomain code:AWSNetworkingErrorUnknown userInfo: ((NSDictionary *)delegate.responseObject)[@"Error"]];
```
palpatim: We should persist the error domain & code from the original error, to make this a purely additive change

```objc
if ([delegate.responseObject isKindOfClass:[NSDictionary class]]) {
    delegate.error = [NSError errorWithDomain:error.domain
                                         code:error.code
                                     userInfo:((NSDictionary *)delegate.responseObject)[@"Error"]];
}
```

AWSCore/Networking/AWSURLSessionManager.m
```diff
@@ -323,7 +323,12 @@ - (void)URLSession:(NSURLSession *)session task:(NSURLSessionTask *)sessionTask
                                                                                                         data:delegate.responseData
                                                                                                        error:&error];
                     if (error) {
-                        delegate.error = error;
+                        if([delegate.responseObject isKindOfClass:[NSDictionary class]]) {
```
palpatim: nit: space between `if` and opening parenthesis

palpatim:
I realized only after I merged this that this change will overwrite any existing `userInfo`, which is not the behavior we want. I'll add an additional change to propagate the response object under a known key of `userInfo` rather than altogether replacing the existing `userInfo`.

coredumped:
Interesting but also please note that at that point for some reason userInfo on the original error is already nil.
Take a look at #1062

palpatim:
It might be nil for some cases (notably the EC2 case described in #1062), but since this is core code, we can't rely on every implementation of `-[AWSHTTPURLResponseSerializer responseObjectForResponse:originalRequest:currentRequest:data:error:]` to behave the same.

coredumped:
Awesome! 
Thanks for accepting this small contribution BTW

PR #2102, opened 2019-12-10
KinesisVideoSignaling CI/CD integration

palpatim:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/CircleciScripts/IntegrationTestsConfiguration.json
+++ b/CircleciScripts/IntegrationTestsConfiguration.json
@@ -110,6 +110,12 @@
       "AWSKinesisVideoTests": {}
     }
   },
+  "KinesisVideoSignaling": {
+    "test_list": {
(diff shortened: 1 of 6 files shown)


CircleciScripts/IntegrationTestsConfiguration.json
```diff
@@ -110,6 +110,12 @@
       "AWSKinesisVideoTests": {}
     }
   },
+  "KinesisVideoSignaling": {
+    "test_list": {
+      "AWSKinesisVideoSignalingTests": {},
+      "AWSKinesisVideoSignalingUnitTests": {}
```
lawmicha: looks like the run_unittest_bundle runs the unit test so this can probably be removed? it doesn't look like the other sections in this file are listing the unit test targets either

palpatim: Whoops, good catch, removed

PR #2152, opened 2019-12-13
Use cache key prefix to enable rebuild without pushing new code

palpatim:
*Issue #, if available:*

Setting a cache key prefix to allow us to kick off new builds without a code push. Useful for cases where, e.g., stale test credentials are causing test errors.

Longer term, we should move test credentials and configs out of the cache anyway; there's no reason we shouldn't be pulling those in at test execution time rather than at build time.

*Description of changes:*

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,6 +34,9 @@ commands:
       set environment variables
     steps:
       - run:
+          name: set cache key prefix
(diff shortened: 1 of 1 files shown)


palpatim:
Not working as expected, don't want to spend more time troubleshooting since this is a bit of a hacky fix anyway

PR #2234, opened 2020-01-17
Fix some CI/CD timeout & failures

palpatim:
*Description of changes:*

- `carthage archive` times out occasionally attempting to scan the current source root for buildable schemes. This change causes the archive to be built from hardcoded package names
- Reorganized the list of dependent packages into a shared python module, and ordered them to attempt to minimize the number of CocoaPods release failures we get because of the CDN not being updated immediately after a `trunk push`.

Future enhancements:
- use the package list for all other places where we currently hardcode package names: integ tests, etc
- model dependencies in the package list, rather than simply using the order they're listed in the file
- Parallelize Carthage build step
- Parallelize CocoaPods release by pushing to trunk, and waiting for them to become available on the CDN before building packages that depend on them. The service-SDK pods should be fully parallelizable, which would save quite a bit of time during this process.


AWSMobileAnalytics.podspec
```diff
@@ -15,4 +15,7 @@ Pod::Spec.new do |s|
   s.dependency 'AWSCore', '2.12.6'
   s.source_files = 'AWSMobileAnalytics/*.{h,m}', 'AWSMobileAnalytics/**/*.{h,m}'
   s.private_header_files = 'AWSMobileAnalytics/Internal/*.h'
+
+  s.deprecated = true
```
royjit: Should we add this to changelog ?

palpatim: It was already deprecated, this just deprecates the Podspec, but it's probably a good idea to make an explicit note in the changelog.

PR #2254, opened 2020-01-25
Fix NSInternalInconsistencyException on new password required flow

wooj2:
*Description of changes:*
Here's what seems to be happening:
When we get an error back, we are attempting to push a replacement AWSUserPoolNewPasswordRequiredViewController onto the nav stack at the same time we are presenting a UIAlertView. The UIAlertView is preventing the new AWSUserPoolNewPasswordRequiredViewController from appearing on the stack. The ViewControllers and (Bolts)Tasks are currently coupled together, so we are crashing because we are attempting to re-run the completed task that was attached to the stale view controller.

Instead, we should, seems like the right thing to do is the following three items (which I have pushed into the PR):

1.  Remove that old AWSUserPoolNewPasswordRequiredViewController (and therefore the completed task will also be removed) from the navigation stack
2.  Present the new AWSUserPoolNewPasswordRequiredViewController (with the old config) in place of the old AWSUserPoolNewPasswordRequiredViewController that is being removed
3.  Presenting the alert view off of the newly created AWSUserPoolNewPasswordRequiredViewController (which becomes the top view controller)


AWSCognitoIdentityProvider/AWSCognitoIdentityUser.m
```diff
@@ -376,17 +377,22 @@ -(instancetype) initWithUsername: (NSString *)username pool:(AWSCognitoIdentityU
...
+                                                                                                                                       requiredAttributes:requiredAttributesSet];
         AWSTaskCompletionSource<AWSCognitoIdentityNewPasswordRequiredDetails *> *newPasswordRequiredDetails = [AWSTaskCompletionSource<AWSCognitoIdentityNewPasswordRequiredDetails *> new];
         [newPasswordRequiredDelegate getNewPasswordDetails:newPasswordRequiredInput newPasswordRequiredCompletionSource:newPasswordRequiredDetails];
+        if (lastError) {
+            [newPasswordRequiredDelegate didCompleteNewPasswordStepWithError:lastError];
+        }
         return [[newPasswordRequiredDetails.task continueWithSuccessBlock:^id _Nullable(AWSTask<AWSCognitoIdentityNewPasswordRequiredDetails *> * _Nonnull task) {
             return [self performRespondToNewPasswordChallenge:task.result session:lastChallenge.session];
         }] continueWithBlock:^id _Nullable(AWSTask<AWSCognitoIdentityProviderRespondToAuthChallengeResponse *> * _Nonnull task) {
-            [newPasswordRequiredDelegate didCompleteNewPasswordStepWithError:task.error];
```
royjit: This will break the [AWSMobileClient logic](https://github.com/aws-amplify/aws-sdk-ios/blob/master/AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClientExtensions.swift#L872) which uses this delegate callback to return an error back to the user.

Also this will break the existing users of the old Cognito SDK.

wooj2: Thanks so much for bringing this up! You are indeed correct that it would have broken customers.  Thanks again.

I've come up with different approach which leaves the (old) view controller, on the navigation stack until the UIAlertView goes away.  I don't think this is any worse than what we are currently doing since any existing customers depending on existing functionality would have already had to work around the fact that we are pushing a view controller on the stack while previously calling `didCompleteNewPasswordStepWithError:`.  That being said, happy to know what you think about this approach.

Thanks!

AWSAuthSDK/Sources/AWSUserPoolsSignIn/UserPoolsUI/AWSUserPoolNewPasswordRequiredViewController.m
```diff
@@ -93,21 +117,13 @@ -(void) getNewPasswordDetails: (AWSCognitoIdentityNewPasswordRequiredInput *) ne
...
-                                                                                     message:error.userInfo[@"message"]
-                                                                              preferredStyle:UIAlertControllerStyleAlert];
-            UIAlertAction *ok = [UIAlertAction actionWithTitle:@"Ok" style:UIAlertActionStyleDefault handler:nil];
-            [alertController addAction:ok];
-            [self presentViewController:alertController
-                               animated:YES
-                             completion:nil];
-        }
-    });
+    self.errorOfCompletedTask = error;
```
wooj2: This approach is not ideal, but given that viewControllers are attached to tasks, we are assuming that if this method `didCompleteNewPasswordStepWithError` is called, we are assuming a new view controller will be presented on the screen (via the retry logic in `startNewPasswordRequiredUI`), and therefore, viewDidDisappear is expected to called (and we can present the error at that time).

Issue #2261, opened 2020-01-28
Custom auth callback not reset on `signOut`


MightyFine:
The tl;dr is that when I call `AWSMobileClient.default().signOut()` there's a custom auth challenge task (`UserPoolOperationsHandler.sharedInstance.customAuthChallengeTaskCompletionSource` which is an internal API) that isn't reset, so when I subsequently call `AWSMobileClient.default().signIn(...)` it tries to sign in via a custom auth challenge and then I get an error "Missing required parameter ANSWER".

The reason I'm getting into this state is that I'm working on a feature where the flow is that a user can enter a phone number to sign in, which is implemented on AWS as a custom auth challenge which sends a code, which the user then inputs to confirm, and they're signed in.

Part of the requirement is that the user can tap on a "I didn't receive a code" button, which takes them back, and they can change their phone number or just sign in again to retrigger the flow.

The problem I'm running into is that once `AWSMobileClient` signs in and is issued a custom auth challenge, it sets a `customAuthChallengeTaskCompletionSource` task, which is then what's used during `confirmSignIn` to perform the confirmation. I can see that this is also used in `signIn` if it is populated, and after it's used is the only time that it's set back to `nil`.

The problem that this poses is that if  `AWSMobileClient.default().signOut()` is called, there's a flow internally which checks for populated callbacks and eventually calls `invalidateSignInCallbacks`, which sets the sign in callbacks to nil, but the auth tasks aren't included in this and don't have any apparent way to be reset.

It seems to me like `signOut()` would also be the place to reset the `customAuthChallengeTaskCompletionSource` value and the other auth tasks, since if `signOut()` is invoked, it doesn't seem to make sense for the API to still be in a state where it's going to try and fill a custom auth challenge when a user signs in again?

There's a good chance I'm misunderstanding the SDK and how it works internally, so if anyone cares to correct me, please do, and if anyone has a workaround or a way this is meant to be used, I'd appreciate it!

It's worth noting that aside from this one edge case, the rest of the custom auth works properly, so our implementation is doing what it's meant to do aside from this.

Thanks,
Tim

kneekey23:
hi @MightyFine I reproduced this bug pretty quickly and just submitted a PR for it as you can see above. This seems to correct this issue, I will get this merged in as soon as a CR takes place. Thanks!

MightyFine:
Hey @kneekey23, that's awesome, thank you!

kneekey23:
[...]

PR #2265, opened 2020-01-30
Clear CustomAuthChallenge Task on SignOut

kneekey23:
*Issue #, if available:* Providing a fix for #2261 

*Description of changes:*
CustomAuthChallengeTask does not clear on sign out so if sign in challenge wasn't completed and sign out was called first, you receive an error of "Missing required parameter ANSWER"


wooj2:
1.  Commits for this package should be against develop
2.  please update change log
3.  Code looks fine, but... maybe Jithin could take a look, as he has modified custom auth more than I have :)

royjit:
LGTM

Issue #2260, opened 2020-01-27
'AWSMobileClient-Swift.h' not found with CocoaPods as static libraries 


lazy-var:
**Describe the bug**
When building a project with Pods as [static libraries](http://blog.cocoapods.org/CocoaPods-1.5.0/) by not including `use_frameworks!` in the `Podfile`, but adding `use_modular_headers!` or `:modular_headers => true` to a single `pod` I get following error:

> 'AWSMobileClient/AWSMobileClient-Swift.h' file not found

**Which AWS service(s) are affected?**
Local AWSMobileClient pod. Not aws services affected.

**Environment(please complete the following information):**
 - SDK Version: pod 'AWSMobileClient', '~> 2.12.2'
 - Dependency Manager: Cocoapods
 - Swift Version : 5.x
 - Target: iOS 13.0
 - Xcode: 11.3.1

** Code related to the issue**
https://github.com/aws-amplify/aws-sdk-ios/blob/b64ed76724f662239f3f5e8df0c40745098626b1/AWSAuthSDK/Sources/AWSMobileClient/Internal/_AWSMobileClient.m#L18

**Possible fix?**
```
-- #import <AWSMobileClient/AWSMobileClient-Swift.h>
++ #import "AWSMobileClient-Swift.h"
```
As `_AWSMobileClient.m` already lives on `AWSMobileClient/`. 
When applying this, project compiles and runs apparently fine.

**Additional context**
I couldn't find a `AWSMobileClient-Swift.h` file inside `AWSMobileClient` directory:

![_AWSMobileCLient m](https://user-images.githubusercontent.com/2856396/73174316-bb3af080-4107-11ea-9db7-d05c1954dfc3.png)

**Related issue**
https://github.com/aws-amplify/aws-sdk-ios/issues/1126

wooj2:
Hi,
Thank you so much for reporting this.

I did some research and found a potential workaround here: https://github.com/CocoaPods/CocoaPods/issues/7594 , and have submitted a PR here: https://github.com/aws-amplify/aws-sdk-ios/pull/2267 .  Please be warned that, I haven't exhaustively tested the implications of compiling each and every library statically, so if/when this PR gets pushed into an official release, please use caution and your best judgement in using this feature.

Thanks again!

lazy-var:
Thank you! I couldn't find that solution.

rajuashok:
I'm still having this issue, but based on this it should be fixed right?

[...]

PR #2267, opened 2020-02-01
Allow for static lib generation using workaround for swift/obj-c mixed projects

wooj2:
I am attempting to adopt a workaround that was discussed on a Cocoapods issue ( https://github.com/CocoaPods/CocoaPods/issues/7594 ), which allows for mixed objc/swift projects to be compiled for both dynamic and static libs under cocoapods.  I currently don't have a way to exhaustively test the implications of compiling/loading all of these libraries statically into a project, so please use good judgement when using this feature.

I tested compiling & running with a sample auth app on my local box w/ a `Podfile` containing:
```
  pod 'AWSMobileClient', :path => '/<redacted>/aws-sdk-ios'
  pod 'AWSAuthUI', :path => '/<redacted>/aws-sdk-ios'
  pod 'AWSUserPoolsSignIn', :path => '/<redacted>/aws-sdk-ios'
  pod 'AWSCognitoIdentityProvider', :path => '/<redacted>/aws-sdk-ios'
```
With case 1:
```
  use_modular_headers!
```
and case 2:
```
use_frameworks!
```

Both cases seem to compile and execute basic boilerplate calls into the framework.  I didn't observe any crashes, nor any abnormal behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@
 ### Misc. Updates
 
 - Updated **AWSMobileAnalytics** podspec with `deprecated` and `deprecated_in_favor_of` attributes
-
+- Added workaround for `use_modular_headers!` inside of Podfile (experimental)
 - Model updates for the following services
   - Amazon EC2
   - AWS KMS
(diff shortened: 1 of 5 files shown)


wooj2:
Original request: https://github.com/aws-amplify/aws-sdk-ios/issues/2260

CHANGELOG.md
```diff
@@ -15,7 +15,7 @@
 ### Misc. Updates
 
 - Updated **AWSMobileAnalytics** podspec with `deprecated` and `deprecated_in_favor_of` attributes
-
+- Added workaround for `use_modular_headers!` inside of Podfile (experimental)
```
drochetti: do we need documentation updates for this?

wooj2: Are you suggesting that we need to advertise support for modular headers in our documents(or something else?) ?  Using modular headers is a feature provided by  CocoaPods, and I think it is up to the user if they want to use it or not.

PR #2281, opened 2020-02-07
Copy x86_64-apple-ios-simulator swiftmodule files for AWSMobileClient…

wooj2:
Tested locally on my box by importing in generated build artifacts into a new project file.  Made sure to compile and run code that was calling into the AWSMobileClient.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,4 @@
 # AWS Mobile SDK for iOS CHANGELOG
-## Unreleased
-- Fixed issue 2274 where we were not packaging x86_64-apple-ios-simulator.swiftmodule files for binary releases
 
 ## 2.12.7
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,4 +1,6 @@
 # AWS Mobile SDK for iOS CHANGELOG
+## Unreleased
+- Fixed issue 2274 where we were not packaging x86_64-apple-ios-simulator.swiftmodule files for binary releases
```
palpatim: nit: include issue link in changelog notes. See previous examples.

CircleciScripts/package_sdk.sh
```diff
@@ -74,6 +74,7 @@ mkdir -p $FRAMEWORK_DIR
 echo "copy framework"
 echo "builtFramework/Release-iphoneos/${project_name}.framework $FRAMEWORK_DIR"
```
palpatim: nit: remove, we should use `set -v` if we want to echo commands to console as we execute

wooj2: Updated.

CHANGELOG.md
wooj2: Updated.

PR #2284, opened 2020-02-10
[iot] Update models to latest

awsmobilesdk:
[iot] Update models to latest


palpatim:
This changes the "resources" by updating the `ThrottlingException` to reflect the actual system behavior of a 400 HTTP status code, rather than the previously documented 429. There is no system behavior change associated with this.

PR #2324, opened 2020-03-01
CHANGELOG for 2020-03-01-08-15 model updates

awsmobilesdk:
CHANGELOG for 2020-03-01-08-15 model updates


lawmicha:
latest changelog update containis this

PR #2326, opened 2020-03-02
Fix test in IoT

royjit:
Test was using a local variable to keep track of the message received count. This change makes it
to use the XCTestExpectation  to track the messages.

### DIFF
--- a/AWSIoTTests/AWSIoTDataManagerTests.swift
+++ b/AWSIoTTests/AWSIoTDataManagerTests.swift
@@ -670,7 +670,6 @@ class AWSIoTDataManagerTests: XCTestCase {
     }
 
     func publishSubscribeWithCert(useALPN:Bool) {
-        var messageCount = 0
         var connected = false
         let hasConnected = self.expectation(description: "MQTT connection has been established")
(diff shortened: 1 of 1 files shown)


AWSIoTTests/AWSIoTDataManagerTests.swift
```diff
@@ -741,21 +740,18 @@ class AWSIoTDataManagerTests: XCTestCase {
...
         let testTopic = "TestTopic"
-
+        gotMessage.expectedFulfillmentCount = 5
         //Subscribe to TestTopic
         iotDataManager.subscribe(toTopic: testTopic, qoS: .messageDeliveryAttemptedAtLeastOnce, messageCallback: {
             (payload) ->Void in
             let stringValue:String = NSString(data: payload, encoding: String.Encoding.utf8.rawValue)! as String
             print("received: \(stringValue)")
             XCTAssertEqual(testMessage, stringValue)
-            messageCount = messageCount+1
```
lawmicha: looks like the problem with the old code is that there isn't a semphore on updating messageCount so it may never reach 5 even if 5 callbacks are triggered

PR #2333, opened 2020-03-03
adding xterm and changelog updates for 2.13.0

lawmicha:
this adds the `TERM: xterm-256color` environment for places which is using `configure_aws` step that installs aws cli. we noticed this was installing 2.0.0 that started breaking in a later step with

```
WARNING: terminal is not fully functional
-  (press RETURN)
```

Looks like this is the possible fix

ref https://discuss.circleci.com/t/on-macos-runners-warning-terminal-is-not-fully-functional/33656/5

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -284,6 +284,8 @@ jobs:
   pre_integrationtest:
     macos:
       xcode: "11.2.1"
+    environment: 
+      TERM: xterm-256color
     steps:
       - skip_job_unless_required
       - set_environment_variables
@@ -642,6 +644,8 @@ jobs:
   release_ios_s3:
     macos:
       xcode: "11.2.1"
+    environment:
+      TERM: xterm-256color
     steps:
       - skip_job_unless_required
       - checkout
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Bug Fixes
 
-- **Breaking Change** Updated nullability flags for the `AWSSignatureSignerUtility`, `AWSSignatureV4Signer`, `AWSSignatureV2Signer` and `AWSS3ChunkedEncodingInputStream` classes. Methods in these classes will have new signatures in Swift code, and may require code changes in your app to consume. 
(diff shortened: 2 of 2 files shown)


lawmicha:
line 479 
```
release_carthage:
    macos:
      xcode: "11.2.1"
    steps:
      - skip_job_unless_required
      - checkout
      - pre_start_simulator
      - update_brew
      - run:
          name: install github-release
          command: HOMEBREW_NO_AUTO_UPDATE=1 brew install github-release
      - run:
          name: install aws cli
          command: |
            HOMEBREW_NO_AUTO_UPDATE=1 brew install awscli
```

release_carthage also installs aws cli

.circleci/config.yml
```diff
@@ -284,6 +284,8 @@ jobs:
   pre_integrationtest:
     macos:
       xcode: "11.2.1"
+    environment: 
+      TERM: xterm-256color
```
lawmicha: not sure if we need this under `pre_integrationtest` since it's working without

lawmicha:
trying to hardcode awscli version to 2.0.0 as observed in the failing workflow's installation version,  testing locally `brew install awscli@2.0.0` 
`Error: No available formula with the name "awscli@2.0.0"`

.circleci/config.yml
lawmicha: it's very odd, a recent `pre_Integrationtest` run is installing awscli 1.6.260, not 2.0.0

`==> Downloading https://homebrew.bintray.com/bottles/awscli-1.16.260.catalina.bo`

lawmicha: added this in just in case it starts breaking in the future

palpatim: This flow uses the `s3` CLI command, vs. the `s3api` command. It's possible that it behaves differently, but it shouldn't hurt to explicitly specify the term setting here.

CHANGELOG.md
```diff
@@ -2,14 +2,10 @@
 
 ## 2.13.0
 
-### Misc. Updates
+### Bug Fixes
 
 - **Breaking Change** Updated nullability flags for the `AWSSignatureSignerUtility`, `AWSSignatureV4Signer`, `AWSSignatureV2Signer` and `AWSS3ChunkedEncodingInputStream` classes. Methods in these classes will have new signatures in Swift code, and may require code changes in your app to consume.
```
palpatim: While you're in here, please add a link to the PR & Issue that tracks this change.

PR #2387, opened 2020-03-26
Add AWS Pinpoint Journey Push Notification Handling

zhinang-amazon:
Changes:
- Update AWSPinpointNotificationManager to parse Journey Push similar to Campaign Push and add Journey related attributes to PutEvent API
- Update AWSPinpointAnalyticsClient to save either Campaign or Journey global attributes since they do not coexist. 
- Rename related non-public methods/properties so that they are not specific to Campaign
- Add AWSPinpointNotificationManagerTests to test changes in AWSPinpointNotificationManager
- Add test in AWSPinpointEventRecorderTests to test that Campaign and Journey attributes overwrite each other


AWSPinpoint/AWSPinpointSessionClient.m
```diff
@@ -72,9 +72,7 @@ @interface AWSPinpointSessionClient()
 @end
 
 @interface AWSPinpointAnalyticsClient()
-
-- (void)removeAllGlobalCampaignAttributes;
-- (void)setCampaignAttributes:(NSDictionary*) campaign;
+- (void)removeAllGlobalFeatureAttributes;
```
wooj2: I know you didn't introduce this, but seeing this here is kind of disappointing that we don't have a clean callback interface from our low-level session client all the way back up to the analytics client.  I'd expect something like the analytics client to conform to a protocol that allows to be notified of its downstream session being ended but.. hm. yea.. okay..  i guess we can refactor this later.

AWSPinpointTests/AWSPinpointEventRecorderTests.m
```diff
@@ -305,6 +305,83 @@ - (void) testUpdateSessionCampaignAttributes {
...
+                    XCTAssertNotNil(task.result);
+                    XCTAssertTrue([task.result isKindOfClass:[AWSPinpointEvent class]]);
+                    AWSPinpointEvent *sessionStartResult = task.result;
+                    XCTAssertTrue([sessionStartResult.eventType isEqualToString:@"_session.start"]);
+                    XCTAssertTrue([sessionStartResult.session.sessionId isEqualToString:pinpoint.sessionClient.session.sessionId]);
+                    XCTAssertTrue([sessionStartResult.session.startTime.description isEqualToString:pinpoint.sessionClient.session.startTime.description]);
+                    XCTAssertEqual(sessionStartResult.allAttributes.count, 1);
+                    XCTAssertTrue([sessionStartResult.allAttributes[@"campaignAttrKey"] isEqualToString:@"campaignAttrVal"]);
+                    
+                    // Overwrite  with journey attributes
```
wooj2: remove double space between `Overwrite` and `with`

AWSPinpointTests/AWSPinpointEventRecorderTests.m
```diff
@@ -305,6 +305,83 @@ - (void) testUpdateSessionCampaignAttributes {
     }];
 }
 
+- (void) testUpdateSessionOverwriteCampaignAttributesWithJourney {
```
wooj2: seems like we should have an additional test which mirrors testUpdateSessionCampaignAttributes, but is only for a journey

AWSPinpointTests/AWSPinpointNotificationManagerTests.m
```diff
@@ -0,0 +1,153 @@
...
+@property (nonatomic, strong) NSDictionary *journeyPushPayload;
+
+@end
+
+@implementation AWSPinpointNotificationManagerTests
+
+- (void)setUp {
+    [super setUp];
+
+    [[NSUserDefaults standardUserDefaults] removeSuiteNamed:@"AWSPinpointNotificationManagerTests"];
```
wooj2: define string in a const static NSString?

AWSPinpointTests/AWSPinpointNotificationManagerTests.m
```diff
@@ -0,0 +1,153 @@
...
+                     withPushActionType:(AWSPinpointPushActionType) pushActionType;
+@end
+
+@interface AWSPinpointConfiguration()
+@property (nonnull, strong) NSUserDefaults *userDefaults;
+@end
+
+@interface AWSPinpointNotificationManagerTests : XCTestCase
+@property (nonatomic, strong) AWSPinpoint *pinpoint;
+@property (nonatomic, strong) NSUserDefaults *userDefaults;
```
wooj2: Do we really need this need this as a property?  It looks like on line 60 this could just be:
```
configuration.userDefaults = [[NSUserDefaults alloc] initWithSuiteName:@"AWSPinpointNotificationManagerTests"];
```

AWSPinpointTests/AWSPinpointNotificationManagerTests.m
```diff
@@ -0,0 +1,153 @@
...
+@property (nonatomic, strong) NSDictionary *campaignPushPayload;
+@property (nonatomic, strong) NSDictionary *journeyPushPayload;
```
wooj2: Can we make these const some how?   perhaps even static const?

AWSPinpointTests/AWSPinpointNotificationManagerTests.m
```diff
@@ -0,0 +1,153 @@
...
+
+- (void)tearDown {
+    [self.pinpoint.analyticsClient.eventRecorder removeAllEvents];
+    [self.pinpoint.analyticsClient removeAllGlobalFeatureAttributes];
+    [super tearDown];
+}
+
+- (void)testAddPinpointMetadataForCampaignEvent {
+    AWSPinpointEvent *pushNotificationEvent = [self.pinpoint.analyticsClient createEventWithEventType:@"_campaign.received_foreground"];
+    [_pinpoint.notificationManager addPinpointMetadataForEvent:pushNotificationEvent
```
wooj2: is there any reason why we are using the iVar here and not `self.pinpoint` ? (along with everywhere else in this file)

AWSPinpointTests/AWSPinpointNotificationManagerTests.m
```diff
@@ -0,0 +1,153 @@
...
+- (void)testAddPinpointMetadataForJourneyEvent {
+    AWSPinpointEvent *pushNotificationEvent = [self.pinpoint.analyticsClient createEventWithEventType:@"_journey.received_background"];
+    [_pinpoint.notificationManager addPinpointMetadataForEvent:pushNotificationEvent
+                                              withNotification:self.journeyPushPayload];
+    XCTAssertTrue([pushNotificationEvent.eventType isEqualToString:@"_journey.received_background"]);
+    XCTAssertTrue([pushNotificationEvent.allAttributes[@"journey_id"] isEqualToString:@"testJourneyId"]);
+}
+
+- (void)testAddPinpointMetadataForUnknownEvent {
+    NSDictionary *userInfo = @{
```
wooj2: add const

AWSPinpointTests/AWSPinpointNotificationManagerTests.m
```diff
@@ -0,0 +1,153 @@
...
+    XCTAssertEqual(self.pinpoint.analyticsClient.globalFeatureAttributes[@"campaign_id"], @"testCampaignId");
+}
+
+- (void)testAddGlobalPinpointMetadataForJourneyNotification {
+    [_pinpoint.notificationManager addGlobalPinpointMetadataForNotification:self.journeyPushPayload];
+    XCTAssertEqual([self.pinpoint.analyticsClient.globalFeatureAttributes count], 1);
+    XCTAssertEqual(self.pinpoint.analyticsClient.globalFeatureAttributes[@"journey_id"], @"testJourneyId");
+}
+
+- (void)testBuildEventType {
```
wooj2: These should be broken up into separate, individual, isolated unit tests

AWSPinpointTests/AWSPinpointNotificationManagerTests.m
```diff
@@ -0,0 +1,153 @@
...
+
+@interface AWSPinpointNotificationManagerTests : XCTestCase
+@property (nonatomic, strong) AWSPinpoint *pinpoint;
+@property (nonatomic, strong) NSUserDefaults *userDefaults;
+@property (nonatomic, strong) NSDictionary *campaignPushPayload;
+@property (nonatomic, strong) NSDictionary *journeyPushPayload;
+
+@end
+
+@implementation AWSPinpointNotificationManagerTests
```
wooj2: Seems like these are unit tests?  Seems like these belong in the AWSPinpointUnitTests folder (and to the AWSPinpointUnitTests target)

AWSPinpointTests/AWSPinpointNotificationManagerTests.m
```diff
@@ -0,0 +1,153 @@
...
+                @"randomKey": @{
+                    @"randomAttribute": @"randomValue"
+                }
+            }
+        }
+    };
+    AWSPinpointEvent *pushNotificationEvent = [self.pinpoint.analyticsClient createEventWithEventType:@"_unknown.received_background"];
+    [_pinpoint.notificationManager addPinpointMetadataForEvent:pushNotificationEvent
+                                              withNotification:userInfo];
+    XCTAssertEqual([pushNotificationEvent.allAttributes count], 0);
```
wooj2: do we need to also add:
```
    XCTAssert(pushNotificationEvent.allAttributes != nil)
```

AWSPinpointTests/AWSPinpointNotificationManagerTests.m
```diff
@@ -0,0 +1,153 @@
...
+@interface AWSPinpointNotificationManager()
+- (void)addPinpointMetadataForEvent:(AWSPinpointEvent *) event
+                   withNotification:(NSDictionary *) userInfo;
+- (void)addGlobalPinpointMetadataForNotification:(NSDictionary *) userInfo;
+- (NSString*)buildEventTypeFromUserInfo:(NSDictionary *) userInfo
+                     withPushActionType:(AWSPinpointPushActionType) pushActionType;
+@end
+
+@interface AWSPinpointConfiguration()
+@property (nonnull, strong) NSUserDefaults *userDefaults;
```
wooj2: Did we mean to put `nonatomic` here, vs `nonnull` ?

AWSPinpointTests/AWSPinpointNotificationManagerTests.m
```diff
@@ -0,0 +1,153 @@
...
+// http://aws.amazon.com/apache2.0
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+//
+
+#import <XCTest/XCTest.h>
+#import "OCMock.h"
```
wooj2: lets remove `OCMock.h` if we are not using it.

AWSPinpoint/AWSPinpointAnalyticsClient.m
```diff
@@ -44,13 +44,13 @@ @interface AWSPinpointAnalyticsClient()
 @property (nonatomic, strong) NSMutableDictionary* eventTypeMetrics;
 @property (nonatomic, strong) NSMutableDictionary* globalAttributes;
 @property (nonatomic, strong) NSMutableDictionary* globalMetrics;
-@property (nonatomic, strong) NSDictionary* globalCampaignAttributes;
+@property (nonatomic, strong) NSDictionary* globalFeatureAttributes;
```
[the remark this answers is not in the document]
zhinang-amazon: This is because campaign and journey do not exist at the same time. And the globalCampaignAttributes was used to cache these attributes and attach to events later on. One thing we want to avoid is for one event to have both campaign and journey attributes. This is easier to maintain than caching both and determining which one to attach for following events.

AWSPinpoint/AWSPinpointNotificationManager.m
```diff
@@ -257,4 +288,26 @@ - (void)addApplicationStateAttributeToEvent:(AWSPinpointEvent *) event
...
+                     withPushActionType:(AWSPinpointPushActionType) pushActionType {
+    NSDictionary *pinpointData = userInfo[AWSDataKey][AWSPinpointKey];
+    NSString *eventTypePrefix;
+    if (pinpointData[AWSPinpointCampaignKey]) {
+        eventTypePrefix = AWSPinpointCampaignKey;
+    } else if (pinpointData[AWSPinpointJourneyKey]) {
+        eventTypePrefix = AWSPinpointJourneyKey;
+    } else {
+        eventTypePrefix = AWSPinpointUnknownKey;
+    }
```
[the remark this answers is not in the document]
zhinang-amazon: Good call. I will refactor this to smaller helper functions and build the event instead of just the eventType using these helper functions.

AWSPinpoint/AWSPinpointNotificationManager.m
```diff
@@ -207,30 +214,54 @@ - (void)recordMessageOpenedEventForNotification:(NSDictionary *) userInfo
...
-    
-    NSDictionary *campaign = userInfo[AWSDataKey][AWSPinpointKey][AWSPinpointCampaignKey];
-    AWSDDLogVerbose(@"Adding campaign global attributes: %@", campaign);
-    [self.context.analyticsClient setCampaignAttributes:campaign];
-    
-    for (NSString *key in [campaign allKeys]) {
-        [self.context.analyticsClient addGlobalAttribute:campaign[key] forKey:key];
+
+    NSDictionary *pinpointData = userInfo[AWSDataKey][AWSPinpointKey];
+    if (pinpointData[AWSPinpointCampaignKey]) {
```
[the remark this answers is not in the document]
zhinang-amazon: 1. It's not possible, one notification cannot contain attributes for both.
2. I was hesitant to do this because of exactly what you said. But it does make those two method a lot cleaner. I will make this change.

AWSPinpoint/AWSPinpointNotificationManager.m
```diff
@@ -257,4 +288,26 @@ - (void)addApplicationStateAttributeToEvent:(AWSPinpointEvent *) event
...
+- (NSString*)buildEventTypeFromUserInfo:(NSDictionary *) userInfo
+                     withPushActionType:(AWSPinpointPushActionType) pushActionType {
+    NSDictionary *pinpointData = userInfo[AWSDataKey][AWSPinpointKey];
+    NSString *eventTypePrefix;
+    if (pinpointData[AWSPinpointCampaignKey]) {
+        eventTypePrefix = AWSPinpointCampaignKey;
+    } else if (pinpointData[AWSPinpointJourneyKey]) {
+        eventTypePrefix = AWSPinpointJourneyKey;
+    } else {
+        eventTypePrefix = AWSPinpointUnknownKey;
```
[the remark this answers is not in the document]
zhinang-amazon: Will fix.

AWSPinpoint/AWSPinpointNotificationManager.m
```diff
@@ -153,9 +155,9 @@ - (void)handleNotificationReceived:(UIApplication *) app
   shouldHandleNotificationDeepLink:(BOOL) shouldHandleNotificationDeepLink {
     UIApplicationState state = [app applicationState];
     
+    [self addGlobalPinpointMetadataForNotification:userInfo];
```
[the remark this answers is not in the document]
zhinang-amazon: I think this was a potential bug, correct me if I'm wrong. If a push notification with new campaign/journey attributes is received while active, the global attributes would not be updated without this change.

AWSPinpoint/AWSPinpointAnalyticsClient.m
```diff
@@ -393,16 +393,20 @@ -(void) removeGlobalMetricForKey:(NSString*) theKey
     }
 }
 
-- (void) setCampaignAttributes:(NSDictionary*) campaign {
-    _globalCampaignAttributes = campaign;
-    [self.eventRecorder updateSessionStartWithCampaignAttributes:campaign];
+- (void) setFeatureAttributes:(NSDictionary*) attributes {
+    // Remove previous global feature attributes from _globalAttributes
+    // This is to prevent _globalAttributes containing attributes from multiple features (campaign/journey)
+    [self removeAllGlobalFeatureAttributes];
```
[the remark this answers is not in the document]
zhinang-amazon: This is necessary because these attributes are added to both _globalAttributes and _globalCampaign attributes in `AWSPinpointNotificationManager::addGlobalPinpointMetadataForNotification`
It seems redundant to me but I'm sure it was there for a reason. I can add unit test

AWSPinpoint/AWSPinpointNotificationManager.m
```diff
@@ -164,14 +166,17 @@ - (void)handleNotificationReceived:(UIApplication *) app
...
-        [self addGlobalCampaignMetadataForNotification:userInfo];
+        NSString *eventType = [self buildEventTypeFromUserInfo:userInfo
+                                            withPushActionType:AWSPinpointPushActionTypeReceivedBackground];
         [self recordMessageReceivedEventForNotification:userInfo
-                                              eventType:AWSEventTypeReceivedBackground
+                                              eventType:eventType
                                    withApplicationState:state];
     } else {
         AWSDDLogVerbose(@"Received notification with app on foreground.");
+        NSString *eventType = [self buildEventTypeFromUserInfo:userInfo
```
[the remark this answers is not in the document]
zhinang-amazon: Good idea. Will extract a method to map to AWSPinpointPushActionType

AWSPinpoint/AWSPinpointNotificationManager.m
```diff
@@ -207,30 +214,54 @@ - (void)recordMessageOpenedEventForNotification:(NSDictionary *) userInfo
...
+        }
+    } else if (pinpointData[AWSPinpointJourneyKey]) {
+        NSDictionary *journey = pinpointData[AWSPinpointJourneyKey];
+        AWSDDLogVerbose(@"Adding journey attributes to event[%@]: %@", event.eventType, journey);
+
+        for (NSString *key in [journey allKeys]) {
+            [event addAttribute:journey[key] forKey:key];
+        }
+    } else {
+        AWSDDLogVerbose(@"No valid Pinpoint Push payload found");
```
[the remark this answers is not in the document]
zhinang-amazon: Will fix.

AWSPinpointTests/AWSPinpointEventRecorderTests.m
zhinang-amazon: Will Remove

AWSPinpointTests/AWSPinpointEventRecorderTests.m
zhinang-amazon: This is not necessarily needed since from the AWSPinpointEvent and AWSPinpointEventRecorder point of view they are just attributes and do not need to know what the key or value is, but I will add one for journey to show that these are the possible "feature" attributes that `updateSessionStartWithFeatureAttributes` can be invoked with.

AWSPinpointTests/AWSPinpointNotificationManagerTests.m
zhinang-amazon: Will add

AWSPinpointTests/AWSPinpointNotificationManagerTests.m
zhinang-amazon: Will remove

AWSPinpointTests/AWSPinpointNotificationManagerTests.m
zhinang-amazon: I can make it static. Did not figure out a way to make it const though

AWSPinpointTests/AWSPinpointNotificationManagerTests.m
zhinang-amazon: Will update

AWSPinpointTests/AWSPinpointNotificationManagerTests.m
zhinang-amazon: Will move to static

AWSPinpointTests/AWSPinpointNotificationManagerTests.m
zhinang-amazon: Will update.

AWSPinpointTests/AWSPinpointNotificationManagerTests.m
zhinang-amazon: Will move

AWSPinpointTests/AWSPinpointNotificationManagerTests.m
zhinang-amazon: This test will be updated

AWSPinpointTests/AWSPinpointNotificationManagerTests.m
zhinang-amazon: It's nonatomic in AWSPinpointConfiguration.m but nonnull in all the test files. I will change all the test files to use nonatomic.

AWSPinpointTests/AWSPinpointNotificationManagerTests.m
zhinang-amazon: Will remove

AWSPinpoint/AWSPinpointNotificationManager.m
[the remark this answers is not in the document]
wooj2: That's how I read the code, but I'm not sure if there was a reason for not logging.  Is this something you can confirm?  Maybe we can take a look to see what Android is doing?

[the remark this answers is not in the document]
zhinang-amazon: On Android adding global attributes is only done [here](https://github.com/aws-amplify/aws-sdk-android/blob/master/aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/NotificationClientBase.java#L838) and this is invoked [here](https://github.com/aws-amplify/aws-sdk-android/blob/master/aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/NotificationClientBase.java#L900) Looks like Android is only doing this for Open event type, not background or foreground. I will discuss this with @drochetti

AWSPinpoint/AWSPinpointAnalyticsClient.m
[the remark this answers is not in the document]
wooj2: The main reason I don't like this is because by calling `setFeatureAttributes ` we are causing a side effect of removeAllGlobalFeatureAttributes.  This doesn't feel right.

So, the main question comes back to is: What is the main use case of calling `setFeatureAttributes:` ?  It seems like the AWSPinpointNotificationManager is calling it from `handleNotificationReceived:withNotification:shouldHandleNotificationDeepLink` and
`interceptDidFinishLaunchingWithOptions`, and by adding a call to `removeAllGlobalFeatureAttributes` it seems that the concern is that you can get one notification with journey information, and another notification with campaign information?   And if we get these two payloads in succession, it's possible for keys from the journey to "bleed" into the one with campaign information?

Can you help me better understand this so we can come up with a better solution?

AWSPinpoint/AWSPinpointNotificationManager.m
```diff
@@ -257,4 +260,73 @@ - (void)addApplicationStateAttributeToEvent:(AWSPinpointEvent *) event
...
+- (NSString*)getEventTypeSuffixFromPushActionType:(AWSPinpointPushActionType) pushActionType {
+    switch (pushActionType) {
+        case AWSPinpointPushActionTypeOpened:
+            return AWSEventTypeOpened;
+        case AWSPinpointPushActionTypeReceivedForeground:
+            return AWSEventTypeReceivedForeground;
+        case AWSPinpointPushActionTypeReceivedBackground:
+            return AWSEventTypeReceivedBackground;
+    }
+}
```
wooj2, line 299: Generally i favor favor reducing the number of exit points in a function, so i prefer something liek:
```
    NSString * eventType;
    switch (pushActionType) {
        case AWSPinpointPushActionTypeOpened:
            eventType = AWSEventTypeOpened;
            break;
        case AWSPinpointPushActionTypeReceivedForeground:
            eventType = AWSEventTypeReceivedForeground;
            break;
        case AWSPinpointPushActionTypeReceivedBackground:
            eventType = AWSEventTypeReceivedBackground;
            break;
    }
    return eventType;
```

AWSPinpoint/AWSPinpointNotificationManager.m
```diff
@@ -153,84 +153,87 @@ - (void)handleNotificationReceived:(UIApplication *) app
...
-    }
-    
-    NSDictionary *campaign = userInfo[AWSDataKey][AWSPinpointKey][AWSPinpointCampaignKey];
-    AWSDDLogVerbose(@"Adding campaign global attributes: %@", campaign);
-    [self.context.analyticsClient setCampaignAttributes:campaign];
-    
-    for (NSString *key in [campaign allKeys]) {
-        [self.context.analyticsClient addGlobalAttribute:campaign[key] forKey:key];
+- (void)addGlobalPinpointMetadata:(NSDictionary *) metadata {
+    [self.context.analyticsClient setFeatureAttributes:metadata];
```
wooj2, line 233: metadata can now be null, but we were previously checking if it was valid?

AWSPinpoint/AWSPinpointAnalyticsClient.m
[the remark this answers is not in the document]
wooj2: Is there a better word than "Feature"?  If i were to look at this code 6 months from now, and try to figure out what a Feature in the context of Pinpoint, I would probably be confused. I'm not fully sold on using this instance variable to serve two purposes but if we do go down this route, i'd probably name this something like globalCampaignOrJourneyAttributes

AWSPinpoint/AWSPinpointNotificationManager.m
zhinang-amazon: Will add the check

AWSPinpoint/AWSPinpointNotificationManager.m
zhinang-amazon: Will update this and `pushActionTypeOfApplicationState`

AWSPinpoint/AWSPinpointNotificationManager.m
```diff
@@ -153,84 +153,87 @@ - (void)handleNotificationReceived:(UIApplication *) app
...
+            //NSString *eventType = [self buildEventTypeFromUserInfo:userInfo
+            //                                    withPushActionType:AWSPinpointPushActionTypeReceivedBackground];
+            [self recordMessageReceivedEventForNotification:userInfo
+                                         withPushActionType:pushActionType];
+            break;
+        }
+        case AWSPinpointPushActionTypeReceivedForeground: {
+            AWSDDLogVerbose(@"Received notification with app on foreground.");
+            //NSString *eventType = [self buildEventTypeFromUserInfo:userInfo
+            //                                    withPushActionType:AWSPinpointPushActionTypeReceivedForeground];
```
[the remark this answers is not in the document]
zhinang-amazon: Well remove

AWSPinpoint/AWSPinpointNotificationManager.m
```diff
@@ -153,84 +153,87 @@ - (void)handleNotificationReceived:(UIApplication *) app
...
+                                           withIdentifier:nil];
+            if (shouldHandleNotificationDeepLink) {
+                [self handleNotificationDeepLinkForNotification:userInfo];
+            }
+            break;
+        }
+        case AWSPinpointPushActionTypeReceivedBackground: {
+            AWSDDLogVerbose(@"Received notification with app on background.");
+            //NSString *eventType = [self buildEventTypeFromUserInfo:userInfo
+            //                                    withPushActionType:AWSPinpointPushActionTypeReceivedBackground];
```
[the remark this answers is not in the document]
zhinang-amazon: Will Remove

AWSPinpoint/AWSPinpointAnalyticsClient.m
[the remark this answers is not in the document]
zhinang-amazon: Discussed offline with @wooj2 and @drochetti and agreed upon using this variable. 
Regarding naming, I'm proposing `globalEventSourceAttributes` instead of `globalFeatureAttributes`, since these attributes come from either Campaign events or Journey events.
Regarding [this comment](https://github.com/aws-amplify/aws-sdk-ios/pull/2387#discussion_r401825522), I will expose and explicitly call the `AWSPinpointAnalyticsClient::removeAllGlobalEventSourceAttributes` rather than calling it in `AWSPinpointAnalyticsClient::setEventSourceAttributes`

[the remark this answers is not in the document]
zhinang-amazon: A different proposed name is `globalServerSideAttributes`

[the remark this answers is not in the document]
drochetti: I like `eventSource` as a generic name for both campaign and journey :)

AWSPinpoint/AWSPinpointNotificationManager.m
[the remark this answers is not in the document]
zhinang-amazon: Got a decision from PM. We do not want to add global attributes if app is already in foreground. For received in the background, we want to ideally only add and invalidate after a certain period of time. Since this will require more work and design, this will be in a separate feature request. For now Android will change to match existing iOS behavior. I will revert this change to existing behavior.

AWSPinpoint/AWSPinpointAnalyticsClient.m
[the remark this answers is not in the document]
zhinang-amazon: `globalEventSourceAttributes ` it is.

AWSPinpoint/AWSPinpointNotificationManager.m
```diff
@@ -91,16 +93,15 @@ - (BOOL)interceptDidFinishLaunchingWithOptions:(nullable NSDictionary *)launchOp
...
-        
-        [self addGlobalCampaignMetadataForNotification:notificationPayload];
-        
+        NSDictionary *metadata = [self getMetadataFromUserInfo:notificationPayload];
+        [self addGlobalPinpointMetadata:metadata];
+
         // Application launch because of notification
         [self recordMessageOpenedEventForNotification:notificationPayload
-                                       withIdentifier:nil
-                                 withApplicationState:[[UIApplication sharedApplication] applicationState]];
```
wooj2, line 103: This function was previously calling:
```
   [self addApplicationStateAttributeToEvent:pushNotificationEvent	
                         withApplicationState:state];
```

Doesn't seem that we are doing this anymore?

To be honest, I'm having a lot of trouble trying to figure out of this code still works...  There were quite a bit of changes, I'm not familiar with the code in the first place, hard to know if we caused any regression.

Do you have a way of sending out smaller CRs so that it is easier to see a progression of changes (for this file only -- the other files, it's obvious what is being changed)?

AWSPinpoint/AWSPinpointNotificationManager.m
```diff
@@ -152,85 +153,94 @@ - (void)handleNotificationReceived:(UIApplication *) app
...
+            AWSDDLogVerbose(@"Received notification with app on background.");
+
+            [self addGlobalPinpointMetadata:metadata];
+            [self recordMessageReceivedEventForNotification:userInfo
+                                         withPushActionType:pushActionType];
+            break;
+        }
+        case AWSPinpointPushActionTypeReceivedForeground: {
+            AWSDDLogVerbose(@"Received notification with app on foreground.");
+
```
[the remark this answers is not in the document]
zhinang-amazon: Will add

AWSPinpoint/AWSPinpointNotificationManager.m
```diff
@@ -152,85 +153,94 @@ - (void)handleNotificationReceived:(UIApplication *) app
...
+    AWSPinpointEvent *pushNotificationEvent = [self buildEventFromUserInfo:userInfo
+                                                        withPushActionType:pushActionType];
+    if (!pushNotificationEvent) {
+        AWSDDLogError(@"Not valid Pinpoint push notification");
+        return;
+    }
+
+    NSDictionary *metadata = [self getMetadataFromUserInfo:userInfo];
+    [self addPinpointMetadataForEvent:pushNotificationEvent
+                     withMetadata:metadata];
```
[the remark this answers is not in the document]
zhinang-amazon: Will fix

AWSPinpoint/AWSPinpointNotificationManager.m
```diff
@@ -152,85 +153,94 @@ - (void)handleNotificationReceived:(UIApplication *) app
...
-        return;
-    }
-    
-    NSDictionary *campaign = userInfo[AWSDataKey][AWSPinpointKey][AWSPinpointCampaignKey];
-    AWSDDLogVerbose(@"Adding campaign global attributes: %@", campaign);
-    [self.context.analyticsClient setCampaignAttributes:campaign];
-    
-    for (NSString *key in [campaign allKeys]) {
-        [self.context.analyticsClient addGlobalAttribute:campaign[key] forKey:key];
+- (void)addGlobalPinpointMetadata:(NSDictionary *) metadata {
```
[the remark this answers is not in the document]
zhinang-amazon: Will update

AWSPinpoint/AWSPinpointNotificationManager.m
zhinang-amazon: This attribute is added to all push notification events, so I moved it to `buildEventFromUserInfo:withPushActionType` Let's discuss what's the best way to show the progression

royjit:
This PR was split into another one [here](https://github.com/aws-amplify/aws-sdk-ios/pull/2407). Closing this.

PR #2410, opened 2020-04-04
Add customDemographic field for pinpoint to override default demographic 

wooj2:
Prior to moving forward with this PR, I want to verify that this will resolve the customer's request.

In this prototype, the customer should be able to do something like:
```
//Setup custom fields that will be used in subsequent events that are sent to pinpoint:
let customDemographic: AWSPinpointEndpointProfileDemographic! = AWSPinpointEndpointProfileDemographic.default()
customDemographic.appVersion = "1.0.1.2.3.4.5.6"
customDemographic.platform = "MyCustomPlatform"
customDemographic.model = "MyCustomDeviceModel"

//No longer going with static variable named customDemographic
//AWSPinpointEndpointProfile.setCustomDemographic(customDemographic)

//Then initialize pinpoint as normal, perhaps something like this:
let pinpointConfiguration = AWSPinpointConfiguration.defaultPinpointConfiguration(launchOptions: launchOptions)
let pinpoint = AWSPinpoint(configuration: pinpointConfiguration)
//Overrides default demographic:
pinpoint.targetingClient.currentEndpointProfile().customDemographic = customDemographic

//Create some events:
analyticsClient.createEvent(withEventType: "UserPushedButton")

//Submit some events:
analyticsClient.submitEvents(completionBlock: {...})
```

And as  result, we'd expect to see the request to pinpoint being sent as:
```
{
...........
   "Demographic" : 
 {"Locale":"en_US",
"Timezone":"America\/Los_Angeles",
"PlatformVersion":"13.3",
"AppVersion":"1.0.1.2.3.4.5.6",
"Model":"MyCustomDeviceModel",
"Make":"apple",
"Platform":"MyCustomPlatform"},
........
}
```

### DIFF
--- a/AWSPinpoint/AWSPinpointEndpointProfile.m
+++ b/AWSPinpoint/AWSPinpointEndpointProfile.m
@@ -109,7 +109,7 @@ NSString *DEBUG_CHANNEL_TYPE = @"APNS_SANDBOX";
         
         //this updates demograhpic information.
         _location = [AWSPinpointEndpointProfileLocation new];
-        _demographic = [AWSPinpointEndpointProfileDemographic defaultAWSPinpointEndpointProfileDemographic];
+        _demographic = self.customDemographic ?: [AWSPinpointEndpointProfileDemographic defaultAWSPinpointEndpointProfileDemographic];
         _effectiveDate = [AWSPinpointDateUtils utcTimeMillisNow];
     }
 }
--- a/AWSPinpoint/AWSPinpointEndpointProfile.h
+++ b/AWSPinpoint/AWSPinpointEndpointProfile.h
@@ -76,6 +76,13 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, readwrite) AWSPinpointEndpointProfileDemographic *_Nullable demographic;
 
 /**
+ The AWSPinpointEndpointProfileDemographic facet that, if it is set, will overwrite the default demographic
+ for subsequent requests to Pinpoint
+ @returns demographic
+ */
+@property (nonatomic, strong) AWSPinpointEndpointProfileDemographic *_Nullable customDemographic;
+
+/**
  The AWSPinpointEndpointProfileUser facet of the endpoint profile
  @returns user
  */
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
 # AWS Mobile SDK for iOS CHANGELOG
+## Unreleased Changes
+- Added support for adding custom demographic to pinpoint [PR: #2410](https://github.com/aws-amplify/aws-sdk-ios/pull/2410)
 
 ## 2.13.2
 


wooj2:
Custom demographics seem to show up in the pinpoint aws console
Screen Shot 2020-04-03 at 10 38 09 PM

AWSPinpoint/AWSPinpointEndpointProfile.m
```diff
@@ -25,6 +25,7 @@
 static int const MAX_ENDPOINT_ATTRIBUTE_METRIC_KEY_LENGTH = 50;
 static int const MAX_ENDPOINT_ATTRIBUTE_VALUE_LENGTH = 100;
 static int const MAX_ENDPOINT_ATTRIBUTE_VALUES = 50;
+static AWSPinpointEndpointProfileDemographic *customDemographic;
```
palpatim, line 28: I'm not thrilled about using a global to store this, and the flow feels weird to me--in order to get this custom behavior, I have to call `setCustomDemographic` before initialization, and if I call it afterwards, it does not change the stored value. It breaks normal OO get/set behavior.

What about an overload of `-[updateEndpointProfileWithContext:]` like `-[updateEndpointProfileWithContext:customDemographic]` that takes a non-nullable demographic and immediately uses it in place of the default? Alternately, what is it about the current code that makes the already-existing setter for `demographic` not work as expected? 

Out of curiosity, how do Android & JS implement this feature? Also, since this PR is primarily to achieve feature parity, is the naming (`set/getCustomDemographic`) consistent with those platforms?

drochetti: My feedback is pretty much aligned with Tim's.

As for the Android/JS implementation, I believe the expectation from the customer in this case and the way those two platforms approach is, is that the API is transparent whether the default or the custom attributes are being sent. Therefore, from the customer perspective there's no new API, the custom values are set directly into the existing endpoint instance and if they are set, the Pinpoint client uses them, otherwise if falls back to the defaults.

wooj2: Yea, to be honest, I was not thrilled about it either, but the main interface to instantiating Pinpoint was this:
```
        let pinpointConfiguration = AWSPinpointConfiguration.defaultPinpointConfiguration(launchOptions: launchOptions)
        let pinpoint = AWSPinpoint(configuration: pinpointConfiguration)
```

If we try to do something like:
```
let pinpointConfiguration = AWSPinpointConfiguration.defaultPinpointConfiguration(launchOptions: launchOptions)
let pinpoint = AWSPinpoint(configuration: pinpointConfiguration)
pinpoint.targetingClient.currentEndpointProfile().demographic = customDemographic
```
I noticed that it is possible that it is possible that we can send out an event without the customDemographic fields.  This leads me to believe that by creating an overload of `updateEndpointProfileWithContext:customDemographic:` in the targeting client will not work.  That being said, I'll give it try and triple check.

Also, I don't understand what you mean by `and if I call it afterwards, it does not change the stored value.`.  I believe that if you do call it at a later time, it will change the value being sent out.

Android & JS -- i should have dove deeper.. I didn't in this case.  i'll get back to you.

wooj2: Spoke to @palpatim offline about this.  Moving away from the initially proposed static variable.  Concerns about sending out the default demographic first and then sending out a custom one should not be of a concern since the backend should handle updating an endpoint if the custom demographics are updated for that endpoint.  I verified in the pinpoint console and only customDemographics show up. :)

PR #2437, opened 2020-04-16
amazonrelationaldatabaseservice Update models to latest

awsmobilesdk:
amazonrelationaldatabaseservice Update models to latest


palpatim:
The files should be named with the pattern `AWSRDS` as opposed to `AWSrds`. iOS pattern is to let initialisms be all-caps rather than Pascal-case

Issue #861, opened 2018-03-29
Unable to successfully update endpoint.


sauravexodus:
### To help us solve your problem better, please answer the following list of questions.

* What service are you using?
AWS Pinpoint, AWS Mobile Hub

* In what version of SDK are you facing the problem?
2.1.2

* Is the issue limited to Simulators / Actual Devices?
No

* Can your problem be resolved if you bump to a higher version of SDK?
 No

* Is this problem related to specific iOS version?
No

* How are you consuming the SDK? CocoaPods / Carthage / Prebuilt frameworks?
Cocoapods

* Can you give us steps to reproduce with a minimal, complete, and verifiable example? Please include any specific network conditions that might be required to reproduce the problem.

- Add `configuration.json` to your project.
- Follow steps to initialize analytics from [here](https://docs.aws.amazon.com/aws-mobile/latest/developerguide/add-aws-mobile-analytics.html)
- See the following output when you try to initialize the pinpoint instance:

```xclog
Unable to successfully update endpoint. Error Message:Error Domain=com.amazonaws.AWSPinpointTargetingErrorDomain Code=1 "BadRequestException" UserInfo={RequestID=a0145ff1-3320-11e8-b003-6b800874b928, responseStatusCode=400, responseHeaders=<CFBasicHash 0x1c686b780 [0x1b3344310]>{type = immutable dict, count = 12,
entries =>
	0 : Content-Type = <CFString 0x1c0a44ce0 [0x1b3344310]>{contents = "application/json"}
	1 : Access-Control-Allow-Origin = *
	2 : X-Amz-Cf-Id = <CFString 0x1c0486400 [0x1b3344310]>{contents = "JkzQckhTMh87hfT5hC9ygmDhbvwOsoAeLEbRfxdIuwt2hmOrW5Sd9g=="}
	3 : X-Amzn-Trace-Id = <CFString 0x1c0486590 [0x1b3344310]>{contents = "sampled=0;root=1-5abc9201-85f2b0894a37c36c06fea4c3"}
	4 : x-amzn-ErrorType = <CFString 0x1c0a59ec0 [0x1b3344310]>{contents = "BadRequestException"}
	5 : Via = <CFString 0x1c02b3c20 [0x1b3344310]>{contents = "1.1 415c421af891d7c3f4ade60aba542014.cloudfront.net (CloudFront)"}
[...]

PR #2455, opened 2020-04-17
Fix Pinpoint Missing Address issue by setting OptOut based on device token

lawmicha:
*Issue #, if available:*

https://github.com/aws-amplify/aws-sdk-ios/issues/861

*Description of changes:*

Developers can use AWS Pinpoint SDK to update an user's endpoint (segment the user) and record events for their particular app use case. An endpoint has a field 'optOut' which determines whether the user has opted out in receiving messages from the app developer. https://docs.aws.amazon.com/pinpoint/latest/apireference/apps-application-id-endpoints.html 

Enabling push notifications on the app requires the app developer to request notification permissions from the user. If the user allows the app to receive push notifications, a device token is assigned and provided back to the app. If the developer decides to use pinpoint as a way to create campaigns/journeys to target their users and send messages via push notification, then that device token also needs to be updated on the endpoint and optOut needs to be set to NONE.

So the app gets the device token back and then calls `interceptDidRegisterForRemoteNotifications` to register the device token with the endpoint:
```
func application(_ application: UIApplication, didRegisterForRemoteNotificationsWithDeviceToken deviceToken: Data) {
    // Register the device token with Pinpoint as the endpoint for this user
   pinpoint!.notificationManager.interceptDidRegisterForRemoteNotifications(withDeviceToken: deviceToken)
}
```

When this is done, `optOut` will automatically be set to NONE using the logic: 
1. Use configuration field `applicationLevelOptOut`
2. If not set, check `UIApplication.isRegisteredForRemoteNotifications`

So the issue is for **_apps which have enabled push notifications, is using the pinpoint SDK, but is not registering the token with the endpoint_**. Pinpoint SDK will set the optOut to NONE on the endpoint and the response will be 400 Bad request, "Missing Address" error. Causing failiures to update the endpoint.

The issue does not exist on Android https://aws-amplify.github.io/docs/sdk/android/push-notifications
Developers register the token:
```
public void onNewToken(String token) {
    ...
    Log.d(TAG, "Registering push notifications token: " + token);  
    MainActivity.getPinpointManager(getApplicationContext())
        .getNotificationClient()
        .registerDeviceToken(token);
}
```
Then optOut is set based on if the app notification is enabled and if deviceToken exists.

https://github.com/aws-amplify/aws-sdk-android/blob/master/aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/endpointProfile/EndpointProfile.java#L202

**Description of changes**
The change here is to update the logic to check is the app has enabled push notifications and if the device token exists to determine the value of `optOut`, same as Android, so that app developers can enable push notifications in their app without failing to update the endpoint profiles.


AWSPinpoint/AWSPinpointEndpointProfile.m
```diff
@@ -122,7 +122,9 @@ - (BOOL) isApplicationLevelOptOut:(AWSPinpointContext *) context {
 }
 
 - (void) setEndpointOptOut:(BOOL) applicationLevelOptOut {
-    BOOL isOptedOutForRemoteNotifications = ![AWSPinpointNotificationManager isNotificationEnabled];
+    BOOL isNotificationEnabled = [AWSPinpointNotificationManager isNotificationEnabled];
+    BOOL isOptedOutForRemoteNotifications = !(isNotificationEnabled && [_address length]);
+
```
wooj2, line 127: 1.  We can use self.address.length rather than _address.  Doing so will use the accessor which has the `nonatomic` and `readwrite` modifiers to the accessor.
2.  I believe your logic is correct, but it might be easier to read if we change it to the following:
```
    BOOL isUsingPinpointForNotifications = [AWSPinpointNotificationManager isNotificationEnabled] && self.address.length
    BOOL isOptedOutForNotifications = !isUsingPinpointForNotifications
```
I think this is easier because `isUsingPinpointForNotifications` describes a case where the user has NotificationsEnabled and the developer has actually called the `intercept` call and passed in the deviceid.

AWSPinpointTests/AWSPinpointTargetingClientTests.m
```diff
@@ -107,27 +110,15 @@ - (void)initializePinpointWithConfiguration:(AWSPinpointConfiguration *)configur
...
     id mockApplication = OCMClassMock([UIApplication class]);
 
     OCMStub([mockApplication sharedApplication]).andReturn(mockApplication);
 
-    [self setSystemLevelOptOut:mockApplication withOptOut:optOut];
-}
-
-- (void)setSystemLevelOptOut:(UIApplication *)application withOptOut:(BOOL)optOut {
-    UIUserNotificationType notificationType = optOut ? UIUserNotificationTypeNone : UIUserNotificationTypeAlert;
+    UIUserNotificationType notificationType = withRemoteNotifications ? UIUserNotificationTypeNone : UIUserNotificationTypeAlert;
```
wooj2, line 118: if `withRemoteNotifications` == true, then I would expect notificationType to equal `UIUserNotificationTypeAlert`? but that doesn't seem to be the case.  Maybe we can chat offline about this?

AWSPinpointTests/AWSPinpointTargetingClientTests.m
```diff
@@ -79,19 +80,21 @@ - (AWSPinpointConfiguration *)getDefaultAWSPinpointConfiguration {
...
 }
 
-- (void)setApplicationLevelOptOut:(BOOL)optOut {
-    [self initializePinpointWithConfiguration:[self getAWSPinpointConfigurationWithOptOut:optOut]];
+- (void)setDeviceTokenInUserDefaults {
+    const unsigned char currentTokenBytes[] = {
+        0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07,
+        0x08, 0x09, 0x0a, 0x0b, 0x0c, 0x0d, 0x0e, 0x0f
+    };
+    NSData *currentTokenData = [[NSData alloc] initWithBytes:currentTokenBytes length:16];
```
wooj2, line 95: If i think back to my C-programming days, i think this should be:
```
    NSData *currentTokenData = [[NSData alloc] initWithBytes:currentTokenBytes length:sizeof(currentTokenBytes)/sizeof(unsigned char)]
```

AWSPinpointTests/AWSPinpointTargetingClientTests.m
```diff
@@ -79,19 +80,21 @@ - (AWSPinpointConfiguration *)getDefaultAWSPinpointConfiguration {
...
+    [self.userDefaults removeObjectForKey:AWSDeviceTokenKey];
+    [self.userDefaults setObject:currentTokenData forKey:AWSDeviceTokenKey];
```
wooj2, line 97: we probably want to call synchronize, to make sure this gets flushed out to nsuserdefaults:
```
[self.userDefaults synchronize];
```

AWSPinpointTests/AWSPinpointTargetingClientTests.m
wooj2: oh, this seems to be copy and pasted across the entire class.. hmm.. leaving this as 16 is probably fine.

AWSPinpointTests/AWSPinpointTargetingClientTests.m
lawmicha: nice catch, i changed swapped the logic here from 
`OCMStub([application isRegisteredForRemoteNotifications]).andReturn(optOut == NO);`
to `OCMStub([mockApplication isRegisteredForRemoteNotifications]).andReturn(withRemoteNotifications);` but forgot to do it for `UIUserNotificationType notificationType = withRemoteNotifications ? UIUserNotificationTypeNone : UIUserNotificationTypeAlert;`

AWSPinpoint/AWSPinpointEndpointProfile.m
lawmicha: thanks, updating!

AWSPinpointTests/AWSPinpointTargetingClientTests.m
lawmicha: adding

AWSPinpointTests/AWSPinpointTargetingClientTests.m
```diff
@@ -79,19 +80,22 @@ - (AWSPinpointConfiguration *)getDefaultAWSPinpointConfiguration {
...
```
wooj2, line 97: The unit test should have its own constant for AWSDeviceTokenKey, rather than relying on the one in production code.

AWSPinpointTests/AWSPinpointTargetingClientTests.m
```diff
@@ -249,49 +241,56 @@ - (void)testNewDeviceTokenUpdateWithStandardUserDefaults {
...
+    [self initializePinpointWithConfiguration:[self getAWSPinpointConfigurationWithOptOut:YES] forceCreate:YES];
+    [self initializeMockApplicationWithRemoteNotifications:YES];
+    [self setDeviceTokenInUserDefaults];
+
+    AWSPinpointEndpointProfile *profile = [self.pinpoint.targetingClient currentEndpointProfile];
     XCTAssertTrue([profile.optOut isEqualToString:@"ALL"]);
 }
 
-- (void) testCurrentProfileWithSystemOptInAndApplicationOptOutBackgroundThread {
+- (void) testCurrentProfileWithNotificationsEnabledBackgroundThread {
```
wooj2, line 284: should we have a similar test where we notification is disabled by not setting the deviceTokenInUserDefaults?

AWSPinpointTests/AWSPinpointTargetingClientTests.m
```diff
@@ -55,10 +55,11 @@ - (void)setUp {
     [[NSUserDefaults standardUserDefaults] removeSuiteNamed:@"AWSPinpointTargetingClientTests"];
     self.userDefaults = [[NSUserDefaults alloc] initWithSuiteName:@"AWSPinpointTargetingClientTests"];
 
-    [self initializeMockApplicationWithOptOut:YES];
+    [self initializeMockApplicationWithRemoteNotifications:NO];
     [self initializePinpointWithConfiguration:[self getDefaultAWSPinpointConfiguration] forceCreate:NO];
```
wooj2, line 59: We should remove these lines here, and put them in individual functions (so that we aren't reinitializing state)

lawmicha: that one should be covered by line 250, `testCurrentProfileReturnsOptOutAllWhenNotificationsEnabledAndDeviceTokenNotSet` and this test based on executing on background thread, i think, was meant mainly to test calling into the main thread with remote notifications enabled. but doesn't hurt add more test coverage!

AWSPinpointTests/AWSPinpointTargetingClientTests.m
lawmicha: ah lost my comment, this should be covered by `testCurrentProfileReturnsOptOutAllWhenNotificationsEnabledAndDeviceTokenNotSet`, but will add this for the execution on background thread testing

AWSPinpointTests/AWSPinpointTargetingClientTests.m
lawmicha: addiing

AWSPinpointTests/AWSPinpointTargetingClientTests.m
lawmicha: i agree, it's hard to know what's going on when the setUp portion of the code is setting up the arrange part of a test, especially the mock application with remote notifications set to false

AWSPinpointTests/AWSPinpointTargetingClientTests.m
```diff
@@ -55,10 +55,10 @@ - (void)setUp {
     [[NSUserDefaults standardUserDefaults] removeSuiteNamed:@"AWSPinpointTargetingClientTests"];
     self.userDefaults = [[NSUserDefaults alloc] initWithSuiteName:@"AWSPinpointTargetingClientTests"];
 
-    [self initializeMockApplicationWithOptOut:YES];
-    [self initializePinpointWithConfiguration:[self getDefaultAWSPinpointConfiguration] forceCreate:NO];
+    //[self initializePinpointWithConfiguration:[self getDefaultAWSPinpointConfiguration] forceCreate:NO];
```
wooj2, line 58: remove?

AWSPinpointTests/AWSPinpointTargetingClientTests.m
```diff
@@ -55,10 +55,10 @@ - (void)setUp {
...
     [self.userDefaults removeObjectForKey:@"AWSPinpointEndpointAttributesKey"];
     [self.userDefaults removeObjectForKey:@"AWSPinpointEndpointMetricsKey"];
```
wooj2, line 60: we should make these constants too, but not a big deal

AWSPinpointTests/AWSPinpointTargetingClientTests.m
lawmicha: woops, removing

AWSPinpointTests/AWSPinpointTargetingClientTests.m
lawmicha: nice catch, i'll do this

PR #2502, opened 2020-04-27
Phani/integtest updates

phani-srikar:
*Issue #, if available:*
Update Integration Test suite to use session based temporary credentials

*Description of changes:*
Refactor integration tests of first bunch of packages to use Session based temporary credentials.
These test suites now read from testconfiguration.json instead of credentials.json.

### DIFF
--- a/AWSAPIGatewayTests/AWSLambdaMicroserviceClient.m
+++ b/AWSAPIGatewayTests/AWSLambdaMicroserviceClient.m
@@ -64,7 +64,7 @@ static NSString *const AWSInfoClientKey = @"AWSLambdaMicroserviceClient";
 
 static AWSSynchronizedMutableDictionary *_serviceClients = nil;
 
-+ (instancetype)defaultClient {
++ (instancetype)defaultClient:(NSString *)endpointURL {
     AWSServiceConfiguration *serviceConfiguration = nil;
     AWSServiceInfo *serviceInfo = [[AWSInfo defaultAWSInfo] defaultServiceInfo:AWSInfoClientKey];
     if (serviceInfo) {
@@ -76,6 +76,14 @@ static AWSSynchronizedMutableDictionary *_serviceClients = nil;
         serviceConfiguration = [[AWSServiceConfiguration alloc] initWithRegion:AWSRegionUnknown
                                                            credentialsProvider:nil];
     }
+    
+    if ([endpointURL hasSuffix:@"/"]) {
+        endpointURL = [endpointURL substringToIndex:[endpointURL length] - 1];
+    }
+    serviceConfiguration.endpoint = [[AWSEndpoint alloc] initWithRegion:serviceConfiguration.regionType
+                                                                service:AWSServiceAPIGateway
+                                                                    URL:[NSURL URLWithString:endpointURL]];
+
 
     static AWSLambdaMicroserviceClient *_defaultClient = nil;
     static dispatch_once_t onceToken;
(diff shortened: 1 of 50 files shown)


AWSAPIGatewayTests/AWSLambdaMicroserviceClient.m
```diff
@@ -130,7 +130,7 @@ - (instancetype)initWithConfiguration:(AWSServiceConfiguration *)configuration {
     if (self = [super init]) {
         _configuration = [configuration copy];
 
-        NSString *URLString = @"https://mykov6r7rh.execute-api.us-east-1.amazonaws.com/prod";
+        NSString *URLString = @"https://yjzm4utpcl.execute-api.us-east-1.amazonaws.com/prod";
```
palpatim, line 133: We need this to be parameterized rather than hardcoded. I believe this file is generally code-generated and downloaded, but we can extend it slightly since we're using it as a test fixture.

AWSCoreTests/AWSTestUtility.h
```diff
@@ -22,6 +22,12 @@ FOUNDATION_EXPORT NSString *const AWSTestUtilityCognitoIdentityServiceKey;
 
 @interface AWSTestUtility : NSObject
 
++ (void)setupSessionCredentialsProvider;
++ (NSDictionary *)getIntegrationTestConfigurationFor:(NSString *) packageId;
```
palpatim, line 26: nit: ObjectiveC naming is a bit more verbose than Swift. Change to:

```objective-c
+ (NSDictionary *)getIntegrationTestConfigurationForPackageId:(NSString *) packageId;
```

AWSCoreTests/AWSTestUtility.m
```diff
@@ -65,6 +65,57 @@ + (void)initialize {
...
 }
 
++ (void) setupSessionCredentialsProvider {
+    
+    NSDictionary *testConfigurationJson = [self getTestConfigurationJSON];
+    AWSBasicSessionCredentialsProvider *credentialsProvider = [[AWSBasicSessionCredentialsProvider alloc]
+                                                               initWithAccessKey:testConfigurationJson[@"Credentials"][@"accessKey"]
+                                                                       secretKey:testConfigurationJson[@"Credentials"][@"secretKey"]
+                                                                    sessionToken:testConfigurationJson[@"Credentials"][@"sessionToken"]];
+    AWSServiceConfiguration *configuration = [[AWSServiceConfiguration alloc] initWithRegion: [testConfigurationJson[@"Credentials"][@"region"] aws_regionTypeValue]
```
palpatim, line 75: It seems weird that "region" would be in the "credentials" block of the configuration.

AWSCoreTests/AWSTestUtility.m
```diff
@@ -65,6 +65,57 @@ + (void)initialize {
     [super initialize];
 }
 
++ (void) setupSessionCredentialsProvider {
+    
+    NSDictionary *testConfigurationJson = [self getTestConfigurationJSON];
+    AWSBasicSessionCredentialsProvider *credentialsProvider = [[AWSBasicSessionCredentialsProvider alloc]
+                                                               initWithAccessKey:testConfigurationJson[@"Credentials"][@"accessKey"]
```
palpatim, line 72: As much as you're referencing `testConfigurationJson[@"Credentials"]` in this file, it would be nice to make a helper method for it

AWSAPIGatewayTests/AWSLambdaMicroserviceClient.m
phani-srikar: modified the default client to require endpoint url for initialization

AWSCoreTests/AWSTestUtility.h
phani-srikar: Done

AWSCoreTests/AWSTestUtility.m
phani-srikar: done

AWSCoreTests/AWSTestUtility.m
phani-srikar: Aiming to have a single region for all test resources. Handling multiple regions needs to be discussed.

phani-srikar: "region" is shifted to "common" block which also has a default cognito identity pool and related info to be used by tests requiring authorization using cognito idp (cognito should be supported in that region)

PR #2525, opened 2020-04-30
Release process refactoring & fixes

palpatim:
- ci: Add CHANGELOG bump during bump_sdk_version
- ci: Refactor bump_sdk_version
- ci: Regroup framework list into less granular tranches
- ci: Remove duplicate AWSAutoScaling entry from framework list
- docs: Add 'Unreleased' section to CHANGELOG
- fix: Change bump_sdk target branch to develop instead of test_develop
- style: Reformatted bump_sdk_version and functions

*Issue #, if available:*

*Description of changes:*
In addition to refactoring the `bump_sdk_version` script, the big changes that I’d like you to focus on are:
- My reorganization/regrouping of the framework list -- I grouped these into less granular groups to reduce the number of hardcoded waits in the cocoapods release. These groups are based on the dependencies noted in the framework list itself.
- My change of the bump_sdk_version branch from `test_develop` to `develop`. This stopped working with the 2.13.2 release, for reasons we can't quite identify. There appears to have been a change that caused the `test_develop` and `develop` branches to stay un-synchronized; and PRs were no longer being posted to the `develop` branch. My investigation didn't reveal any changes to either the CircleCI config or the support scripts that seemed suspicious, so all I can figure is that the change was an unintended side effect of some of the reordering of release steps to put the cocoapods & carthage releases at the end of the process. My evaluation of this change is that it's innocuous, but obviously we'll want to keep pretty close eye on the generated bump-version PR during the next release.

### DIFF
--- a/CircleciScripts/framework_list.py
+++ b/CircleciScripts/framework_list.py
@@ -15,7 +15,7 @@ grouped_frameworks = [
 
     [
         # Depends only on AWSCognitoIdentityProviderASF
-        'AWSCognitoAuth'
+        'AWSCognitoAuth',
 
         # Depends on AWSCore and AWSCognitoIdentityProviderASF
         'AWSCognitoIdentityProvider',
@@ -62,7 +62,7 @@ grouped_frameworks = [
 
     [
         # Depends only on AWSCognito service-api package
-        'AWSCognitoSync'
+        'AWSCognitoSync',
 
         # Depends on AWSCore and AWSAuthCore
         'AWSAuthUI',
@@ -79,7 +79,7 @@ grouped_frameworks = [
 
     [
         # Depends on most previous packages except auth
-        'AWSiOSSDKv2'
+        'AWSiOSSDKv2',
 
         # Depends on AWSAuthCore, AWSFacebookSignIn, AWSGoogleSignIn,
         # AWSUserPoolsSignIn and AWSAuthUI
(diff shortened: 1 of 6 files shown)


palpatim:
> this PR: [CocoaPods/cocoapods-trunk#147](https://github.com/CocoaPods/cocoapods-trunk/pull/147) has been merged in. shouldn't we look to refactor this to push the pods directly to the cocoapods repo instead of to the cdn and remove the grouping/wait times? we can use the `--synchronous` flag now.

It was only merged a few days ago, and I'm not sure when it will make it into a stable release. This PR feels like a reasonable interim step.

CircleciScripts/framework_list.py
```diff
@@ -3,24 +3,27 @@
...
 grouped_frameworks = [
     # No dependencies
     [
         'AWSCore',
         'AWSCognitoIdentityProviderASF',
     ],
 
-    # Depends only on AWSCognitoIdentityProviderASF
-    ['AWSCognitoAuth'],
+    [
```
kneekey23, line 16: so this regrouping works? any way to test this?

palpatim: I've been trying to think of a way to test without a full release, and all I can think of is trying to push to a local repo. However, that doesn't really test the timing and dependencies very well.

PR #2535, opened 2020-05-01
Use AWSTestResources instead of hardcoded testconfiguration.json file

palpatim:
- feat(test): Add AWSTestResources module to provide test configurations
- fix(core): Move Headers build phase before Compile Sources for all targets

*Issue #, if available:*

*Description of changes:*
This includes a stub script to generate the testconfig from command line, which I will flesh out in a future PR. This PR should be sufficient to allow the project to build, since it removes direct project dependencies on the `testconfiguration.json` file in favor of making `AWSCoreTests` depend on the new `AWSTestResources` package.

### DIFF
--- a/AWSTestResources/AWSTestConfiguration.m
+++ b/AWSTestResources/AWSTestConfiguration.m
@@ -1,0 +1,55 @@
+//
+// Copyright 2010-2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License").
+// You may not use this file except in compliance with the License.
+// A copy of the License is located at
+//
+// http://aws.amazon.com/apache2.0
+//
+// or in the "license" file accompanying this file. This file is distributed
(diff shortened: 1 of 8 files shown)


AWSTestResources/AWSTestConfiguration.m
```diff
@@ -0,0 +1,55 @@
...
+                @"common": @{
+                        @"region": @"us-east-1"
+                }
+        }
+    };
+    
+    NSBundle *bundle = [NSBundle bundleForClass:[self class]];
+    NSString *filePath = [bundle pathForResource:@"testconfiguration" ofType:@"json"];
+    if (!filePath) {
+        return emptyDictionary;
```
phani-srikar, line 37: returning empty dictionary vs throwing error that "file does not exist in right format".
Here and throughout the method.
Need more context.

AWSTestResources/AWSTestResources.h
```diff
@@ -0,0 +1,26 @@
+//
```
phani-srikar, line 1: The implementation is still TODO for this right?

palpatim: Correct

PR #2545, opened 2020-05-06
Link testconfiguration.json file during build

palpatim:
- Script to generate test configuration in user's HOME directory
- Update Xcode image to 11.4.1
- Minor CircleCI config file cleanup
- Remove testconfiguration.json references from AuthSDK tests
- Update some mobile client test keys to use testconfiguration.json

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/Scripts/generate-test-config.sh
+++ b/Scripts/generate-test-config.sh
@@ -2,20 +2,21 @@
 
 # vim: set ts=2 sw=2 sts=2 et:
 
+
 ##################################################
 # SCRIPT METADATA
 
-SCRIPT_NAME=$(basename $0)
-SUMMARY="Generates a testconfiguration.json file for use in Amplify and AWS mobile SDK integration tests"
-SYNOPSIS="${SCRIPT_NAME} <android|ios> [options]"
+readonly SCRIPT_NAME=$(basename "$0")
+readonly SUMMARY="Generates a testconfiguration.json file for use in Amplify and AWS mobile SDK integration tests"
+readonly SYNOPSIS="${SCRIPT_NAME} <android|ios> [options]"
 
 
 ##################################################
 # DEFAULTS & DECLARATIONS
 ##################################################
-LOG_LEVEL_INFO=0
-LOG_LEVEL_DEBUG=1
-LOG_LEVEL_TRACE=2
+readonly LOG_LEVEL_INFO=0
+readonly LOG_LEVEL_DEBUG=1
+readonly LOG_LEVEL_TRACE=2
 LOG_LEVEL=0
 
 
@@ -27,7 +28,7 @@ LOG_LEVEL=0
 
 function log {
   msg="$(date) $*"
-  echo $msg >&2
+  echo "$msg" >&2
 }
 
 function log_debug {
@@ -61,7 +62,7 @@ function log_trace {
 }
 
 function die {
-  log_error $*
+  log_error "$@"
   exit 1
 }
 
@@ -79,8 +80,8 @@ ${usage}
 SYNOPSIS
     ${SYNOPSIS}
 
-$SCRIPT_NAME generates a test configuration file for AWS SDK and
-Amplify integration tests.
+DESCRIPTION
+    ${SUMMARY}
 
 REQUIREMENTS
     $SCRIPT_NAME requires the following utilities to be installed
@@ -131,7 +132,7 @@ assume_role=""
 platform=""
 region=${AWS_DEFAULT_REGION}
 
-while getopts "ap:qvh" optchar ; do
+while getopts "ap:qr:vh" optchar ; do
   case "$optchar" in
     a)
       assume_role=1
@@ -147,7 +148,7 @@ while getopts "ap:qvh" optchar ; do
       LOG_LEVEL=-1
       ;;
     v)
-      LOG_LEVEL+=1
+      ((LOG_LEVEL++))
       ;;
     h)
       print_help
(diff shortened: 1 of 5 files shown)


Scripts/generate-test-config.sh
```diff
@@ -1,14 +1,269 @@
...
 
-DESTINATION=$1
+# vim: set ts=2 sw=2 sts=2 et:
 
-if [[ -z $DESTINATION ]] ; then
-  echo "Destination not specified" >&2
+##################################################
+# SCRIPT METADATA
+
+SCRIPT_NAME=$(basename $0)
```
jamesonwilliams, line 8: Bash supports the `readonly` keyword, which is good for global constant values, like these:
```
readonly SCRIPT_NAME="$(basename $0)"
readonly SYNOPSIS="$SCRIPT_NAME"
```
etc.

As an alternative, instead of forward-declaring these global vars, you could mimic the constructs of a general purpose programming language:

```
main() {
    # Bash (not shell) supports both final-assignment and function-scoped variables.
    readonly local program_name="$(dirname $0)"
    # ...
}

main $@
```

Scripts/generate-test-config.sh
```diff
@@ -1,14 +1,269 @@
...
+function log_debug {
+  if [[ $LOG_LEVEL -lt LOG_LEVEL_DEBUG ]] ; then
+    return
+  fi
+
+  log "[D]:: $*"
+}
+
+# Always logged
+function log_error {
```
jamesonwilliams, line 42: To the same benefit of modularization in other programming languages, you could extract your logging stuff to a `logging.sh`, and then source it into this script with `source logging.sh` or `. logging.sh`.

Scripts/generate-test-config.sh
```diff
@@ -1,14 +1,269 @@
...
+
+# Usage & help
+
+function print_usage {
+  echo "usage: ${SYNOPSIS}"
+}
+
+function print_help {
+  usage=$(print_usage)
+  cat <<EOF
```
jamesonwilliams, line 76: It won't be much of an improvement, but the `<<-EOF` will at least allow you to pad all of this content into the document with a `\t`, so that the function block is more clear.

Scripts/generate-test-config.sh
```diff
@@ -1,14 +1,269 @@
...
+      ;;
+  esac
+done
+
+
+##################################################
+# VALIDATIONS AND ASSIGNMENTS
+
+cmd_quiet_flag="--quiet"
+if [[ $LOG_LEVEL -gt 1 ]] ; then
```
jamesonwilliams, line 168: It hardly matters, but as a factoid, `[[` is an external program, while `[` is a Bash built-in. So it is more efficient to use `[`, when you can. `[` does support `-z`, `-gt`, etc., as you use in this block.

Scripts/generate-test-config.sh
```diff
@@ -1,14 +1,269 @@
...
+
+if [[ -z $region ]] ; then
+  die "'region' not specified"
+fi
+
+python_cmd=$(which python3)
+if [[ -z $python_cmd ]] ; then
+  die "Cannot find 'python3' in PATH"
+fi
+
```
jamesonwilliams, line 184: My short-hand for these would be like:
```
[ -x "$(which python3)" ] || die "No exeuctable 'python3' in PATH"
```

But, since this is a check you repeat often, you could iterate over a list of dependencies, perhaps:
```
readonly dependencies=(
    python3
    aws
    jq
)
for cmd in $dependencies; do
    [ -x "$(which $cmd)" ] || die "No executable '$cmd' in PATH."
done
```

the `|| die` syntax is still useful for the other non-`-x` checks, though.

Scripts/generate-test-config.sh
```diff
@@ -1,14 +1,269 @@
...
+if [[ $assume_role ]] ; then
+  aws_cli_cmd=$(which aws)
+  if [[ -z $aws_cli_cmd ]] ; then
+    die "Cannot find 'aws' in PATH"
+  fi
+
+  jq_cmd=$(which jq)
```
jamesonwilliams, line 191: Let's say that a naked call to `jq` will resolve to `/on/your/path/jq`. Then `$(which jq)` would _also_ result in a call to `/on/your/path/jq`. So, you may not gain a lot from storing the path, above just doing `jq`, later.

If you want to be careful about the script path, you could clobber `PATH` at the beginning of your script, and then reconstruct it with "safe" directories, where you believe you will find well-known and un-compromised utility software (`"/usr/bin:/bin:/usr/sbin:/sbin"`, for example.)

Debian stable used to be even more conservative than that, and not include `/usr/bin` in `root`'s default `$PATH`, I believe.

Scripts/generate-test-config.sh
```diff
@@ -1,14 +1,269 @@
...
+  if [[ -z $jq_cmd ]] ; then
+    die "Cannot find 'jq' in PATH"
+  fi
+fi
+
+
+##################################################
+# MAIN
```
jamesonwilliams, line 199: A lot of the procedural stuff you have below here could be encapsulated into functions

Scripts/generate-test-config.sh
```diff
@@ -1,14 +1,269 @@
...
+
+log_debug "Installing python requirements"
+pip3 install --upgrade pip $cmd_quiet_flag
+pip3 install -r requirements.txt $cmd_quiet_flag
+
+# Assume test execution role and export those values so the builder script can pick them up from the environment
+if [[ $assume_role ]] ; then
+  log_debug "Assuming test execution IAM role"
+  circleci_execution_role_arn=$(aws ssm get-parameter --name '/mobile-sdk/ios/common/circleci_execution_role' | jq -r .Parameter.Value)
+  assume_role_creds=$(aws sts assume-role --duration-seconds 3600 --role-arn "${circleci_execution_role_arn}" --role-session-name "IntegTest-$(date +%Y%m%d%H%M%S)")
```
jamesonwilliams, line 251: Does this particular `date` command work accross `Darwin` and `Linux` `uname`s?

Scripts/generate-test-config.sh
palpatim: Thanks for all the feedback on this--lots of new bash learnings for me!

Scripts/generate-test-config.sh
palpatim: I considered but rejected that, in favor of keeping the script and its functional dependencies self-contained. (I also don't know of a good strategy for dependency management for bash--we'll usually be invoking these scripts from a directory other than the one the script itself is contained in, which makes discovery of the "library" files tricky. I guess I could do something like:

```bash
readonly SCRIPT_ROOT=$(dirname $0)
source ${SCRIPT_ROOT}/logging.sh
```

For now, I'm going to postpone this refactor until we get another bash script that needs these functions, at which point I'll extract the library (or move to a different language :) )

Scripts/generate-test-config.sh
palpatim: That's another one I considered but rejected, because the non-indented form makes it easier to see the width of the text as it will be output at runtime. If you're aware of a good way of transparently handling text wrapping I'll reconsider this, but since I'm currently manually handling line breaks in the heredoc, I wanted to know the "real" width without having to mentally subtract the indention space.

Scripts/generate-test-config.sh
palpatim: I never realized `[[` isn't a builtin--the docs certainly imply it's a builtin . I'm going to keep it for now since it supports standard constructs like `[[ $foo -eq "foo" ]]` where $foo isn't actually set, whereas `[ $foo -eq "foo" ]` generates a 'unary operator' error unless I wrap $foo in quotes (`[ "$foo" -eq "foo" ]`), which I always neglect to do.

Scripts/generate-test-config.sh
palpatim: Yeah, the loop is a better construct here. I've never quite gotten bash arrays under my fingers, so I don't usually remember to use them.

Scripts/generate-test-config.sh
palpatim: Thanks for the callout. That's a leftover from me originally specifying paths to executable dependencies as hardcoded values rather than auto-discovering them with `which`. Funny thing is, I didn't even use `$jq_cmd` and `$aws_cli_cmd` down where I actually invoke it, so the assignment wasn't going to be useful anyway. Sheesh.

I'm not going to be overly paranoid about the path, since these are going to be invoked interactively by developers who may have installed into non-traditional paths (e.g., python in pyenv shim folders, homebrew-managed packages, etc).

Scripts/generate-test-config.sh
palpatim: Yep, done.

Scripts/generate-test-config.sh
palpatim: In my experience, yes; the `+` with normal `strftime` symbols seems to be pretty universal. I've had mixed results with extensions like "natural language" date math ("today", "7 days ago"), which are supported on some variants and not others, but I've never had problems with simple timestamp formats of the current date.

jpignata: Seems to, FWIW:

*macOS*
```console
$ date +%Y%m%d%H%M%S   
20200506103905
```

*Linux*
```console
$ date +%Y%m%d%H%M%S
20200506143909
```

Scripts/generate-test-config.sh
```diff
@@ -1,14 +1,269 @@
 #!/bin/bash
```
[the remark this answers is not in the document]
palpatim: > I'm slightly concerned with the complexity of the script and wonder if we're trying too hard to make this automatic. It might be better to just tell people you need boto3 and Python and leave it at that, but, if this portably works and doesn't create problems, 👍!

Regarding `python3`, `aws` and `jq`: There's no automatic installation of these dependency utilities, just validation that they exist, which is something that we would have to do anyway or face weird errors during invocation. 

Regarding `boto3` and any other deps in the `requirements.txt`: I'm far from a Python expert, but the `venv` pattern here seems to be a pretty common way of encapsulating dependencies without polluting a global environment. And I really want the DX to be as one-shot as possible. The most brittle part of this seems to me to be the hardcoded assumptions about the paths, which I'm not entirely sure how to mitigate--see note below about the canonical home of this script.

The other main chunk of complexity is around the code to automatically assume the role, which is required since the tests require the config file to be written with assumed-role creds of the test execution role. Offloading that work from the customer (even if it were simply copy/paste commands) is complexity I'm willing to pay for, especially since the primary customer for that use case is me. :)

> Does this script belong in aws-amplify/amplify-ci-support as its canonical home?

Yes, definitely. The catch is that this needs to be invoked in platform-specific builds. Maybe we could do something like git submodules, or some other way of bootstrapping the dependency into the various places it needs to live? I'm open to suggestions. For Phase 1 though, it seems like a reasonable compromise to master this in the ci-support repo and manually copy it over to the platform-specific repos.

Scripts/generate-test-config.sh
palpatim: Thanks for the links to Shellguide & Shellcheck; I hadn't seen those before. The uppercase symbols are constants, which the checker had no way of knowing because I wasn't using the `readonly` keyword as @jamesonwilliams pointed out. I'll look at the recommendations from Shellcheck & address

PR #2552, opened 2020-05-07
bug: [Pinpoint] persist explicit set for optOut over default behavior

lawmicha:
*Issue #, if available:*

*Description of changes:*
Persist the optOut value set by the developer, ie.
```
self.pinpoint.targetingClient.currentEndpointProfile.optOut = @"ALL";
or
self.pinpoint.targetingClient.currentEndpointProfile.optOut = @"NONE";
```
in the user defaults to make sure it is used across app restarts.

By default, we maintain the current behavior for checking if the app has remote notifications enabled and allowed and address is set (pinpoint.intercept(deviceToken))

### DIFF
--- a/AWSPinpoint/AWSPinpointEndpointProfile.m
+++ b/AWSPinpoint/AWSPinpointEndpointProfile.m
@@ -25,6 +25,7 @@ static int const MAX_NUM_OF_METRICS_AND_ATTRIBUTES = 20;
 static int const MAX_ENDPOINT_ATTRIBUTE_METRIC_KEY_LENGTH = 50;
 static int const MAX_ENDPOINT_ATTRIBUTE_VALUE_LENGTH = 100;
 static int const MAX_ENDPOINT_ATTRIBUTE_VALUES = 50;
+NSString *const AWSPinpointOverrideDefaultOptOutKey = @"com.amazonaws.AWSPinpointOverrideDefaultOptOutKey";
 
 @interface AWSPinpointEndpointProfile()
 @property (nonatomic, readwrite) NSString *endpointId;
@@ -33,6 +34,7 @@ static int const MAX_ENDPOINT_ATTRIBUTE_VALUES = 50;
 @property (nonatomic, readwrite) NSDictionary<NSString*,NSNumber*> *metrics;
 @property (nonatomic, readwrite) UTCTimeMillis effectiveDate;
 @property (atomic, readonly) int currentNumOfAttributesAndMetrics;
+@property (nonatomic, strong) NSUserDefaults *userDefaults;
 
 @end
 
@@ -42,6 +44,8 @@ static int const MAX_ENDPOINT_ATTRIBUTE_VALUES = 50;
 
 #pragma mark - AWSPinpointEndpointProfile -
 @implementation AWSPinpointEndpointProfile
+
+@synthesize optOut = _optOutBackingVariable;
 
 NSString *CHANNEL_TYPE = @"APNS";
 NSString *DEBUG_CHANNEL_TYPE = @"APNS_SANDBOX";
@@ -77,6 +81,7 @@ NSString *DEBUG_CHANNEL_TYPE = @"APNS_SANDBOX";
         _attributes = [NSMutableDictionary dictionary];
         _metrics = [NSMutableDictionary dictionary];
         _user = [AWSPinpointEndpointProfileUser new];
+        _userDefaults = userDefaults;
     }
     return self;
 }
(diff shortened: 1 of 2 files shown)


AWSPinpoint/AWSPinpointEndpointProfile.m
```diff
@@ -121,12 +126,32 @@ - (BOOL) isApplicationLevelOptOut:(AWSPinpointContext *) context {
...
+}
+
 - (void) setEndpointOptOut:(BOOL) applicationLevelOptOut {
-    BOOL isUsingPinpointForNotifications = [AWSPinpointNotificationManager isNotificationEnabled] && [self.address length];
-    BOOL isOptedOutForNotifications = !isUsingPinpointForNotifications;
+    NSString *overrideDefaultOptOut = [self.userDefaults stringForKey:AWSPinpointOverrideDefaultOptOutKey];
+    if ([overrideDefaultOptOut length]) {
+        @synchronized (self) {
+            self->_optOut = overrideDefaultOptOut;
+        }
```
drochetti, line 147: won't this result in the value being `"NONE"` even when the notifications are disabled at an app level or the `self.address` is missing?

lawmicha: I don't think i fully understand the scenario ur talking about, but if the developer sets the value to `"NONE"`, it will persist in the user defaults, so it will always update the _optOut to the value that it was set in the user defaults, in this case we respect what is there and set it to `"NONE"`. 

if we look at a scenario where the optOut is not explicitly set by the developer, it will take on the value of ALL or NONE as before. NONE only when both UIApplication.isRemoteNotificationEnabled && devicetoken/address is set. otherwise set it to ALL.

If the developer builds a toggle that sets it to ALL or NONE (or anything for that matter), this will be immediately persisted in user defaults, and always take precedent over the default behavior

AWSPinpoint/AWSPinpointEndpointProfile.m
```diff
@@ -121,12 +126,32 @@ - (BOOL) isApplicationLevelOptOut:(AWSPinpointContext *) context {
...
+    if (![overrideDefaultOptOut isEqualToString:optOut]) {
+        [self.userDefaults setObject:optOut forKey:AWSPinpointOverrideDefaultOptOutKey];
+        [self.userDefaults synchronize];
+    }
+
+    _optOut = optOut;
+}
+- (NSString *)optOut {
+    return _optOut;
+}
```
wooj2, line 140: It is generally advisable not to override setters and getters for properties.  I think it would be better to introduce a public API which allows them to override the optValue  perhaps something like
```
- (void) overrideOptOut:(NSString *)optOut {
        self.optOut = optOut;
        [self.userDefaults setObject:optOut forKey:AWSPinpointOverrideDefaultOptOutKey];
        [self.userDefaults synchronize];
}
```

AWSPinpoint/AWSPinpointEndpointProfile.m
drochetti: so the scenario is:
- developer sets to `"NONE"`, `overrideDefaultOptOut ` is now `"NONE"`
- `address` is `nil`
- `optOut` is submitted as `"NONE"`
- *"missing address error"*

lawmicha: i think service response is out of our control. but we can avoid it by doing `if overrideDefaultOptOut == "NONE" and _address == nii, set it to ALL`

drochetti: I think the previous logic needs little modification to work with the newly introduced user override:

```objc
// line 129 was:
self->_optOut = (applicationLevelOptOut || isOptedOutForNotifications)? @"ALL": @"NONE";

// it needs to change to
self->_optOut = (applicationLevelOptOut || isOptedOutForNotifications)? @"ALL": USER_SET_VALUE_OR_NONE_BY_DEFAULT;
```

lawmicha: thanks, updating this and added unit test for setting to `None` but one of the other is not set, will result in `ALL`

AWSPinpoint/AWSPinpointEndpointProfile.m
```diff
@@ -121,12 +126,27 @@ - (BOOL) isApplicationLevelOptOut:(AWSPinpointContext *) context {
...
+- (void) setOptOut:(NSString *)optOut {
+    NSString *overrideDefaultOptOut = [self.userDefaults stringForKey:AWSPinpointOverrideDefaultOptOutKey];
+    if (![overrideDefaultOptOut isEqualToString:optOut]) {
+        [self.userDefaults setObject:optOut forKey:AWSPinpointOverrideDefaultOptOutKey];
+        [self.userDefaults synchronize];
+    }
+
+    _optOutBackingVariable = optOut;
+}
+- (NSString *)optOut {
```
palpatim, line 138: whitespace

AWSPinpoint/AWSPinpointEndpointProfile.m
lawmicha: resolving tihs convo- we chatted about this and decided to keep the custom setting 
- developers are already using this in their code, but SDK was not respecting this. those facing the bug will be fixed if they upgrade to the latest SDK once this is released
- Android's optOut is read-only, platform parity means Android's pinpoint sdk will also need to make this field writable.

Issue #2504, opened 2020-04-28
Stuck on getSession().continueWith after CustomAuthentication


abonko:
**Describe the bug**
Hi! We use CustomAuthentification to exclude password from signing flow - only sms code into AWSCognitoIdentityCustomAuthentication

After handling success didCompleteStepWithError we try to get session from current user.

```
  let awsUserPool = AWSCognitoIdentityUserPool(forKey: AWSCognitoUserPoolsSignInProviderKey)
                
                var accessTokenString = ""
                var idTokenString = ""
                let currentUser = awsUserPool.currentUser()
                
                currentUser?.getSession().continueWith(block: { (session) -> Any? in
                    if let accessToken = session.result?.accessToken,
                        let idToken = session.result?.idToken {
                        accessTokenString = accessToken.tokenString
                        idTokenString = idToken.tokenString
                        observer(.success((accessTokenString, idTokenString)))
                    } else {
                        observer(.error(RxError.noElements))
                    }
                    return nil
                })
```

99% of users signed in success, but some users have problem with getting session after Confirming code. `getSession().continueWith(block: { (session) -> Any? in` block has never called. We tried to retry 3 times with 3 second delay but had no success.
**To Reproduce**
I don't know how to reproduce it because this problem is very intermittent 

**Observed Behavior**

**Expected Behavior**
`getSession().continueWith(block: { (session) -> Any? in` block should be called

**Unique Configuration**
Using customAuthenticationCompletion for AWSCognitoIdentityInteractiveAuthenticationDelegate

**Areas of the SDK you are using (AWSMobileClient, Cognito, Pinpoint, IoT, etc)?**
Signin and Signup

**Environment(please complete the following information):**
 - SDK Version: 2.12.6
 - Dependency Manager: Cocoapods
 - Swift Version : 5.0
 - Xcode Version: 11.4

**Device Information (please complete the following information):**
 - Device: any devices
 - iOS Version: start from iOS 11

palpatim:
`getSession()` can block on other Cognito delegate methods. Have you fully implemented all of the [User Pool delegate methods](https://aws-amplify.github.io/aws-sdk-ios/docs/reference/AWSCognitoIdentityProvider/Protocols/AWSCognitoIdentityInteractiveAuthenticationDelegate.html) to ensure there isn't a case where some of your users are being caught by these? Even though you're implementing custom auth, I wonder if there may be some case that is triggering one of these callbacks.

[...]

Issue #1278, opened 2019-03-11
Crash in AWSMobileClient.getTokens()


zigfreid5:
I'm getting crash reports signaling a crash in AWSMobileClient.getTokens(). This is possibly occuring when `releaseSignInWait()` is called without an active request waiting for sign in? Here's the stacktrace:

```
Crashed: NSOperationQueue 0x283dc0b20 (QOS: UNSPECIFIED)
0  libdispatch.dylib              0x235bc25e8 dispatch_group_leave + 124
1  AWSMobileClient                0x1054e0c60 closure #1 in closure #1 in AWSMobileClient.getTokens(_:) (AWSMobileClientExtensions.swift:719)
2  AWSMobileClient                0x1054db260 thunk for @escaping @callee_guaranteed (@guaranteed AWSTask<AWSCredentials>) -> (@out Any?) (<compiler-generated>)
3  AWSCore                        0x10569322c __56-[AWSTask continueWithExecutor:block:cancellationToken:]_block_invoke (AWSTask.m:331)
4  AWSCore                        0x105660670 __30+[AWSExecutor defaultExecutor]_block_invoke_2 (AWSExecutor.m:66)
5  AWSCore                        0x105660bdc -[AWSExecutor execute:] (AWSExecutor.m:131)
6  AWSCore                        0x105692e24 -[AWSTask runContinuations] (AWSTask.m:308)
7  AWSCore                        0x105692ab0 -[AWSTask trySetError:] (AWSTask.m:266)
8  AWSCore                        0x105694048 -[AWSTaskCompletionSource setError:] (AWSTaskCompletionSource.m:52)
9  AWSCore                        0x10569343c __56-[AWSTask continueWithExecutor:block:cancellationToken:]_block_invoke_2 (AWSTask.m:340)
10 AWSCore                        0x1056932e4 __56-[AWSTask continueWithExecutor:block:cancellationToken:]_block_invoke (AWSTask.m:348)
11 AWSCore                        0x105660670 __30+[AWSExecutor defaultExecutor]_block_invoke_2 (AWSExecutor.m:66)
12 AWSCore                        0x105660bdc -[AWSExecutor execute:] (AWSExecutor.m:131)
13 AWSCore                        0x105692e24 -[AWSTask runContinuations] (AWSTask.m:308)
14 AWSCore                        0x105692ab0 -[AWSTask trySetError:] (AWSTask.m:266)
[...]

PR #2571, opened 2020-05-11
fix(AWSMobileClient): Fixes the crash due to over-leaving dispatch group

royjit:
## Changes

- Removed `tokenFetchLock` leave from releaseSignInWait - Issues #1278 
- Fixed releaseSignInWait for custom Auth
- Removed cyclic retry logic in custom AuthIssues - Issue #2504 
- Added integration test for the changes

## Explanation of tokenFetLock

leave `tokenFetchLock` dispatch group  for userPool federation is already handled in getTokens api. So there is no need to call this again.

`releaseSignInWait` is to release any waiting call that require signIn to happen. SignIn wait could be triggered from 3 scenarios, and this PR is for Scenario 1.

**Scenario 1 - Signed in to userPool through non-UI api**

In this case signIn wait is triggered from getToken api. Here is the flow:
1. getToken call is invoked and enters tokenFetchLock 
1. getToken internally calls getSession from the underlying Cognito SDK.
1. tokenFetchLock waits till we get a result back from getSession
1. getSession can behave in these three ways - Return a valid session, return an error, ask the user to signIn. Last behavior is the signIn wait, based on the signIn result getSession returns a session or error.
1. Point to note here is that in all the three behavior above, the result always returns to getSession call, which release the tokenFetchLock. So there is no need to release it outside from getToken.

**Scenario 2 - Signed in to userPool through Hosted UI** 

In this case signIn wait is again triggered from getToken api. Here is the flow:
1. getToken call is invoked and enters tokenFetchLock 
1. getToken internally calls getSession from AWSCognito Auth
1. tokenFetchLock waits till we get a result back from getSession
1. If getSession returns refreshToken Expired error, it leaves out from getToken with the tokenFetchLock active.
1. In this case it is okay to leave from the dispatch group outside of getToken api.

**Scenario 3 - Signed in to Identity Pool** 

In this case signIn wait is triggered from getAWSCredentials api. Here is the flow:
1. getAWSCredentials call is invoked and enters credentialsFetchLock 
1. getAWSCredentials internally calls `credentials` from underlying credentials Provider
1. credentialsFetchLock waits till we get a result back from `credentials`
1. If `credentials` returns notAuthorized error, it leaves out from getAWSCredentials with the credentialsFetchLock active.
1. In this case it is okay to leave from the dispatch group outside of getAWSCredentials api.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@
 - **Integration tests**
     - AWS Mobile SDK for iOS integration tests are now provisioned from a CloudFormation stack created by [the new amplify-ci-support package](LINK TBD). See [the README](LINK TBD) for details on how to provision your account to run integration tests.
 
+### Bug Fixes
+- **AWSMobileClient**
+    - Crash in `releaseSignInWait` api due to over fulfilling a fetchlock. See [Issue #1278](https://github.com/aws-amplify/aws-sdk-ios/issues/1278) and [PR #2571](https://github.com/aws-amplify/aws-sdk-ios/pull/2571)
+    - Fix `releaseSignInWait` for custom Auth to clear the lock correctly. [PR #2571](https://github.com/aws-amplify/aws-sdk-ios/pull/2571)
+
+- **AWSCognitoIdentityProvider**
+    - Removed cyclic retry logic in custom auth delegate. [PR #2571](https://github.com/aws-amplify/aws-sdk-ios/pull/2571)
+
 ### Misc. Updates
 - Model updates for the following services:
   - AWS Lambda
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,14 @@
 - **Integration tests**
     - AWS Mobile SDK for iOS integration tests are now provisioned from a CloudFormation stack created by [the new amplify-ci-support package](LINK TBD). See [the README](LINK TBD) for details on how to provision your account to run integration tests.
 
+### Bug Fixes
+- **AWSMobileClient**
+    - Crash in `releaseSignInWait` api due to over fulfilling a fetchlock. See [Issue #1278](https://github.com/aws-amplify/aws-sdk-ios/issues/1278) and [PR #2571](https://github.com/aws-amplify/aws-sdk-ios/pull/2571)
+    - Fix `releaseSignInWait` for custom Auth to clear the lock correctly. [PR #2571](https://github.com/aws-amplify/aws-sdk-ios/pull/2571)
+
+- **AWSCognitoIdentityProvider**
+    - Removed cyclic retry logic in custom auth delegate. [PR #2571](https://github.com/aws-amplify/aws-sdk-ios/pull/2571)
```
palpatim, line 14: Also reference #2504

PR #2578, opened 2020-05-12
Integration test updates

palpatim:
The bulk of these changes are mechanical:
- Remove `credentials.json`, `testconfiguration.json`, and `credentials-mc.json` references from test projects
- Add references to the new `AWSTestResources` package in the test project file
- Modify the tests to use the new session-based credentials instead of old hardcoded credentials or Cognito credentials. Unless we're specifically testing different auth type behaviors, we should always use simple session-based credentials.

In addition to these mechanical changes, I made the following substantive changes:
- Removed FreeRTOS tests from normal test runs
- Removed AWSConnect and AWSConnectParticipant tests since they don't support setup via CDK. We can explore re-enabling these in the future, or adding them as manual-only tests with explicit instructions on how to manually set up the backend for test support.
  - **NOTE:** I forgot to disable the integ test runs in the CircleCI config and integration test manifest. I'll update that on a subsequent PR.
- Cleaned old methods, and added new convenience methods, to `AWSTestUtility`
- Removed AWSMachineLearning tests since we're advising customers to use other services
- Quite a bit of refactoring in S3 tests, including removing redundant "large file" tests from TransferManager, since they didn't test anything importantly different from the existing "small file" tests, and since TM is deprecated in favor of TransferUtility.
- Added a session timeout to the AWSPinpointBackgroundBehaviorTests setup. I'm not sure why the tests ever passed previously except that the timing worked out--my read of the code is that background event processing would only happen if the session timeout were > 0.

This effort is not complete, in that there are remaining tests to migrate, but the work that is here is required to complete the overall project. Thus, once we get this PR approved, I will merge it into develop as an incremental improvement to the project state, even though it doesn't completely fix the broken build/broken test conditions.


AWSComprehendTests/AWSComprehendTests.swift
```diff
@@ -50,8 +54,12 @@ class AWSComprehendTests: XCTestCase {
         detectDominantLanguageRequest!.text = "El sol brilla en Seattle";
         
         comprehendClient.detectDominantLanguage(detectDominantLanguageRequest!).continueWith { (task) -> Any? in
+            if let error = task.error {
+                XCTAssertNil(error)
```
royjit, line 58: Since this statement is for better understanding of error in the test report, can we do something like this 
```
XCTFail("Should not produce error - \(error)")
```

AWSAuthSDK/Tests/AWSMobileClientTests/AWSMobileClientTestBase.swift
```diff
@@ -11,28 +11,32 @@ import AWSCognitoIdentityProvider
...
     static var cognitoIdentity: AWSCognitoIdentity!
     static var userPoolsAdminClient: AWSCognitoIdentityProvider!
     
     static var userPoolId: String!
     static var sharedEmail: String!
     static var identityPoolId: String!
-
+    
     // Test password that conforms to default UserPool password policies
     let sharedPassword: String = "Abc123@@!!"
```
royjit, line 23: Should we move to the configuration file?

AWSComprehendTests/AWSComprehendTests.swift
palpatim: I'll see if stringifying the error produces the same level of error detail as simply dumping it in the `XCTAssertNil`. If not, then I can change the message to `XCTAssertNil(error, "Should not produce error")`

AWSAuthSDK/Tests/AWSMobileClientTests/AWSMobileClientTestBase.swift
palpatim: Yeah, probably so. I'll move in a subsequent PR

PR #2587, opened 2020-05-14
Integ test updates continued

palpatim:
There are still known test failures (see below), but this incremental work brings us closer to stable tests sourced from the ci-support stack.

Important differences:
- AWSMobileClient test refactoring to use `testconfiguration.json` for all configurations, including the tests that formerly used `awsconfiguration.json`
- Cleaned up test host dependencies throughout. In general, the test host isn't needed, but there are some cases where it's required. Notably, any tests that need the keychain, or use read/write temporary directory resources, should have an app host.
- `AWSTestResources` **Copy testconfiguration.json** build phase now copies the `testconfiguration.json` file into the target product directory instead of symlinking it--symlinks don't play well with sandboxed simulator apps apparently.
- Removed several integration test cases from pipeline runs, including Connect, ConnectParticipant, and SageMakerRuntime. Connect and ConnectParticipant are still in the project and could be provisioned to run manually, but we should discuss whether it's better to just remove them altogether, since they cannot be provisioned via CDK.
  - This change also involved removing jobs from the CircleCI config file, which "shouldn't" impact builds and releases.

**Known failing test cases**
- AWSAPIGatewayTests:
  - This test relies on a code-generated Lambda client and backend resources. Needs investigation to see the best way to fix.
- AWSMobileClientTests:
  - All UI tests (note that some of these were failing even before)
  - Custom auth tests (we haven't yet provisioned the custom auth lambdas in the stack)
  - `testUploadFileInUserStateListener`. It's unclear to me why this is failing, but I also can't see how any changes to the test configuration would have affected this. Needs more investigation
  - `testPasswordResetChallenge`. Needs investigation
  - `testFederatedSignInDeveloperAuthenticatedIdentities`. I believe we are still missing some stack resources. Needs investigation.


AWSAuthSDK/AWSAuthSDKTestAppUITests/AWSHostedUIUserPoolTests.swift
```diff
@@ -4,20 +4,18 @@
...
-        let filePath = Bundle(for: self).path(forResource: "credentials-mc", ofType: "json")
-        var credentialsJson: [AnyHashable : Any]? = nil
-        if let aPath = NSData(contentsOfFile: filePath ?? "") {
-            credentialsJson = try! JSONSerialization.jsonObject(with: aPath as Data, options: .mutableContainers) as? [AnyHashable : Any]
-        }
-        userpoolUsername = credentialsJson?["UserpoolUsername"] as? String
-        userpoolPassword = credentialsJson?["UserpoolPassword"] as? String
+        userpoolUsername = AWSTestConfiguration.getIntegrationTestConfigurationValue(forPackageId: "mobileclient",
+                                                                                     configKey: "userpool_username")
+        userpoolUsername = AWSTestConfiguration.getIntegrationTestConfigurationValue(forPackageId: "mobileclient",
```
royjit, line 17: userpoolPassword

AWSAuthSDK/AWSAuthSDKTestAppUITests/AWSDropInUIUserPoolTests.swift
```diff
@@ -15,21 +16,27 @@ class AWSDropInUIUserPoolTests: XCTestCase {
...
-        userPoolId = credentialsJson?["mc-userpool_id"] as? String
-        sharedEmail = credentialsJson?["mc-email"] as? String
-        
-        let credentialsProvider = AWSStaticCredentialsProvider(accessKey: credentialsJson?["accessKey"] as! String,
-                                                               secretKey: credentialsJson?["secretKey"] as! String)
-        let region = (credentialsJson?["mc-region"] as! String).aws_regionTypeValue()
-        let configuration = AWSServiceConfiguration(region: region, credentialsProvider: credentialsProvider)!
+        userpoolUsername = AWSTestConfiguration.getIntegrationTestConfigurationValue(forPackageId: "mobileclient",
+                                                                                     configKey: "userpool_username")
+        userpoolUsername = AWSTestConfiguration.getIntegrationTestConfigurationValue(forPackageId: "mobileclient",
```
royjit, line 21: userpoolPassword

AWSAuthSDK/AWSAuthSDKTestAppUITests/AWSHostedUIUserPoolTests.swift
palpatim: Good grief. Will fix.

AWSAuthSDK/AWSAuthSDKTestAppUITests/AWSDropInUIUserPoolTests.swift
palpatim: :sigh: Will fix.

Issue #2598, opened 2020-05-16
awscognitoauthclienterror check always fails


milk531:
https://github.com/aws-amplify/aws-sdk-ios/blob/0f8eb78dcde9d22806ebfcdaa2c15789511965ae/AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClientExtensions.swift#L739

AWSCognitoAuthClientErrorType and Error are not the same type, so it never works as expected

eroscai:
Yes, now when the user's token is completely invalid, requesting the getTokens method will trigger an errorExpiredRefreshToken(-7000) error. At that time, as mentioned above, this conversion method will always fail, resulting in our listeners never having the opportunity to get signedOutUserPoolsTokenInvalid notification.

The right code should be like below:

```swift
if let error = error, let cognitoError = AWSCognitoAuthClientErrorType(rawValue: (error as NSError).code) {
    if cognitoError == AWSCognitoAuthClientErrorType.errorExpiredRefreshToken {
        ...
    } else {
        ...
    }
}
```

royjit:
Thank you for reporting the issue. I have added a PR to address this issue here - https://github.com/aws-amplify/aws-sdk-ios/pull/2739

PR #2739, opened 2020-06-12
fix: HostedUI will now send  signedOutUserPoolsTokenInvalid when the token expires

royjit:
*Issue : https://github.com/aws-amplify/aws-sdk-ios/issues/2598

*Description of changes:*
If the token expired while signed in using HostedUI, `AWSMobileClient` will now send the event `.signedOutUserPoolsTokenInvalid` and the request that invoked the token fetch will wait till the user is signed in. In the previous implementation the requested operation will complete with error and user event listener was not invoked.

### DIFF
--- a/AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClientExtensions.swift
+++ b/AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClientExtensions.swift
@@ -342,7 +342,7 @@ extension AWSMobileClient {
                 } else {
                     confirmedStatus = .unconfirmed
                 }
-                
+
                 var codeDeliveryDetails: UserCodeDeliveryDetails? = nil
                 if let deliveryDetails = result.codeDeliveryDetails {
                     switch(deliveryDetails.deliveryMedium) {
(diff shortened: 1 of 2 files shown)


AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClientExtensions.swift
```diff
@@ -745,15 +745,16 @@ extension AWSMobileClient {
...
-                            // return early without releasing the tokenFetch lock.
-                            return
-                        } else {
-                            completionHandler(nil, AWSMobileClientError.makeMobileClientError(from: error))
-                        }
+
+                    if let sessionError = error,
+                        (sessionError as NSError).domain == AWSCognitoAuthErrorDomain,
+                        let errorType = AWSCognitoAuthClientErrorType(rawValue: (sessionError as NSError).code),
+                        (errorType == .errorExpiredRefreshToken) {
```
lawmicha, line 752: maybe `if isExpiredRefreshToken(error)` if it can be reused in other places?

royjit: After this change HostedUI and non-HostedUI should work the same. ie Whenever the session expires they will be notified by the user listener.

PR #2764, opened 2020-06-18
AWSSesv2 Update models to latest

awsmobilesdk:
AWSSesv2 Update models to latest


palpatim:
This contains many breaking changes. The Model file hasn't been updated in 1.5 years, and the Service file contains many breaks even though it was last updated in January 2020. Need to investigate to see why this is such a disruptive change.

PR #2807, opened 2020-06-26
AWSCognitoidentityprovider Update models to latest

awsmobilesdk:
AWSCognitoidentityprovider Update models to latest


palpatim:
We'll regenerate this later after the code generation fixes are in place to hoist files from the `CognitoIdentityProvider` subdirectory.

PR #2815, opened 2020-06-29
fix: Added nullable modifier to the return instance form the method CognitoIdentityUserPoolForKey:

royjit:
*Description of changes:* 

The implementation of the method `CognitoIdentityUserPoolForKey:` looks like this:
```
+ (instancetype)CognitoIdentityUserPoolForKey:(NSString *)key {
    return [_serviceClients objectForKey:key];
}
```
Internally it just returns an item from a dictionary. And this can return a null object back.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 ### Bug Fixes
 
 - **AWSCognitoIdentityProvider**
-  - Added `nullable` modifier to the return instance form the method `CognitoIdentityUserPoolForKey:`. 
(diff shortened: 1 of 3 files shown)


palpatim:
Note: This commit is a breaking change and should be indicated as such in the commit message.

CHANGELOG.md
```diff
@@ -3,6 +3,13 @@
 ## Unreleased
 -Features for next release
 
+## 2.14.0
+
+### Bug Fixes
+
+- **AWSCognitoIdentityProvider**
```
palpatim, line 10: Note this as a breaking change

PR #2842, opened 2020-07-08
ci: Fix the failing package step for iOS > 10

royjit:
*Issue #, if available:*

*Description of changes:*


CircleciScripts/package.sh
```diff
@@ -64,7 +64,7 @@ if [ -f "CircleciScripts/package_sdk.sh" ]; then
 
     CircleciScripts/package_sdk.sh AWSAuthCore "$(pwd)/AWSAuthSDK/AWSAuthSDK.xcodeproj"
     CircleciScripts/package_sdk.sh AWSAuthUI "$(pwd)/AWSAuthSDK/AWSAuthSDK.xcodeproj"
-    CircleciScripts/package_sdk.sh AWSAppleSignIn "$(pwd)/AWSAuthSDK/AWSAuthSDK.xcodeproj"
+    CircleciScripts/package_sdk.sh AWSAppleSignIn "$(pwd)/AWSAuthSDK/AWSAuthSDK.xcodeproj" "yes"
```
palpatim, line 67: "yes" isn't terribly meaningful as an argument with no context. Since you have the ability to define this as whatever string you want, let's make it something like "buildOnly64Bit" or similar.

PR #2846, opened 2020-07-09
AWSSecretsManager Update models to latest

awsmobilesdk:
AWSSecretsManager Update models to latest


phani-srikar:
This is an unsupported model but we had discussions at some point, if we want to support this.  If we still want to support this, please check if it builds successfully before merging it in. Otherwise, close the PR and I will remove this from supported models catalog.

phani-srikar:
Android team decided not to support this until need arises [link to discussion](https://github.com/aws-amplify/aws-sdk-android/pull/2002). To maintain consistency between iOS/android for supported services, closing this PR and removing this from supported models until there is a customer/internal ask for it.

PR #2855, opened 2020-07-13
chore: Deprecate AWSiOSSDKv2 aggregate pod in favor of Amplify

palpatim:
Deprecated the aggregate pod and updated AWSMobileAnalytics and AWSCognitoSync with deprecation messages referring customers to Amplify. We've discussed the deprecation messages internally and agreed that they are fine.


palpatim:
Leaving this as draft while we verify the messaging to make sure it's not implying that the AWS Mobile SDK itself is being deprecated.

palpatim:
Deprecation messages approved, ready for review

PR #2856, opened 2020-07-15
Add username and password to MQTT ioTMQTTConfiguration for enhanced custom authorizer

wooj2:
To use this:
```
        let ioTMQTTConfiguration = AWSIoTMQTTConfiguration()
        ioTMQTTConfiguration.username = "<YOURUSERNAMEHERE>"
        ioTMQTTConfiguration.password = "<YOURPASSWORDHERE>"
        AWSIoTDataManager.register(with: iotDataConfiguration!, with: ioTMQTTConfiguration, forKey: AWS_IOT_DATA_MANAGER_KEY)
        iotDataManager = AWSIoTDataManager(forKey: AWS_IOT_DATA_MANAGER_KEY)
```

As a result, when connecting over a websocket & mqtt, the username and password will be sent with the MQTT hello message.

### DIFF
--- a/AWSIoT/AWSIoTDataManager.m
+++ b/AWSIoT/AWSIoTDataManager.m
@@ -335,9 +335,8 @@ static AWSSynchronizedMutableDictionary *_serviceClients = nil;
 }
 
 - (nonnull NSString*)baseUserMetaDataString:(nullable NSString*)username {
-    return username.length
-    ? [NSString stringWithFormat:@"%@%@%@", username, @"?SDK=iOS&Version=", AWSIoTSDKVersion]
-    : [NSString stringWithFormat:@"%@%@", @"?SDK=iOS&Version=", AWSIoTSDKVersion];
+    NSString *usernameComponent = username.length ? username : @"";
+    return [NSString stringWithFormat:@"%@?SDK=iOS&Version=%@", usernameComponent, AWSIoTSDKVersion];
 }
 
 - (void)enableMetricsCollection:(BOOL)enabled {
--- a/AWSIoT/AWSIoTDataManager.m
+++ b/AWSIoT/AWSIoTDataManager.m
@@ -323,13 +323,21 @@ static AWSSynchronizedMutableDictionary *_serviceClients = nil;
         if(_mqttClient == nil){
             AWSDDLogError(@"**** mqttClient is nil. **** ");
         }
-        _mqttClient.userMetaData = [NSString stringWithFormat:@"%@%@", @"?SDK=iOS&Version=", AWSIoTSDKVersion];
+
+        _mqttClient.userMetaData = [self baseUserMetaDataString:mqttConfig.username];
+        _mqttClient.password = mqttConfig.password.length ? mqttConfig.password : @"";
         _userMetaDataDict = [[NSMutableDictionary alloc] init];
         _mqttClient.associatedObject = self;
         _userDidIssueDisconnect = NO;
         _userDidIssueConnect = NO;
     }
     return self;
+}
+
+- (nonnull NSString*)baseUserMetaDataString:(nullable NSString*)username {
+    return username.length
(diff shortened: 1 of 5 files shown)


AWSIoTUnitTests/AWSIoTDataUnitTests.m
```diff
@@ -164,6 +164,30 @@ -(void)testRegisterIoTDataManagerWithConfigurationWithMQTTConfiguration {
 
 }
 
+-(void)testRegisterIoTDataManagerWithConfigurationUserNamePasswordWithMQTTConfiguration {
+    NSString *key = @"testRegisterIoTDataManagerWithMQTTConfigurationUserNamePassword";
+    AWSCognitoCredentialsProvider *credentialsProvider =
+    [[AWSCognitoCredentialsProvider alloc] initWithRegionType:AWSRegionUSWest2
```
DongQuanRui, line 170: So far, all looks good. 
nit picking: Only here, since in https://docs.aws.amazon.com/iot/latest/developerguide/enhanced-custom-auth-using.html, it is mentioned is available only in the US East (N. Virginia) Region. Although the test is not actually talking to backend service. should we change it to `AWSRegionUSEast1`. 😂
```suggestion
    [[AWSCognitoCredentialsProvider alloc] initWithRegionType:AWSRegionUSWest2
```

AWSIoT/AWSIoTDataManager.m
```diff
@@ -332,6 +334,12 @@ - (instancetype)initWithConfiguration:(AWSServiceConfiguration *)configuration w
     return self;
 }
 
+- (nonnull NSString*)baseUserMetaDataString:(nullable NSString*)username {
+    return username.length
+    ? [NSString stringWithFormat:@"%@%@%@", username, @"?SDK=iOS&Version=", AWSIoTSDKVersion]
```
palpatim, line 339: You can simplify this by putting the hardcoded string as part of the format specifier: `[NSString stringWithFormat:@"%@?SDK=iOS&Version=%@", username, AWSIoTSDKVersion]`. You could simplify it further by assigning the username token then returning the same formatted string for each:

```objective-c
NSString *usernameComponent = username.length ? username : @"";
return [NSString stringWithFormat:@"%@?SDK=iOS&Version=%@", usernameComponent, AWSIoTSDKVersion];
```

My recollection is that interpolating a `nil` username into a string results in a string like `"(null)"` or similar, so the assignment is probably a good idea

AWSIoT/Internal/AWSIoTMQTTClient.h
```diff
@@ -91,6 +91,7 @@
 @property(atomic, assign) BOOL isMetricsEnabled;
 @property(atomic, assign) NSUInteger publishRetryThrottle;
 @property(atomic, strong) NSString *userMetaData;
+@property(atomic, copy) NSString *password;
```
palpatim, line 94: Why `copy` rather than `strong`, and would it then make sense to update `userMetaData` to `copy` as well?

AWSIoTUnitTests/AWSIoTDataUnitTests.m
palpatim: I would prefer not to make that distinction in unit tests. If we have a codepath that checks for existence of a beta endpoint, let's test it explicitly elsewhere, but otherwise, we should assume that unit tests are region-independent. If we use USWest2 elsewhere for whatever reason, we can continue to use that here.

Integ tests are a whole other story: be sure to set up resources for them through the integ test infrastructure scripts, as discussed with @wooj2.

AWSIoT/Internal/AWSIoTMQTTClient.h
wooj2: There was a lot of discussion years ago that it's generally a good rule of thumb to use copy for NSString.
https://stackoverflow.com/questions/387959/nsstring-property-copy-or-retain

I was trying to reduce the blast radius of my change, but really, once a user sets a value, they should not be able to modify it.  I will update userMetaData to use `copy`

Issue #2650, opened 2020-05-26
Line of code after return statement, never executed


OceanBuzz:
https://github.com/aws-amplify/aws-sdk-ios/blob/71a3a35b4622452d4a25418596a6151b31a791f2/AWSPinpoint/AWSPinpointSessionClient.m#L296

These two lines are after the return statement, and so are never executed.

palpatim:
Thanks for the report @OceanBuzz. We've released the fix for this on https://github.com/aws-amplify/aws-sdk-ios/releases/tag/2.15.0

PR #2858, opened 2020-07-16
fix(pinpoint): remove global event source attributes on session end

lawmicha:
*Issue #, if available:*
https://github.com/aws-amplify/aws-sdk-ios/issues/2650

bug was introduced in [2.6.34](https://github.com/aws-amplify/aws-sdk-ios/commit/aeb130753243778f373add44e2fa68ef42958064)

*Description of changes:*
The global source attributes were not being removed on session end. When handling notifications, `addGlobalEventSourceMetadata` is called on `open` and `background` push types. This means that a session event will be associated with the related campaign/journey data retrieved from the global event attributes. By not removing the global attributes, the new session will have the attributes additionally associated with the new session resulting in multiple  sessions associated with the same campaign/journey.

*Testing done:*
Manual test with the previous code: trigger a notification through pinpoint campaign, have the app in the background to receive it, tap on the notification to bring app to foreground, ensure that the global attributes are added. bring the app to the background. open the app to start a new session and check that the new session has the campaign associated with it.

### DIFF
--- a/AWSPinpoint/AWSPinpointSessionClient.m
+++ b/AWSPinpoint/AWSPinpointSessionClient.m
@@ -289,11 +289,11 @@ NSObject *sessionLock;
         _session = nil;
     }
 
-    return [self.context.analyticsClient recordEvent:stopEvent];
-
     //Remove global event source attributes
     AWSDDLogVerbose(@"Removed global event source attributes");
     [self.context.analyticsClient removeAllGlobalEventSourceAttributes];
+
+    return [self.context.analyticsClient recordEvent:stopEvent];
 }
 
 - (void) endCurrentSessionWithBlock:(AWSPinpointTimeoutBlock) block {


palpatim:
This is a behavior change right? Since global attributes are currently usable by customers outside of a journey, there may be customers who are relying on the fact that global attributes do not disappear when a session ends. Am I correct about this, or is the usage different from what I'm describing?￼￼

palpatim:
Although now that I look at this more closely, I think this was actually added relatively recently, which means that it's not something a customer would actually be relying on. Let's evaluate this, and make sure that it's not going to have a customer impact before we merge it. Otherwise, the change looks fine.￼

lawmicha:
> This is a behavior change right? Since global attributes are currently usable by customers outside of a journey, there may be customers who are relying on the fact that global attributes do not disappear when a session ends. Am I correct about this, or is the usage different from what I'm describing?￼￼

It is safe to add this back to remove "global event source attributes" because it only contains attributes from journeys/campaigns. the customer's global attributes are stored in another variable `globalAttributes` [here](https://github.com/aws-amplify/aws-sdk-ios/blob/main/AWSPinpoint/AWSPinpointAnalyticsClient.m#L247)


> Although now that I look at this more closely, I think this was actually added relatively recently, which means that it's not something a customer would actually be relying on. Let's evaluate this, and make sure that it's not going to have a customer impact before we merge it. Otherwise, the change looks fine.￼

it was touched recently when journeys was being added (`globalCampaignAttributes ` renamed to `globalEventSourceAttributes`) but the bug already existed back in [2.6.34](https://github.com/aws-amplify/aws-sdk-ios/commit/aeb130753243778f373add44e2fa68ef42958064)

PR #2861, opened 2020-07-17
AWSConnect Update models to latest

awsmobilesdk:
AWSConnect Update models to latest

### DIFF
--- a/AWSConnect/AWSConnectModel.h
+++ b/AWSConnect/AWSConnectModel.h
@@ -385,6 +385,13 @@ typedef NS_ENUM(NSInteger, AWSConnectUnit) {
     AWSConnectUnitPercent,
 };
 
+typedef NS_ENUM(NSInteger, AWSConnectVoiceRecordingTrack) {
+    AWSConnectVoiceRecordingTrackUnknown,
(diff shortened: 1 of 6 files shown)


AWSConnect/AWSConnectModel.h
```diff
@@ -1799,6 +1846,42 @@ typedef NS_ENUM(NSInteger, AWSConnectUnit) {
...
+ */
+@property (nonatomic, strong) AWSConnectVoiceRecordingConfiguration * _Nullable voiceRecordingConfiguration;
+
+@end
+
+/**
+ 
+ */
+@interface AWSConnectStartContactRecordingResponse : AWSModel
+
```
royjit, line 1883: Why is this empty?

palpatim: Per [the docs](https://docs.aws.amazon.com/connect/latest/APIReference/API_StartContactRecording.html):

> If the action is successful, the service sends back an HTTP 200 response with an empty HTTP body.

So it's expected that this is empty

palpatim:
Looks like this was already merged as part of 2.15.0. Closing PR

PR #2868, opened 2020-07-22
Replace block type alias borrowed from AVAudioSession with block sign…

ghost:
…ature. Removed in iOS 14.

*Issue #, if available:*
In iOS 14 the block type alias in `AVAudioSession.requestRecordPermission` was removed and replaced with the block signature (see: https://developer.apple.com/documentation/avfoundation/avaudiosession). This change causes the build to fail under iOS 14 beta 1 and beta 2. 

*Description of changes:*
I've replaced the removed block type alias with it's signature.

### DIFF
--- a/AWSLex/AWSLexInteractionKit.h
+++ b/AWSLex/AWSLexInteractionKit.h
@@ -516,7 +516,7 @@ typedef NS_ENUM(NSInteger, AWSLexSpeechEncoding) {
 /**
  Request record permission to AVAudioSession.
  */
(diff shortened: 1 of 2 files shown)


palpatim:
@FMJobr Thanks for the PR, we'll evaluate this for an upcoming release.

palpatim:
Wish I understood why the `PermissionBlock` typealias was removed--the docs don't seem to indicate. In any case, merging this in.

PR #2875, opened 2020-07-24
test(IoT): Integration test for enhanced custom auth

wooj2:
IoT integration test which targets use case for enhanced auth

### DIFF
--- a/AWSIoTTests/AWSIoTDataManagerTests.swift
+++ b/AWSIoTTests/AWSIoTDataManagerTests.swift
@@ -21,6 +21,7 @@ class AWSIoTDataManagerTests: XCTestCase {
     static let certificateSigningRequestCountryName = "US"
     static let certificateSigningRequestOrganizationName = "Amazon.com"
     static let certificateSigningRequestOrganizationalUnitName = "Amazon Web Services"
+    static let iotDataManagerBrokerCustomAuthEnhanced = "iot-data-manager-broker-custom-auth-enhanced"
 
     var policyName: String? = nil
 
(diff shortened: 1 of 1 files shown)


AWSIoTTests/AWSIoTDataManagerTests.swift
```diff
@@ -21,6 +21,8 @@ class AWSIoTDataManagerTests: XCTestCase {
     static let certificateSigningRequestCountryName = "US"
     static let certificateSigningRequestOrganizationName = "Amazon.com"
     static let certificateSigningRequestOrganizationalUnitName = "Amazon Web Services"
+    static let mqttTestUserName = "aws-sdk-user"
+    static let mqttTestPassword = "%%aws-sdk-password**"
```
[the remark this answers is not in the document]
wooj2: updated.

AWSIoTTests/AWSIoTDataManagerTests.swift
```diff
@@ -150,6 +184,19 @@ class AWSIoTDataManagerTests: XCTestCase {
...
         tokenSignature = packageConfig["custom_authorizer_token_signature"]
         tokenValue = packageConfig["custom_authorizer_token_value"]
+        
+        customAuthorizerName_customAuthUserPass = packageConfig["custom_authorizer_user_pass_name"]
+        tokenKeyName_customAuthUserPass = packageConfig["custom_authorizer_user_pass_token_key_name"]
+        tokenSignature_customAuthUserPass = packageConfig["custom_authorizer_user_pass_token_signature"]
+        tokenValue_customAuthUserPass = packageConfig["custom_authorizer_user_pass_token_value"]
+
+        let custom_authorizer_user_pass_username = packageConfig["custom_authorizer_user_pass_username"]
+        guard let entireUserName = custom_authorizer_user_pass_username else {
```
palpatim, line 194: OOC, why do we pass the entire username in the test config, rather than the username string? Now the test has to know how to split on question mark, which technically shouldn't be something the test code is doing, since we should be exercising (as much as practical) the public API

wooj2: yea.. i was on the fence about this... thanks for calling this out.  I've updated the CDK code to just use "aws-sdk-user" as the user name.

PR #2896, opened 2020-07-30
AWSSES Update models to latest

awsmobilesdk:
AWSSES Update models to latest


palpatim:
This is a massively breaking change to the existing SES SDK model. Will discuss with team to see how we want to address.

palpatim:
Looks like this was generated against the SES V2 model, rather than the V1 model.

PR #2900, opened 2020-07-31
bug(UICKeychainStore): serialize access to keychain on setData

wooj2:
OSStatus error bug reported here:
https://github.com/aws-amplify/amplify-ios/issues/649

We looked into the above issue and noticed that there were two processes attempting to store the same key into the keychain.  This PR attempts to serialize calls when calling `setData:` into the keychain.


mickeysox:
Hi all,

Thank you so much for getting this in so quickly.  I am on the latest everything and I can confirm this error has been fixed.

Thanks again.

Issue #1490, opened 2019-04-22
Crash in AWSSRWebSocket.m line 773


docee:
**Describe the bug**

Fabric shows more crashes here, but I can't reproduce the crash at this time.

**To Reproduce**

Can not reproduce

**Which AWS service(s) are affected?**

IoT

**Expected behavior**
Do not crash

**Environment(please complete the following information):**
 - SDK Version: 2.9.5
 - Dependency Manager: Cocoapods
 - Swift Version : Objc

**Device Information (please complete the following information):**
 - Device: iPhone7, iPhone 7P, iPhone 8, etc
 - iOS Version: iOS 12
 - Specific to simulators:

**Additional context**
[com.netviewtech.mynetvue_issue_crash_7594c2f694154697ab846b0077db3092_DNE_0_v2.txt](https://github.com/aws-amplify/aws-sdk-ios/files/3101815/com.netviewtech.mynetvue_issue_crash_7594c2f694154697ab846b0077db3092_DNE_0_v2.txt)
[com.netviewtech.mynetvue_issue_crash_2b62ed61449a459183239990f180fe62_DNE_0_v2.txt](https://github.com/aws-amplify/aws-sdk-ios/files/3101816/com.netviewtech.mynetvue_issue_crash_2b62ed61449a459183239990f180fe62_DNE_0_v2.txt)
[com.netviewtech.mynetvue_issue_crash_6f1a203829c4440cb50c7b39a7c710e8_DNE_0_v2.txt](https://github.com/aws-amplify/aws-sdk-ios/files/3101817/com.netviewtech.mynetvue_issue_crash_6f1a203829c4440cb50c7b39a7c710e8_DNE_0_v2.txt)

palpatim:
@docee 

Thank you for the report and the stack traces. We know that we're As we investigate this, can you confirm whether you're seeing this crash on the most recent version of the SDK? We did an IoT bug fix in 2.9.4 that may impact some instabilities during the connection release/reconnect phase.

That commit didn't touch SocketRocket in particular, so it may be unrelated, but we'd still like to confirm the behavior.

docee:
@palpatim 

I am very sorry, I wrote the wrong SDK version, it should be 2.9.5.

When our device list changes, we need to update the STS. After updating the STS, we need to reconnect the IoT Client (not IoT Data, so we need to reconnect) to control the new device with the new STS permission.

[...]

Issue #1434, opened 2019-04-09
Crashes in AWSSRWebSocket.m


Jonge:
**Describe the bug**
Seeing a couple crash logs from AWSSRWebSocket.m.

**To Reproduce**
Can't seem to reproduce, as it only happens to a small amount of users (probably a threading issue?). Here are the crash stacks, hope that helps a bit.

```
Crashed: Thread EXC_BAD_ACCESS KERN_INVALID_ADDRESS 0x0000000000000054
--
0 | libdispatch.dylib | dispatch_async$VARIANT$mp + 160
1 | AWSIoT | AWSSRWebSocket.m line 1555__37-[AWSSRWebSocket stream:handleEvent:]_block_invoke_3
2 | libdispatch.dylib | _dispatch_call_block_and_release + 24
3 | libdispatch.dylib | _dispatch_client_callout + 16
4 | libdispatch.dylib | _dispatch_lane_serial_drain$VARIANT$mp + 592
5 | libdispatch.dylib | _dispatch_lane_invoke$VARIANT$mp + 428
6 | libdispatch.dylib | _dispatch_workloop_worker_thread + 596
7 | libsystem_pthread.dylib | _pthread_wqthread + 304
8 | libsystem_pthread.dylib | start_wqthread + 4
```

```
Crashed: Thread SIGABRT ABORT 0x00000001bc3d00bc
--
0 | libsystem_kernel.dylib | __pthread_kill + 8
1 | libsystem_pthread.dylib | pthread_kill + 300
2 | (Missing) | (Missing)
3 | (Missing) | (Missing)
4 | (Missing) | (Missing)
5 | AWSIoT | AWSSRWebSocket.m line 1378-[AWSSRWebSocket _pumpScanner]
6 | AWSIoT | AWSSRWebSocket.m line 1551__37-[AWSSRWebSocket stream:handleEvent:]_block_invoke_3
7 | libdispatch.dylib | _dispatch_call_block_and_release + 32
```

**Which AWS service(s) are affected?**
AWS IoT

**Expected behavior**
App should not crash.

**Environment(please complete the following information):**
 - SDK Version: 2.9.4
 - Dependency Manager: Cocoapods
 - Swift Version: 4.2

**Device Information (please complete the following information):**
 - Device: Various devices (iPhone SE, iPhone X, iPad Pro...)
 - iOS Version:  iOS 12.2 and 12.3
 - Specific to simulators: no

wooj2:
[...]

PR #2921, opened 2020-08-07
Fix overcommit crash in AWSSRWebSocket

Jonge:
This should fix various crashes related to threading issues in `AWSSRWebSocket`.

The issue seems to be that it's calling `dispatch_async` to `_workQueue`, but we already are in `_workQueue` at that time.

We have released a build with this fix and are no longer seeing any crashes in `AWSSRWebSocket`, which were relatively frequent before.

Should fix the following issues:
https://github.com/aws-amplify/aws-sdk-ios/issues/1434
https://github.com/aws-amplify/aws-sdk-ios/issues/1490

### DIFF
--- a/AWSIoT/Internal/SocketRocket/AWSSRWebSocket.m
+++ b/AWSIoT/Internal/SocketRocket/AWSSRWebSocket.m
@@ -1552,22 +1552,20 @@ static const size_t SRFrameHeaderOverhead = 32;
                 if (aStream.streamError) {
                     [self _failWithError:aStream.streamError];
                 } else {
-                    dispatch_async(self->_workQueue, ^{
-                        if (self.readyState != AWSSR_CLOSED) {
-                            self.readyState = AWSSR_CLOSED;
-                            self->_selfRetain = nil;
-                        }
+                    if (self.readyState != AWSSR_CLOSED) {
+                        self.readyState = AWSSR_CLOSED;
+                        self->_selfRetain = nil;
+                    }
 
-                        if (!self->_sentClose && !self->_failed) {
-                            self->_sentClose = YES;
-                            // If we get closed in this state it's probably not clean because we should be sending this when we send messages
-                            [self _performDelegateBlock:^{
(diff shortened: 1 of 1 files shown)


palpatim:
@Jonge Thanks for the PR! As you probably know, these files are a fork of the old https://github.com/facebookarchive/SocketRocket project, so we really appreciate you diving into this to investigate.

I'm asking @wooj2 to take a look at this, as he's been spending quite a bit of time in the IoT stack and may have some additional context on this, but it's promising that you've got some real-world data to support this fix.

wooj2:
Hi @Jonge !

Thank you so much for taking the time to propose a fix for these crashes.

1.  Looking at issue #1434 w/ the crash on line 1555, I understand your PR with wanting to remove the call to `dispatch_async`, but I am not sure how this applies to the second crash which involves line 1378 with `_pumpScanner`.

2.  Looking at the issue in #1490, it seems to indicate that there is a crash dealing with `_failWithError`, but your proposed code change does not seem to touch `_failWithError`.  I'm a little bit confused as to why you believe that your fix will also fix this issue.

Taking a step back, I spent some time to look at our version of SocketRocket, your proposed PR, and the differences in the master branch of SocketRocket.  I noticed this PR (https://github.com/facebookarchive/SocketRocket/pull/169) which was merged into the main SocketRocket repo about a month after the first version of AWSSRWebSocket.m showed up in our repository.  I suspect that this PR directly addressed the crashes you are seeing.

The following is my best effort in attempting to back port this fix into our codebase.  I've done some basic testing on my side, but would more confident if you were able to test this change out in your application.  Would you be able to help us out?

https://github.com/aws-amplify/aws-sdk-ios/pull/2984

Jonge:
Thank you! I checked and am actually still seeing crashes in `_failWithError`, so sorry about the incorrect info.

I will add your fix into our next public beta, though it may take a couple days time before it's released and then properly tested.

wooj2:
Hi @Jonge 

Thank you so much for proposing this change.  It helped focus our eyes on this area of the code and resulted in a fix that was released in 2.16.0-- can't thank you enough for that!  For this specific change in this PR, we are not moving forward and therefore closing this PR.

Thanks again!

PR #2928, opened 2020-08-11
Fix updated claims that contain special characters, such as chinese c…

rolisanchez:
…hars, by replacing _ and - with / and +

*Issue #, if available:*

*Description of changes:*

The current code to get claims from a JWT Token works well **_as long as there are no special characters_** in encoded in the base64.

By trying to decode claims that contain a name in Mandarin, the variable `claimsData` will be nil, and we are unable to obtain the JSON dictionary.

In order to fix the padded claims, the fix is to replace _ and - (which are invalid characters in a base64 string) with / and +, respectively. (See example [here](https://stackoverflow.com/questions/1228701/code-for-decoding-encoding-a-modified-base64-url))

The fix is the following lines in [aws-sdk-ios/AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClientExtensions.swift](https://github.com/aws-amplify/aws-sdk-ios/blob/main/AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClientExtensions.swift)
```
var updatedClaims = claims.padding(toLength: paddedLength, withPad: "=", startingAt: 0)

// Replace _ with /
// Replace - with +
updatedClaims = updatedClaims.replacingOccurrences(of: "_", with: "/")
updatedClaims = updatedClaims.replacingOccurrences(of: "-", with: "+")
```

### DIFF
--- a/AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClientExtensions.swift
+++ b/AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClientExtensions.swift
@@ -36,25 +36,25 @@ public struct SessionToken {
                 return nil
             }
             let claims = tokenSplit[1]
+                .replacingOccurrences(of: "-", with: "+")
+                .replacingOccurrences(of: "_", with: "/")
             
             let paddedLength = claims.count + (4 - (claims.count % 4)) % 4
             //JWT is not padded with =, pad it if necessary
-            var updatedClaims = claims.padding(toLength: paddedLength, withPad: "=", startingAt: 0)
-            updatedClaims = updatedClaims.replacingOccurrences(of: "_", with: "/")
-            updatedClaims = updatedClaims.replacingOccurrences(of: "-", with: "+")
+            let updatedClaims = claims.padding(toLength: paddedLength, withPad: "=", startingAt: 0)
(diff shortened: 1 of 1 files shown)


wooj2:
Thank you so much for doing this.  We actually have a duplicate copy of this function in our amplify code, located here, which has the of the algorithm you are suggesting:

https://github.com/aws-amplify/amplify-ios/blob/732f7e72afbd3b4451ddd0a1c5ea426c40de6572/AmplifyPlugins/Core/AWSPluginsCore/Auth/AWSAuthService.swift#L55

Could you update your PR to match the implementation that we have in our Amplify code base?  Doing so, would allow the code to avoid setting updatedClaims to a `var`, as well as make the two copies of the code exactly the same.

rolisanchez:
CodeSign failed on the CircleCI unit tests. Not sure how to solve this.

```
The following build commands failed:
	PrecompileSwiftBridgingHeader normal x86_64
	CodeSign /Users/distiller/Library/Developer/Xcode/DerivedData/AWSiOSSDKv2-bvcffhrgbfjnfybjecennhscmaho/Build/Products/Debug-iphonesimulator/AWSAllTestsHost.app
(2 failures)
```

wooj2:
I re-ran the unit tests to overcome that error.  Sorry about that!

wooj2 (approved):
LGTM!

Issue #2939, opened 2020-08-13
Bad Networks Cause Continuous Failure in MQTT Connection


realityworks:
**Describe the bug**
Our app, after going into the background, waiting for 10-15 minutes, then reopening, will normally reconnect to the MQTT socket correctly. If however, the network connection is bad then the MQTT service fails to connect and continues to fail indefinitely.

We have seen this on local Wifi and LTE wireless broadband networks. This issue can easily be reproduced by using the Link Conditioner plugin and setting it to Very Bad Network. 

**To Reproduce**
1. Open App with MQTT Service
2. Connect MQTT Service successfully
3. Put app in the Background
4. Wait 10-20 minutes
5. Reopen the app. Please see the resulting output. App takes an extremely long time to reconnect.

**Observed Behavior**
App sits waiting to connect to the MQTT service, leaving the loading indicator on indefinitely.

**Expected Behavior**
Expect the MQTT service to be connected

**Stack Trace**
N/A

**Code Snippet**
N/A

**Unique Configuration**
N/A

**Areas of the SDK you are using (AWSMobileClient, Cognito, Pinpoint, IoT, etc)?**
AWSIoT

**Screenshots**
N/A

**Environment(please complete the following information):**
 - SDK Version: iOS 13.5
 - Dependency Manager: CocoaPods
 - Swift Version : 5
 - Xcode Version: 11.6

**Device Information (please complete the following information):**
 - Device: All Devices (Including Simulator)
 - iOS Version: iOS 13
 - Specific to simulators:

**Additional context**
Use Link Conditioner to reproduce (Set to Very Bad Network)

**Relevant Console Output**
Detailed Log of Observed Behaviour 
```
2020-08-13 19:41:20:135 PCA[93170:1571905] Initializing AWSIoTMqttConfiguration with KeepAlive:300.000000, baseReconnectTime:1.000000,minimumConnectionTime:5.000000, maximumReconnectTime:16.000000, autoResubscribe:Enabled, lwt topic: message: 
2020-08-13 19:41:20:572 PCA[93170:1572364] IOTDataManager: Connecting to IoT using websocket, client id: 91E18881-B258-41B4-AA96-D727BB467941
2020-08-13 19:41:20:572 PCA[93170:1572364] AWSIoTMQTTClient: connecting via websocket. 
2020-08-13 19:41:20:572 PCA[93170:1572364] Using AWSCredentials.
[...]

PR #2949, opened 2020-08-17
chore(IoT): Stop multiple threads from entering CR while connecting

wooj2:
In attempts to address one of the issues reported here:
https://github.com/aws-amplify/aws-sdk-ios/issues/2939

It seems that there are competing threads attempting to connect to the websocket.


wooj2:
We are not considering this as a "fix", because it is possible that mqttStatus can be updated while entering the critical region (CR), but we think this approach will reduce the chance of a second thread disconnecting a connecting websocket from the first thread.  We consider this an incremental change (hence the name change from `fix` -> `chore`) towards a path of trying to work towards a solid solution.

PR #2955, opened 2020-08-18
AWSSES Update models to latest

awsmobilesdk:
AWSSES Update models to latest

New API service model update


palpatim:
Breaking change, need to investigate

Issue #2770, opened 2020-06-19
Crash on [AWSIoTMQTTClient webSocketDidOpen:] + 1173


florentdouine:
**Describe the bug**
Crashlytics shows up some crashs in the AWS SDK. The most common one is a crash on "webSocketDidOpen" function. We didn't identify a clear way to reproduce this crash but it seems to appear more often when returning to app from background.

This crash has already been details in this [issue](https://github.com/aws-amplify/aws-sdk-ios/issues/2313) 

**Stack Trace**
AWSIoTMQTTClient.m – line 1173
-[AWSIoTMQTTClient webSocketDidOpen:]

Crashed: com.apple.root.default-qos
EXC_BAD_ACCESS KERN_INVALID_ADDRESS 0x0000000a6586a5d0

0  libobjc.A.dylib                0x1b6817c70 objc_retain + 16
1  AWSIoT                         0x107522ab8 -[AWSIoTMQTTClient webSocketDidOpen:] + 1173 (AWSIoTMQTTClient.m:1173)
2  AWSIoT                         0x107586dd8 __39-[AWSSRWebSocket _HTTPHeadersDidFinish]_block_invoke + 476 (AWSSRWebSocket.m:476)
3  libdispatch.dylib              0x1b6786ec4 _dispatch_call_block_and_release + 32
4  libdispatch.dylib              0x1b678833c _dispatch_client_callout + 20
5  libdispatch.dylib              0x1b67976e8 _dispatch_root_queue_drain + 644
6  libdispatch.dylib              0x1b6797d9c _dispatch_worker_thread2 + 116
7  libsystem_pthread.dylib        0x1b67ef6d8 _pthread_wqthread + 216
8  libsystem_pthread.dylib        0x1b67f59c8 start_wqthread + 8

**Environment(please complete the following information):**
 - SDK Version: 2.13.0
 - Dependency Manager: Carthage
 - Swift Version : 5.1
 - Xcode Version: 11.3

**Device Information (please complete the following information):**
 - All kind of devices

**Additional context**
The bug occured only on foreground

palpatim:
@florentdouine As you've noted, we've encountered this issue before but been unable to make any headway on it. In order for us to be able to help, we'll need repro steps, and possibly a code snippet showing your configuration and initialization.

However, in the meantime, please see the very recent update to that issue: https://github.com/aws-amplify/aws-sdk-ios/issues/2313#issuecomment-647947479, where @brendancallahan reports:

[...]

PR #2957, opened 2020-08-18
chore(IoT): Cleanup toDecoderStream delegate and runloop

wooj2:
I was unable to reproduce the crash in #2770, but in looking at the crash, it seems to be crashing on setting the delegate on `self.toDecoderStream`.  My theory is that the NSOutputStream is not getting cleaned up properly (when the websocket closes/fails? or maybe when a call to disconnect() happens), so, I'm attempting to clean it up properly in efforts of reduce the amount of crashes customers are seeing with regards to this line.

### DIFF
--- a/AWSIoT/Internal/AWSIoTMQTTClient.m
+++ b/AWSIoT/Internal/AWSIoTMQTTClient.m
@@ -645,6 +645,13 @@
     }
 }
 
+- (void)cleanUpToDecoderStream {
+    self.toDecoderStream.delegate = nil;
+    [self.toDecoderStream removeFromRunLoop:[NSRunLoop currentRunLoop] forMode:NSDefaultRunLoopMode];
+    [self.toDecoderStream close];
+    self.toDecoderStream = nil;
+}
+
 - (void)reconnectToSession {
     
     self.reconnectTimer = nil;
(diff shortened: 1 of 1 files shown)


AWSIoT/Internal/AWSIoTMQTTClient.m
```diff
@@ -767,9 +759,15 @@ - (void)openStreams:(id)sender
...
         }
         [self.session close];
-        
-        if ( self.webSocket) {
+
+        if (self.toDecoderStream != nil) {
+            self.toDecoderStream.delegate = nil;
+            [self.toDecoderStream removeFromRunLoop:runLoopForStreamsThread forMode:NSDefaultRunLoopMode];
             [self.toDecoderStream close];
+            self.toDecoderStream = nil;
```
palpatim, line 767: It looks like we reuse this in a few places. Maybe factor it into a private `cleanUpToDecoderStream` method or similar?

wooj2: I was thinking to do this, but ultra paranoid about putting anything on the stack of attempting to remove these items immediately.  But on a second opinion, this is probably too paranoid. Will update and submit.

PR #2969, opened 2020-08-25
copy with Zone method for copy AWTask instance

amjadkhan120:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/AWSCore/Bolts/AWSTask.m
+++ b/AWSCore/Bolts/AWSTask.m
@@ -83,6 +83,20 @@ NSString *const AWSTaskMultipleErrorsUserInfoKey = @"errors";
     return self;
 }
 
+- (id)copyWithZone:(NSZone *)zone {
+    AWSTask *task = [AWSTask allocWithZone:zone];
+    task.lock = self.lock;
+    task.condition = self.condition;
+    task.callbacks = self.callbacks;
+    
+    task.cancelled = self.cancelled;
+    task.completed = self.completed;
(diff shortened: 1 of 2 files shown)


AWSCore/Bolts/AWSTask.m
```diff
@@ -83,6 +83,20 @@ - (instancetype)initCancelled {
...
+- (id)copyWithZone:(NSZone *)zone {
+    AWSTask *task = [AWSTask allocWithZone:zone];
+    task.lock = self.lock;
+    task.condition = self.condition;
+    task.callbacks = self.callbacks;
+    
+    task.cancelled = self.cancelled;
+    task.completed = self.completed;
+    task.faulted   = self.faulted;
+    [task trySetError:_result];
```
palpatim, line 95: I think you meant `trySetResult` here?

palpatim:
@amjadkhan120 Thanks for the PR, however this is not something we want to merge. We don't have any use cases inside the SDK that require a `copyWithZone` method, and generally, we believe it's not good practice to provide a mutable copy of an AWSTask in the first place. Tasks are generally emitted from work blocks and are closely tied to their execution path.

It's not clear what use case you're accomplishing by adding this method, but be aware that you can create new tasks with the existing `+[AWSTask taskWithResult]` or `+[AWSTask taskWithError]` methods to start a new task chain. If your use case absolutely requires a `copyWithZone` implementation, consider adding the method to an Objective-C extension in your app.

PR #2976, opened 2020-08-27
AWSEC2 Update models to latest

awsmobilesdk:
AWSEC2 Update models to latest


palpatim:
The `AWSEC2CapacityReservationInstancePlatformWindowsBYOL` was added in error on a previous release; this breaking change is OK to commit, although we should note it in the CHANGELOG.

PR #2984, opened 2020-08-28
fix(IoT): Backport crash fix into AWSSRWebSocket

wooj2:
Attempting to backport crash fix from:
https://github.com/facebookarchive/SocketRocket/pull/169

### DIFF
--- a/AWSIoT/Internal/SocketRocket/AWSSRWebSocket.m
+++ b/AWSIoT/Internal/SocketRocket/AWSSRWebSocket.m
@@ -242,6 +242,7 @@ typedef void (^data_callback)(AWSSRWebSocket *webSocket,  NSData *data);
 
     BOOL _sentClose;
     BOOL _didFail;
+    BOOL _cleanupScheduled;
     int _closeCode;
     
     BOOL _isPumping;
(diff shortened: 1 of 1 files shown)


AWSIoT/Internal/SocketRocket/AWSSRWebSocket.m
```diff
@@ -1230,6 +1232,40 @@ - (void)_addConsumerWithScanner:(stream_scanner)consumer callback:(data_callback
...
+{
+    @synchronized(self) {
+        // Nuke NSStream delegate's
+        _inputStream.delegate = nil;
+        _outputStream.delegate = nil;
+
+        // Remove the streams, right now, from the networkRunLoop
+        [_inputStream close];
+        [_outputStream close];
+    }
```
palpatim, line 1261: As noted in the review comment, I think we need to call `[timer invalidate]` in this `@synchronized` block

wooj2: Great call out.  I read through the documentation here:
https://developer.apple.com/documentation/foundation/nstimer

> Once scheduled on a run loop, the timer fires at the specified interval until it is invalidated. A nonrepeating timer invalidates itself immediately after it fires. However, for a repeating timer, you must invalidate the timer object yourself by calling its invalidate method.


Since this is a non-repeating timer, I believe that calling `invalidate` is not needed, because we are passing in `.. repeats:NO`:
```
 NSTimer *timer = [NSTimer timerWithTimeInterval:(0.0f) target:self selector:@selector(_cleanupSelfReference:) userInfo:nil repeats:NO];
```

palpatim: Thanks, you're absolutely right.

PR #3006, opened 2020-09-08
fix(core): prevent overwriting serviceName in Endpoint

phani-srikar:
*Description of changes:*
The `serviceName` property of `AWSEndpoint` could already be set in the `AWSServiceConfiguration` (using these [initializers](https://github.com/aws-amplify/aws-sdk-ios/blob/24ecaabd4f182f44bce00eef63ca600c1ea4aac1/AWSCore/Service/AWSService.m#L376)). If we look at any code generated service client for example, [this](https://github.com/aws-amplify/aws-sdk-ios/blob/24ecaabd4f182f44bce00eef63ca600c1ea4aac1/AWSAutoScaling/AWSAutoScalingService.m#L232) call to `setRegion:(AWSRegionType)regionType service:(AWSServiceType)serviceType`, seems to not only set the region, but also override the `serviceName`. 

### DIFF
--- a/AWSCore/Service/AWSService.m
+++ b/AWSCore/Service/AWSService.m
@@ -426,7 +426,10 @@ static NSString *const AWSServiceNameTranslate = @"translate";
     _regionType = regionType;
     _serviceType = serviceType;
     _regionName = [AWSEndpoint regionNameFromType:regionType];
-    _serviceName = [self serviceNameFromType:serviceType];
+    // set the `serviceName` only if it is not already set
+    if (!_serviceName.length){
+        _serviceName = [self serviceNameFromType:serviceType];
+    }
 }
 
 + (NSString *)regionNameFromType:(AWSRegionType)regionType {


AWSCore/Service/AWSService.m
```diff
@@ -426,7 +426,10 @@ - (void) setRegion:(AWSRegionType)regionType service:(AWSServiceType)serviceType
     _regionType = regionType;
     _serviceType = serviceType;
     _regionName = [AWSEndpoint regionNameFromType:regionType];
-    _serviceName = [self serviceNameFromType:serviceType];
```
kneekey23, line 429: is there a reason it was being overwritten here?

phani-srikar: `serviceName` is used in several places like [signing requests](https://github.com/aws-amplify/aws-sdk-ios/blob/24ecaabd4f182f44bce00eef63ca600c1ea4aac1/AWSCore/Authentication/AWSSignature.m#L203), [creating the host url](https://github.com/aws-amplify/aws-sdk-ios/blob/24ecaabd4f182f44bce00eef63ca600c1ea4aac1/AWSCore/Service/AWSService.m#L655) etc. I believe the `serviceName` is expected to be a constant for a service. Hence, we are always setting it to the same value defined  [here](https://github.com/aws-amplify/aws-sdk-ios/blob/24ecaabd4f182f44bce00eef63ca600c1ea4aac1/AWSCore/Service/AWSService.m#L262)

royjit: The method signature is `setRegion:(AWSRegionType)regionType service:(AWSServiceType)serviceType` ie, it is changing the region and service type. Since serviceType is changed, the service name should also be updated right? ie the value written in initializer should be overriden.

phani-srikar:
As discussed with the concerned team, we will not need this breaking change and this use-case can be addressed using [this](https://github.com/aws-amplify/aws-sdk-ios/blob/24ecaabd4f182f44bce00eef63ca600c1ea4aac1/AWSCore/Service/AWSService.m#L376) and writing their own client initialization logic.

Issue #2914, opened 2020-08-05
Secure code by using NSSecureCoding instead of NSCoding


mmathewsTableau:
NSCoding does not verify the type of object upon deserialization and therefore is vulnerable to object substitution attacks. The code can protect itself from these attacks by using NSSecureCoding instead: https://developer.apple.com/documentation/foundation/nssecurecoding?language=objc 

Various spots where NSCoding is used:
https://github.com/aws-amplify/aws-sdk-ios/search?q=NSCoding&unscoped_q=NSCoding

palpatim:
Thank you for rep​​orting ​​your concern, and for your patience as we investigated. The AWS Mobile SDK’s use of NSCoding does not represent a security issue. However, NSSecureCoding introduces improvements, so we are currently replacing the SDK’s use of NSCoding with NSSecureCoding. We’ll update this issue as we make progress on those changes.

mmathewsTableau:
Thanks @palpatim for looking into. It sounds like you have gathered enough information to conclude that the specific usages are not an issue. 

It's a conclusion that I wanted to try to get to myself but I couldn't without enough context about how the code works. To help alleviate my concerns, would you be able to explain why the usage in AWS SDK are not a security issue? Thanks

palpatim:
All subclasses of `AWSMTLModel` now support `NSSecureCoding`, with a few exceptions:

  - AWS IoT:
    - `AWSIoTDataDeleteThingShadowResponse`
    - `AWSIoTDataGetThingShadowResponse`
    - `AWSIoTDataPublishRequest`
    - `AWSIoTDataUpdateThingShadowRequest`
    - `AWSIoTDataUpdateThingShadowResponse`
  - Amazon Lambda:
    - `AWSLambdaInvocationRequest`
    - `AWSLambdaInvocationResponse`
  - Amazon S3:
    - `AWSS3GetObjectOutput`
    - `AWSS3GetObjectTorrentOutput`
    - `AWSS3PutObjectRequest`
    - `AWSS3RecordsEvent`
    - `AWSS3UploadPartRequest`

    To support `NSSecureCoding` for those types, host apps will have to create a subclass of the base type, and override the appropriate `initWithCoder:` methods to provide a type-safe unarchiving method.

[...]

PR #3031, opened 2020-09-21
feat(Pinpoint): Support NSSecureCoding

palpatim:
*Issue #, if available:*

https://github.com/aws-amplify/aws-sdk-ios/issues/2914

*Description of changes:*

- Patches AWSPinpoint high-level client, and the fork of JKBigInteger, to support NSSecureCoding
- Introduces an NSCodingUtilities helper class for version-safe `NSKeyedArchiver` and `NSKeyedUnarchiver` functions that default to NSSecureCoding if available

### DIFF
--- a/AWSCore/Utility/AWSNSCodingUtilities.m
+++ b/AWSCore/Utility/AWSNSCodingUtilities.m
@@ -1,0 +1,108 @@
+//
+// Copyright 2010-2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License").
+// You may not use this file except in compliance with the License.
+// A copy of the License is located at
+//
+// http://aws.amazon.com/apache2.0
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+//
+
+#import "AWSCocoaLumberjack.h"
+#import "AWSNSCodingUtilities.h"
+
+@implementation AWSNSCodingUtilities
+
+#pragma mark - Archive
+
++ (nullable NSData *)versionSafeArchivedDataWithRootObject:(id)obj
+                                     requiringSecureCoding:(BOOL)requireSecureCoding
+                                                     error:(NSError *__autoreleasing *)error {
+    NSData *archivedData;
+    if (@available(iOS 11, *)) {
+        archivedData = [NSKeyedArchiver archivedDataWithRootObject:obj
+                                             requiringSecureCoding:requireSecureCoding
+                                                             error:error];
+        if (error && *error) {
+            AWSDDLogError(@"Error archiving object: %@", *error);
+            return nil;
+        }
+    } else {
+        archivedData = [NSKeyedArchiver archivedDataWithRootObject:obj];
+    }
+
+    return archivedData;
+}
+
+#pragma mark - Unarchive
+
++ (nullable id)versionSafeUnarchivedObjectOfClass:(Class)cls
+                                         fromData:(NSData *)data
+                                            error:(NSError *__autoreleasing *)error {
+    id returnValue;
+
+    if (@available(iOS 11, *)) {
+        returnValue = [NSKeyedUnarchiver unarchivedObjectOfClass:cls
+                                                        fromData:data
+                                                           error:error];
+        if (error && *error) {
+            AWSDDLogError(@"Error unarchiving class `%@`: %@", cls, *error);
+            return nil;
+        }
+    } else {
+        returnValue = [NSKeyedUnarchiver unarchiveObjectWithData:data];
+    }
+
+    return returnValue;
+}
+
++ (nullable id)versionSafeUnarchivedObjectOfClasses:(NSSet<Class> *)classes
+                                           fromData:(NSData *)data
+                                              error:(NSError *__autoreleasing *)error {
+    id returnValue;
+
+    if (@available(iOS 11, *)) {
+        returnValue = [NSKeyedUnarchiver unarchivedObjectOfClasses:classes
+                                                          fromData:data
+                                                             error:error];
+        if (error && *error) {
+            AWSDDLogError(@"Error unarchiving data into allowed classes `%@`: %@", classes, *error);
+            return nil;
+        }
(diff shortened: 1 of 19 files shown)


AWSCore/Utility/AWSNSCodingUtilities.m
```diff
@@ -0,0 +1,108 @@
...
+
+    return returnValue;
+}
+
++ (NSMutableDictionary *)versionSafeMutableDictionaryFromData:(NSData *)data
+                                                        error:(NSError *__autoreleasing *)error {
+    NSMutableDictionary *returnValue;
+
+    if (@available(iOS 11, *)) {
+        NSSet *allowableClasses = [[NSSet alloc] initWithObjects:[NSMutableString class],
```
royjit, line 89: Why do we have mutable string in the allowed classes? Is it the content of dictionary?

AWSPinpoint/AWSPinpointEndpointProfile.m
```diff
@@ -335,20 +336,23 @@ - (NSDictionary *) toDictionary {
...
+        _applicationId = [decoder decodeObjectOfClass:[NSString class] forKey:@"applicationId"];
+        _endpointId = [decoder decodeObjectOfClass:[NSString class] forKey:@"endpointId"];
+        _channelType = [decoder decodeObjectOfClass:[NSString class] forKey:@"channelType"];
+        _address = [decoder decodeObjectOfClass:[NSString class] forKey:@"address"];
+        _optOutBackingVariable = [decoder decodeObjectOfClass:[NSString class] forKey:@"optOut"];
         _effectiveDate = [decoder decodeInt64ForKey:@"effectiveDate"];
-        _optOutBackingVariable = [decoder decodeObjectForKey:@"optOut"];
+        _location = [decoder decodeObjectOfClass:[AWSPinpointEndpointProfileLocation class] forKey:@"location"];
+        _demographic = [decoder decodeObjectOfClass:[AWSPinpointEndpointProfileDemographic class] forKey:@"demographic"];
+        _attributes = [decoder decodeObjectOfClass:[NSDictionary class] forKey:@"attributes"];
```
royjit, line 353: Can we verify the contents of dictionary or array?

AWSPinpoint/AWSPinpointEventRecorder.m
```diff
@@ -198,10 +207,17 @@ - (NSString *)validateOrRetrieveSessionId:(NSString *) sessionId {
     }
     NSData *sessionData = [self.context.configuration.userDefaults dataForKey:AWSPinpointSessionKey];
     if (sessionData) {
-        AWSPinpointSession *previousSession = [NSKeyedUnarchiver unarchiveObjectWithData:sessionData];
+        NSError *error;
+        AWSPinpointSession *previousSession = [AWSNSCodingUtilities versionSafeUnarchivedObjectOfClass:[AWSPinpointSession class]
+                                                                                              fromData:sessionData
+                                                                                                 error:&error];
+        if (error || !previousSession) {
```
royjit, line 214: `!previousSession` is not an error in restoring the archive right? It can mean that we do not have an archived object. So, should we just have ` if (error) { ... }`

AWSPinpoint/AWSPinpointEventRecorder.m
```diff
@@ -1149,4 +1268,20 @@ - (AWSPinpointTargetingPutEventsRequest*) putEventsRequestForEvents:(NSDictionar
...
+                                             forColumnName:(NSString *)columnName
+                                                     error:(NSError *__autoreleasing *)error {
+    NSSet *allowableClasses = [[NSSet alloc] initWithObjects:[NSMutableString class],
+                               [NSDictionary class],
+                               nil];
+    NSDictionary *immutableDict = [AWSNSCodingUtilities versionSafeUnarchivedObjectOfClasses:allowableClasses
+                                                                                    fromData:[rs dataForColumn:columnName]
+                                                                                       error:error];
+    if (*error) {
+        return nil;
```
royjit, line 1281: Log this error?

AWSPinpoint/AWSPinpointEventRecorder.m
```diff
@@ -1149,4 +1268,20 @@ - (AWSPinpointTargetingPutEventsRequest*) putEventsRequestForEvents:(NSDictionar
...
+    }
+    NSMutableDictionary *mutableDict = [immutableDict mutableCopy];
+    return mutableDict;
```
royjit, line 1284: just `return [immutableDict mutableCopy];`

AWSCore/Utility/AWSNSCodingUtilities.m
palpatim: Yes, `NSMutableDictionary` serializes to an archive that includes `NSMutableString` in its classes.

AWSPinpoint/AWSPinpointEndpointProfile.m
palpatim: Not to my knowledge. `NSDictionary` (or its mutable equivalent) is the actual class that gets instantiated, regardless of the type of the keys & values. It might be that a complex type (e.g., an `NSDictionary` with objects as keys) would require additional class declarations, but I haven't been able to verify that experimentally.

AWSPinpoint/AWSPinpointEventRecorder.m
palpatim: I like declaring intervening variables to make it easier to debug.

AWSPinpoint/AWSPinpointEventRecorder.m
palpatim: If we didn't have an archived object, we shouldn't have been able to hit this code in the first place, since we check for the existence of `sessionData` on L209 above. Thus, if we have retrieved session data, but are unable to unarchive it, it implies the archived data was corrupted and we should follow the error path.

Practically speaking, it's no change in behavior, since you can see below that the very next thing we try is to return `previousSession.sessionId`, returning the default if it's not present.

AWSPinpoint/AWSPinpointEventRecorder.m
palpatim: Good callout, will do

royjit (approved):

AWSPinpoint/AWSPinpointEventRecorder.m
palpatim: As discussed offline, I will add this in a subsequent PR

Issue #3036, opened 2020-09-22
Undefined symbols for architecture x86_64:   "_OBJC_CLASS_$_AWSCognitoIdentityProviderASF"


vimitahirpara:
Getting error while using AWSMobileClient pod in application 

Showing All Messages
**Ld /Users/maya_somase/Library/Developer/Xcode/DerivedData/ClinicalWrapper-ghlyjxzrixwurjeruzoucyzdsrlu/Build/Products/Debug-iphonesimulator/AWSMobileClient/AWSMobileClient.framework/AWSMobileClient normal (in target 'AWSMobileClient' from project 'Pods')
    cd /Users/maya_somase/Downloads/dht-mobile-wrapper-TARDIS_Auth_API_Impl/ClinicalWrapperApp_iOS/Pods
[...]

Issue #2927, opened 2020-08-11
Build failure with 2.15.x in Xcode 12


corey-plugshare:
**Describe the bug**
Getting the following build failure:

/Users/user/Desktop/aws-test/Pods/AWSCognitoIdentityProviderASF/AWSCognitoIdentityProviderASF/Internal/libAWSCognitoIdentityProviderASFBinary.a(AWSCognitoIdentityASF.o), building for iOS Simulator, but linking in object file built for iOS, for architecture arm6

These errors also occur in 2.15.0 and 2.14.2, so this is not specific to 2.15.

**To Reproduce**

- In Xcode 12.0 Beta 4 (the current beta at the time of opening this issue) create a new project
- Initialize Cocoapods and install the following pods: AWSMobileClient, AWSAuthUI, AWSUserPoolsSignIn
- Open the project in Xcode using the .xcworkspace file
- Build

**Observed Behavior**
Build will fail with the following error:

/Users/user/Desktop/aws-test/Pods/AWSCognitoIdentityProviderASF/AWSCognitoIdentityProviderASF/Internal/libAWSCognitoIdentityProviderASFBinary.a(AWSCognitoIdentityASF.o), building for iOS Simulator, but linking in object file built for iOS, for architecture arm6

**Expected Behavior**
A successful build.

**Stack Trace**
There is no stack trace, but here is a snippet from the build log:

Ld /Users/user/Library/Developer/Xcode/DerivedData/aws-test-ghjmeyubqzmpsyevlhtmgwyoogpf/Build/Intermediates.noindex/Pods.build/Release-iphonesimulator/AWSCognitoIdentityProviderASF.build/Objects-normal/arm64/Binary/AWSCognitoIdentityProviderASF normal arm64 (in target 'AWSCognitoIdentityProviderASF' from project 'Pods')
    cd /Users/user/Desktop/aws-test/Pods
[...]

PR #3039, opened 2020-09-22
fix: Fix dependency declarations and linking for all targets

palpatim:
- Normalize deployment target & file inclusion for tests
- Remove hardcoded LD_RUNPATH_SEARCH_PATHS

*Issue #, if available:*

#2927
https://github.com/aws-amplify/aws-sdk-ios/issues/3036

**Description of changes:**

The previous project inconsistently linked frameworks in the Xcode project. SDK targets relied on "Dependency" build phases and `LD_RUNPATH_SEARCH_PATHS` build settings to track down and link build dependencies.

This change changes several build time settings:

- **Breaking change** AWSCognitoIdentityProviderASF no longer ships the prepackaged `libAWSCognitoIdentityProviderASFBinary.a` static library. AWSCognitoIdentityProviderASF is also now versioned the same as the rest of the SDK.
- Removes hardcoded `LD_RUNPATH_SEARCH_PATHS` from all targets. This appeared to be used prior to Xcode 12 to help resolve frameworks at linking time. However, our best information is that this attribute (which populates the linker's `rpath` argument), should be used to resolve paths at runtime, not build time. Apple's [guidance](https://developer.apple.com/videos/play/wwdc2018/415/) for resolving framework dependencies is to add them as explicit members of the "Dependency" build phase (which sets up an explicit dependency, but no linking), or as members of the "Link binary with libraries" build phase (which sets up an implicit dependency if that option is selected in the scheme editor, and also links as desired).
- For the most part, removes dependencies from the "Dependency" build phases in favor of adding them to the "Link binary with libraries" build phase. The latter is required to properly link referenced symbols; the former appears to be derivable from adding the "Link binaries" build phase. Some test targets retain a "Dependency" for `AWSAllTestsHost`, since that host must be built to act as a host target for the unit or integration test. In those targets, `AWSAllTestsHost` is *not* added to the "Link binary with libraries" build phase.
- For all test targets, links `AWSCore`, the SDK under test, and `AWSTestResources`. Some test targets may not actually use the `AWSTestResources` functionality, but it's innocuous to include it and makes for easier reference patterns for future SDK additions
- Removes unnecessary "Copy files" phases that inconsistently copied frameworks to build paths for unit tests only
- Builds Carthage and direct binaries with a script from https://github.com/Carthage/Carthage/issues/3019#issuecomment-665136323 . This script works around the error building Carthage binaries in Xcode 12

**Testing**

- [x] CircleCI builds, unit tests pass, Xcode 11
- [x] All unit tests pass, Xcode 11
- [x] All unit tests pass, Xcode 12
- [x] All integ tests build, Xcode 11
- [x] All integ tests build, Xcode 12
- [x] All SDK targets build locally in Xcode 11
- [x] All SDK targets build locally in Xcode 12
- [x] All SDK targets build via Carthage using new script, Xcode 11
- [x] All SDK targets build via Carthage using new script, Xcode 12
- [x] CocoaPods `pod lib lint`, Xcode 11
- [x] CocoaPods `pod lib lint`, Xcode 12
- [x] Amplify installs from pod, builds, Xcode 11
- [x] Amplify installs from pod, builds, Xcode 12
- Ensure sample apps consume & build as expected:
  - [x] Amplify, CocoaPods, Xcode 11 (@palpatim)
  - [x] Amplify, CocoaPods, Xcode 12 (@palpatim)
  - [x] SDK, CocoaPods, Xcode 11 (@wooj2)
  - [x] SDK, CocoaPods, Xcode 12 (@wooj2)
  - [x] SDK, Carthage, Xcode 11 (@royjit, and will build Carthage Xcode 11 archive for use in AppSync test below)
  - [x] SDK, Carthage, Xcode 12 (@royjit, and will build Carthage Xcode 11 archive for use in AppSync test below)

* From your SDK project workspace: `git fetch origin && git checkout -b palpatim/xcode-12-build-fix origin/palpatim/xcode-12-build-fix`
* Setup private spec repo for SDK - `./Scripts/setup_private_specs.sh 2.17.0 2.17.0`

* From your Amplify project workspace: `git fetch origin && git checkout -b palpatim/xcode-12-build origin/palpatim/xcode-12-build`
* Setup private spec repo for Amplify - `./Scripts/setup_private_specs.sh 1.2.1 1.2.1`

* Select Xcode: `sudo xcode-select --switch <path to /Applications/your Xcode version.app>`
* Verify Xcode selection: `xcode-select -p`

* Switch to your test app project workspace
* Sample project should have imports for all tested frameworks

* Carthage testing:
  - `./CircleciScripts/carthage-build.sh --platform iOS --no-skip-current --cache-builds | tee buildout.txt`
  - `rm -f CircleciScripts/.python-version && python3 CircleciScripts/create_carthage_archive.py && git checkout -- ./CircleciScripts/.python-version`
  - Copy the archive .zip file to your workstation, unzip it
  - Install required frameworks in your test project: "Link binary with libraries" build phase step
  - Quit Xcode
  - Clear derived data: `rm -rf $HOME/Library/Developer/Xcode/DerivedData`
  - Restart Xcode
  - Immediately build project before Indexing starts

* CocoaPods testing
  - Quit Xcode
  - Clear derived data: `rm -rf $HOME/Library/Developer/Xcode/DerivedData`
  - Clears all installed pods and cache: `find . -name Pods -o -name Podfile.lock | xargs rm -rf && pod cache clean --all`
  - Install pods: `root=$PWD ; find . -name Podfile | while read podPath ; do d=$( dirname $podPath ); cd $d && pod install --verbose && cd $root ; done`
    - Watch the output of the verbose pod install. You're looking to make sure that the pod installations are happening from your private spec repo, not trunk. E.g., you want to see a git clone like `/usr/bin/git clone file:///Users/...` not `/usr/bin/git clone https://github...`
  - Restart Xcode
  - Immediately build project before Indexing starts

---


.circleci/config.yml
```diff
@@ -467,7 +467,8 @@ jobs:
       - run:
           name: Build Carthage
           command: |
-            carthage build --no-skip-current | tee buildout.txt
+            # Build Carthage via this script to workaround lipo error due to multiple architectures in same file
+            . ./CircleciScripts/carthage-build.sh --no-skip-current --platform iOS | tee buildout.txt
```
wooj2, line 471: I'm probably wrong, but this seems like this could just be:

```
./CircleciScripts/carthage-build.sh --no-skip-current --platform iOS | tee buildout.txt
```
??

AWSCognitoIdentityProviderASF.podspec
```diff
@@ -10,10 +10,9 @@ Pod::Spec.new do |s|
   s.author       = { 'Amazon Web Services' => 'amazonwebservices' }
   s.platform     = :ios, '9.0'
   s.source       = { :git => 'https://github.com/aws/aws-sdk-ios.git',
-                     :tag => '2.14.2'}
+                     :tag => s.version}
```
wooj2, line 13: nice!

CircleciScripts/carthage-build.sh
```diff
@@ -0,0 +1,37 @@
...
+# Update this to include new values as new Xcode 12 versions are released
+################################################################################
+
+# carthage-build.sh
+# Usage example: ./carthage-build.sh --platform iOS
+
+set -euo pipefail
+
+xcconfig=$(mktemp /tmp/static.xcconfig.XXXXXX)
+trap 'rm -f "$xcconfig"' INT TERM HUP EXIT
```
wooj2, line 16: I've used `trap` with `INT` and `HUP` but not `EXIT` -- I am assuming that that when the script runs to completion, that we will remove the file?

CircleciScripts/carthage-build.sh
```diff
@@ -0,0 +1,37 @@
...
+# Xcode 12 beta 5
+echo 'EXCLUDED_ARCHS__EFFECTIVE_PLATFORM_SUFFIX_simulator__NATIVE_ARCH_64_BIT_x86_64__XCODE_1200__BUILD_12A8189h = arm64 arm64e armv7 armv7s armv6 armv8' >> $xcconfig
+# Xcode 12 beta 6
+echo 'EXCLUDED_ARCHS__EFFECTIVE_PLATFORM_SUFFIX_simulator__NATIVE_ARCH_64_BIT_x86_64__XCODE_1200__BUILD_12A8189n = arm64 arm64e armv7 armv7s armv6 armv8' >> $xcconfig
+# Xcode 12 GM (12A7208)
+echo 'EXCLUDED_ARCHS__EFFECTIVE_PLATFORM_SUFFIX_simulator__NATIVE_ARCH_64_BIT_x86_64__XCODE_1200__BUILD_12A7208 = arm64 arm64e armv7 armv7s armv6 armv8' >> $xcconfig
+# Xcode 12 GM (12A7209)
+echo 'EXCLUDED_ARCHS__EFFECTIVE_PLATFORM_SUFFIX_simulator__NATIVE_ARCH_64_BIT_x86_64__XCODE_1200__BUILD_12A7209 = arm64 arm64e armv7 armv7s armv6 armv8' >> $xcconfig
+
+echo 'EXCLUDED_ARCHS__EFFECTIVE_PLATFORM_SUFFIX_simulator__NATIVE_ARCH_64_BIT_x86_64__XCODE_1200 = $(EXCLUDED_ARCHS__EFFECTIVE_PLATFORM_SUFFIX_simulator__NATIVE_ARCH_64_BIT_x86_64__XCODE_1200__BUILD_$(XCODE_PRODUCT_BUILD_VERSION))' >> $xcconfig
```
wooj2, line 33: Maybe it makes sense to check the presence of the variable and warn anyone using this script.. perhaps something like this:
```
if [[ -z "${XCODE_PRODUCT_BUILD_VERSION}" ]]; then
    echo "warning: XCODE_PRODUCT_BUILD_VERSION is not defined, carthage-build.sh needs to be updated to capture correct build version of xcode"
fi
```

(and maybe for the variable XCODE_VERSION_MAJOR as well?)

CircleciScripts/generate_documentation.sh
```diff
@@ -78,7 +78,6 @@ OBJC_SDK_LIST=$(find $SOURCE_ROOT ${SOURCE_ROOT}/AWSAuthSDK/Sources -type d -max
   -not -name "*.xcodeproj" \
   -not -name "*Test*" \
   -not -name "AWSAuthSDK" \
-  -not -name "AWSCognitoIdentityProviderASF" \
```
wooj2, line 81: I guess we are generating documentation for this now because it's no longer binary distribution?

Scripts/setup_private_specs.sh
```diff
@@ -90,7 +94,9 @@ while read -r podspec_file ; do
   podspec_file_names+=("$podspec_file_name")
   pod_name=$( basename "$podspec_file_name" .podspec )
 
-  dst_dir="${COCOAPODS_REPO_DIR}/${pod_name}/${new_version}"
+  resolved_version=$( resolve_version_for_pod "${pod_name}" )
```
wooj2, line 97: when i first read this, i thought we were passing in "${pod_name}" as an argument to `resolve_version_for_pod`, but instead it looks like we are just appending the two together -- i think it would be more readable if we broke this into multiple lines -- one line for calling the funciton, and another for appending the "${pod_name}"

.circleci/config.yml
palpatim: As long as the execute bit remains set, yes. That said, we've had pipeline errors before because a script wasn't set to execute, so I usually do a leading `.` to be defensive

CircleciScripts/carthage-build.sh
palpatim: Yep, however, there's actually a bug--thanks for pointing that out! If I execute defensively (like I you pointed out above) then the trap gets assigned to the current shell, and I don't know when or if it would execute. Rather than executing with the current shell, I'll change the `carthage-build` execution to directly invoke `bash`.

.circleci/config.yml
palpatim: **Update** See below--executing with the dot-command causes the cleanup traps to not be executed. I'll change this to `bash` instead, to explicitly invoke a new subprocess and assign the traps to it

CircleciScripts/carthage-build.sh
palpatim: Both `XCODE_PRODUCT_BUILD_VERSION` and `XCODE_VERSION_MAJOR` are populated by `xcodebuild`. Is your concern that they won't be present, or that we'll miss a future version of Xcode? If the latter, we should probably switch through the known product versions. In fact, we've already got a use case for this b/c Xcode 12.2 beta is already out

CircleciScripts/generate_documentation.sh
palpatim: Correct

Scripts/setup_private_specs.sh
palpatim: We are passing `pod_name` as an argument to `resolve_version_for_pod`. It's there as a hook for pods that need special handling, like `AWSCognitoIdentityProviderASF` used to. It's actually safe to remove it now, but it's a useful extension point.

CircleciScripts/carthage-build.sh
wooj2: Concern is if Xcode ever changes their variable names and then we start putting an empty string there without a clue on how to fix it.

wooj2 (approved):

Issue #2944, opened 2020-08-14
Facebook SDK version needs upgrade


joeboyscout04:
**Describe the bug**
The `AWSFacebookSignIn` portion of the AWS SDK is currently locked to version 5.8 of `FBSDKCoreKit`/`FBSDKLoginKit` in Cocoapods.  

Facebook recently had a major outage which caused many apps to crash on startup.  They recommend all client developers updating to at least 7.0.1 of the SDK:
https://github.com/facebook/facebook-ios-sdk/blob/master/CHANGELOG.md#701

Would it be possible to upgrade the AWS SDK so we can stay on top of these changes from Facebook? 

**Areas of the SDK you are using (AWSMobileClient, Cognito, Pinpoint, IoT, etc)?**
AWSFacebookSignIn, Cognito, AWSMobileClient

**Environment:**
 - SDK Version: 2.15.1 
 - Dependency Manager: Cocoapods
 - Swift Version : 5.1
 - Xcode Version: 11.6

palpatim:
Implementation note: In addition to exploring the Facebook SDK itself, we need to understand how later versions of the SDK work with the Facebook API version. Currently Cognito only supports through API v6.0. If the SDK and API versions are tied together, that limits our ability to upgrade.

joeboyscout04:
I believe that in recent times the FB SDK major versions reflect API version, so version 7.x.x of the FB SDK uses version 7.0 of Facebooks API, version 6.x.x uses version 6.0.

palpatim:
Notes for the team:
* *Next steps* - understand how later versions of the SDK work with the Facebook API version. Currently Cognito only supports through API v6.0. If the SDK and API versions are tied together, that limits our ability to upgrade.

palpatim:
This was released on the [2.17.0](https://github.com/aws-amplify/aws-sdk-ios/releases/tag/2.17.0) release. If you pin to a particular version of the SDK, please update your Podfiles to point to at least `~> 2.17.0`.

Please note that this release includes a couple of breaking changes, so please review the release notes to see if those changes affect you.

Please let us know if you have any questions.

royjit:
PR pending - https://github.com/aws-amplify/aws-sdk-ios/pull/3043

ruiguoamz:
PR Pending - [#3082](https://github.com/aws-amplify/aws-sdk-ios/pull/3082)

wooj2:
[...]

PR #3043, opened 2020-09-24
(AWSAuthSDK) Upgrade Facebook SDK to v7.1.1

ruiguoamz:
*Issue:*

In file `FBSDKCoreKitImport.h`
the code is like
```swift
#if FBSDK_SWIFT_PACKAGE
#import <FBSDKCoreKit.h>
#else
#import <FBSDKCoreKit/FBSDKCoreKit.h>
#endif
```
but it is build failed because the code runs into `else` block and says `'FBSDKCoreKit/FBSDKCoreKit.h' file not found`. 

The previous code is like:
```swift
#import <FBSDKCoreKit.h>
```
I don't find where does `FBSDK_SWIFT_PACKAGE` defined. I am thinking can I just keep it as the previous way?

*Description of changes:*
- the sample app works (done manual test)

- Update AWSFacebookSignIn.podspec


AWSAuthSDK/AWSAuthSDK.xcodeproj/project.pbxproj
```diff
@@ -199,6 +199,21 @@
 		B9AC6319234F00FA00A713A5 /* AWSAuthUIHelper.m in Sources */ = {isa = PBXBuildFile; fileRef = B9AC6317234F00F900A713A5 /* AWSAuthUIHelper.m */; };
 		B9AC631A234F00FA00A713A5 /* AWSAuthUIHelper.h in Headers */ = {isa = PBXBuildFile; fileRef = B9AC6318234F00FA00A713A5 /* AWSAuthUIHelper.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		B9E2F6CD231EEE63009EAA48 /* AWSMobileClientInitializerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = B9E2F6CC231EEE63009EAA48 /* AWSMobileClientInitializerTests.swift */; };
+		D8FAC273251D1A87005F9FC6 /* FBSDKCoreKitImport.h in Headers */ = {isa = PBXBuildFile; fileRef = D8FAC202251D1A87005F9FC6 /* FBSDKCoreKitImport.h */; };
```
royjit, line 202: What are these changes? Below I only see some file modified and no new files.

ruiguoamz: When I finished replacing all header files and opened project file, I found out that not every header files are in the project file. So I dragged them one by one. Some files in v7 are the same with v6, I believe that's the reason you are not seeing new files.

palpatim:
How did you do the manual test? The only way we can validate this from end to end is by publishing the modified SDK to a private spec repo and consuming the updated SDK via CocoaPods. Simply pointing to a local development path isn't sufficient to ensure the podspec is set up correctly.

palpatim:
Related to https://github.com/aws-amplify/aws-sdk-ios/issues/2944

AWSAuthSDK/AWSAuthSDK.xcodeproj/project.pbxproj
palpatim: In this case, it looks like the opposite: There were files in the directory tree that were not referenced in the project file. This commit fixes that.

palpatim (approved):

Issue #1143, opened 2018-12-13
AWSS3TransferUtility download progress when coming back to foreground


mrikh:
**Describe the bug:**

The bug seems to be that we don't get callbacks for the progress continuously if we start the download, put the application in the background and then come back to the foreground. The download does keep on happening just the progress callbacks aren't received. 

I am sure the download is happening in the background as the UI gets updated once the download is complete but not while the progress is happening after coming back into the foreground.

I tried debugging the issue, it seems that the callback doesn't happen in the `NSURLSession` delegate itself inside the SDK until the download is completed. This is the method i'm using for the download:

**Edit**: I also get a callback in the progress closure just after the coming back into the foreground and just before the download is completed. I've also implemented the appropriate method inside app delegate as well

Before putting app in the background the fraction completed shows values starting from lets say 0 to 0.5 perfectly fine with the appropriate progress values, if i put the app in the background at this point and come back again, the next time i receive the progress will be just before the download is completed and that too only once.

```
static let utility = AWSS3TransferUtility.default()

static func downloadFileFromAWS(_ fileName : String, progress: AWSS3TransferUtilityProgressBlock?, completion: AWSS3TransferUtilityDownloadCompletionHandlerBlock?, transferTask : ((AWSS3TransferUtilityDownloadTask?, Error?)->())?){

        utility.configuration.maxRetryCount = 2

        let expression = AWSS3TransferUtilityDownloadExpression()
        expression.progressBlock = progress
        
        utility.downloadData(fromBucket: AWSKeys.bucket_name, key: fileName, expression: expression, completionHandler: completion).continueWith { (task) -> Any? in

            DispatchQueue.main.async {
                transferTask?(task.result, task.error)
            }

            return nil
        }
    }
```

**To Reproduce**
Use the above method to start the download after passing the appropriate callbacks. Put the application in the background and then back into the foreground. Observe that the progress closure isn't called but once the download is done, the completion is called.

**Expected behavior**
Continue to get progress callbacks.

**Screenshots**
Can't upload screenshots/video cause its an official project.

**Environment(please complete the following information):**
 - SDK Version: Latest installed via cocoapods
 - Dependency Manager: Cocoapods
 - Swift Version : 4.2

**Device Information (please complete the following information):**
 - Device: iPhone 7
 - iOS Version: 12.1
 - Specific to simulators: Nope, haven't run it on the simulator.

**Additional context**

cbommas:
@mrikh 

[...]

PR #3057, opened 2020-09-29
Progress callbacks when app comes into foreground

mrikh:
Added code to resume URLSession Tasks if their state is running when app comes back into foreground (or didBecomeActive). This is a workaround to get progress callbacks.

#1143 

I opened this issue a a couple of years back and mostly forgot about it until recently. The code I added is a workaround endorsed by the people at apple as a solution to the bug in the URLSession class. (https://developer.apple.com/forums/thread/77666)

### DIFF
--- a/AWSS3/AWSS3TransferUtility.m
+++ b/AWSS3/AWSS3TransferUtility.m
@@ -474,11 +474,30 @@ static AWSS3TransferUtility *_defaultS3TransferUtility = nil;
         
         //Recover the state from the previous time this was instantiated
         [self recover:completionHandler];
+
+        [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(applicationDidBecomeActive) name:UIApplicationDidBecomeActiveNotification object:nil];
     }
+
     return self;
 }
 
 #pragma mark - recovery methods
+
+- (void) applicationDidBecomeActive{
+    if (@available(iOS 9.0, *)) {
+        [_session getAllTasksWithCompletionHandler:^(NSArray<__kindof NSURLSessionTask *> * _Nonnull tasks) {
+            [self resumeRunningTasksForSessionTask:tasks];
+        }];
+    }else{
+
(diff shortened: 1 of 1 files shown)


palpatim:
Hi @mrikh and thanks for the PR. As discussed in https://github.com/aws-amplify/aws-sdk-ios/issues/1143, I don't think we need to go to the trouble of adding a lifecycle notification handler. Since we are already loading all tasks from the session, we can just invoke `resume` on the ones in state `NSURLSessionTaskStateRunning`.

ptwobrussell:
@palpatim @mrikh - I'd like to help get this PR get across the line because this is issue is on my shortlist of potential blockers for a new app we are releasing soon that will depend on reliable progress updates for large file uploads.

I've read up on the backlog here - https://github.com/aws-amplify/aws-sdk-ios/issues/1143 - and I'm not quite sure what is the recommended path forward for the fix. I'd be glad to help with a revised PR or some testing, but I'd feel more confident about how to spend the time if I heard back from you first.

mrikh:
@ptwobrussell Basically they don't want to include a life cycle dependency into the SDK. I can understand not wanting to do that from the perspective of developing a SDK but the only potential solution that I have found involves a dependency on it. Since it directly affects the convenience methods offered by the SDK I think it should still be handled by it instead of by the client but that is just my opinion. 

If it is critical to what you are developing, you should be able to handle that on your end. Just get all the current running tasks from `NSURLSession` (should be possible if you and the SDK both use `shared` session, worst case you might have to modify a couple of classes in the SDK to expose that property) and when you get the entered foreground notification, just iterate over all the tasks and call `[task resume]`. You will start getting callbacks for the progress again.

ptwobrussell:
@mrikh Thanks for the clarification (and for surfacing a fix/workaround)!

@mrikh @palpatim In the same way that the consumer of the SDK must opt-in and implement `interceptApplication` in the lifecycle notification `didFinishLaunchingWithOptions`, to handle critical code paths for resumes/restarts, it seems that the consumer of the SDK could be reasonably expected to opt-in here with a utility method exposed by the SDK for this situation. That way, the SDK itself doesn't have a lifecycle dependency but at the same time does make it clear and easy for consumers to handle an underlying bug with some easy copy/paste from the docs. (Otherwise they will have to experience the pain of the bug and track it down the same way we are here...)

In the meanwhile, I may just extend the AWSS3TransferUtility class to do essentially that, per a minor variation of your suggestion, so I'm certainly unblocked. 

Just wanted to offer that additional perspective in case it's helpful for consideration.

PR #3084, opened 2020-10-12
fix(awspinpoint): Mark background task as completed when app is foregrounded

### LINKED ISSUE #2651: Enter foreground but never endBackgroundTask
https://github.com/aws-amplify/aws-sdk-ios/blob/71a3a35b4622452d4a25418596a6151b31a791f2/AWSPinpoint/AWSPinpointSessionClient.m#L149

When the app enters foreground with a backgroundTask still active, it should end it, otherwise iOS writes warnings to the console.

So add something like:
    if (self.bgTask != UIBackgroundTaskInvalid) {
        [[UIApplication sharedApplication] endBackgroundTask:self.bgTask];
        self.bgTask = UIBackgroundTaskInvalid;
    }

diegocstn:
Mark background task as completed when app is foregrounded

fix #2651

*Issue #, if available:* #2651 

*Description of changes:*

### DIFF
--- a/AWSPinpoint/AWSPinpointSessionClient.m
+++ b/AWSPinpoint/AWSPinpointSessionClient.m
@@ -168,6 +168,7 @@ NSObject *sessionLock;
 
 - (void)applicationDidEnterForeground:(NSNotification*)notification {
     [self.bgTimer invalidate];
+    [self endCurrentBackgroundTask];
     [self resumeSession];
 }
 
@@ -325,8 +326,7 @@ NSObject *sessionLock;
             if (block) {
                 block(task);
             }
-            [[UIApplication sharedApplication] endBackgroundTask:self.bgTask];
-            self.bgTask = UIBackgroundTaskInvalid;
+            [self endCurrentBackgroundTask];
             return nil;
         }];
     });
@@ -362,14 +362,20 @@ NSObject *sessionLock;
     return [self.context.analyticsClient recordEvent:resumeEvent];
 }
 
+- (void)endCurrentBackgroundTask {
+    if (self.bgTask != UIBackgroundTaskInvalid) {
+        [[UIApplication sharedApplication] endBackgroundTask:self.bgTask];
+        self.bgTask = UIBackgroundTaskInvalid;
+    }
+}
+
 - (void)waitForSessionTimeoutWithCompletionBlock:(AWSPinpointTimeoutBlock) block {
     dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
         if (self.context.configuration.sessionTimeout > 0) {
             self.bgTask = [[UIApplication sharedApplication] beginBackgroundTaskWithName:AWSPinpointSessionBackgroundTask expirationHandler:^{
                 // If background task expires before timeout then stop the session and submit events.
                 [self endCurrentSessionWithBlock:block];
-                [[UIApplication sharedApplication] endBackgroundTask:self.bgTask];
-                self.bgTask = UIBackgroundTaskInvalid;
+                [self endCurrentBackgroundTask];
             }];
 
             dispatch_async(dispatch_get_main_queue(), ^(){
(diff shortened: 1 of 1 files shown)


palpatim (commented):
I agree with refactoring the duplicated code into the `endCurrentBackgroundTask` utility method. However, I'm unclear what the implication of automatically ending the background task whenever we resume foreground is. If the task is underway, what work will be lost? Is the work that is stopped going to be picked up by the `resumeSession` method?

Something that has helped in the past is a bit of doc comment to explain the lifecycle of what's going on with these kinds of events. I don't particularly like to rely on docs, but in this case it'll be useful to understand what work goes on in what phase of the app lifecycle.

diegocstn:
A new background task is started in `waitForSessionTimeoutWithCompletionBlock:` when an app is backgrounded and is ended either if its expiration handler is called or right after we submit the pending Pinpoint events.
I've identified the following scenarios that may or may not lead to create a new background task without ending a previously created one.

**Scenario 1 (happy path): transition from foreground to background and the scheduled `endCurrentSessionWithBlock` successfully completes**
Application remained foregrounded for enough time to have `endCurrentSessionWithBlock` called and the background task ended

**Scenario 2: transition from foreground to background and foreground again _after_ `endCurrentSessionWithBlock` started**
The background task will eventually marked as ended after submitting any pending event

**Scenario 3: transition from foreground to background and foreground again _before_ `endCurrentSessionWithBlock` started**
In this case, the scheduled `endCurrentSessionWithBlock` is still pending when the app is foregrounded but its timer [gets invalidated ](https://github.com/aws-amplify/aws-sdk-ios/blob/main/AWSPinpoint/AWSPinpointSessionClient.m#L170) so the previously started background task will be never ended. 

This PR addresses _scenario 3_ by ending the background task only when the app resume and there's still a pending `endCurrentSessionWithBlock` operation.

palpatim:
Lifecycle:

### App enters background

1. App enters background. Notification listener invokes
    ```objective-c
    - (void)applicationDidEnterBackground:(NSNotification*)notification {
        [self pauseSessionWithTimeoutEnabled:YES
                      timeoutCompletionBlock:nil];
    }
    ```

2. "Pauses" session with a session timeout
    ```objective-c
    - (AWSTask*)pauseSessionWithTimeoutEnabled:(BOOL) timeoutEnabled
                        timeoutCompletionBlock:(AWSPinpointTimeoutBlock) block {
    ...
            if (_session) {
                return [self pauseCurrentSessionWithTimeoutEnabled:timeoutEnabled
                                            timeoutCompletionBlock:block];
    ...
    }
    ```

3. In the `pauseCurrentSessionWithTimeoutEnabled` method, sets the session `stopTime`, `saveSession` to UserDefaults, sends a "pause event" to `recordEvent`, and in that event's completion handler calls waitForSessionTimeoutWithCompletionBlock:
    ```objective-c
    - (AWSTask*)pauseCurrentSessionWithTimeoutEnabled:(BOOL) timeoutEnabled
                               timeoutCompletionBlock:(AWSPinpointTimeoutBlock) block {
    ...
                [self waitForSessionTimeoutWithCompletionBlock:block];
    ...
    ```

4. in `waitForSessionTimeoutWithCompletionBlock`, starts a background task "AWSPinpointSessionBackgroundTask" to get more time to submit events.
  - The work to be done is:
    - On the main queue, create a bgTimer to fire when session times out. When it fires, fire `self endCurrentSessionWithBlock`, but does not invalidate the bg task
 - If the task expires before it's completed, it calls `self endCurrentSessionWithBlock`, and invalidates the bg task

5. in `endCurrentSessionWithBlock`:
  - On a global bg queue call `endCurrentSession`, which invalidates the bg timer, and records a "session stop" event.
  - Still on the global queue call `submitEvents`, then invalidate the bg task


### App enters foreground

1. App enters foreground, notification listener invokes:

    ```objective-c
    - (void)applicationDidEnterForeground:(NSNotification*)notification {
        [self.bgTimer invalidate];
        [self resumeSession];
    }
    ```

    Note this does nothing to affect the bgTask

2. `self resumeSession` either resumes the current session (if one exists, and if it hasn't expired), or starts a new one. bg task still running

palpatim:
Based on the lifecycle above, I now understand why you're invalidating the bgTask when the app resumes, and agree with the changes. :) Thanks for digging into this.

palpatim (approved):

diegocstn:
Thanks for the detailed review :)

Issue #2961, opened 2020-08-20
Crash at AWSMobileClientExtensions.swift line 109


shaikhshoeb:
I want to do facebook, google sign in
i have used below code but after sign in 
app gets crashed at this line.  Crash at AWSMobileClientExtensions.swift line 109
Can any one help ?

My code to do login--->

 let hostedUIOptions = HostedUIOptions(scopes: ["email"], identityProvider: "Google")
       AWSMobileClient.default().showSignIn(navigationController: self.navigationController!, hostedUIOptions: hostedUIOptions) { (userState, error) in
                 if (error as? AWSMobileClientError) != nil {
                                     
                 }
                 if let userState = userState {
                     print(userState.jsonValue)
                     print("Status: \(userState.rawValue)")
                   
                 }
             }

![Screenshot 2020-08-20 at 2 14 27 PM](https://user-images.githubusercontent.com/13327379/90748594-00b34b00-e2f0-11ea-80f9-1ac10e1be92b.png)

shaikhshoeb:
SDK Version: Latest updated with cocaPod update.
Dependency Manager: Cocoapods
Swift Version : 5.0
Xcode Version:  11.4
Device Information iPhone x and in simulator

Device: All device
iOS Version: iOS 13.0

Yes Crash still occurs. just send the code to login with AWSMobileClient HostOption - Social login like google facebook.

shaikhshoeb:
Using this below code


    let hostedUIOptions = HostedUIOptions(scopes: ["email"], identityProvider: "Facebook")
        AWSMobileClient.default().showSignIn(navigationController: self.navigationController!, hostedUIOptions: hostedUIOptions) { (userState, error) in
            if (error as? AWSMobileClientError) != nil {
                // appDelegate.requestAccessToCalendar()
            }
            if let userState = userState {
                print(userState.jsonValue)

            }
        }

shaikhshoeb:
It still crashing -
Check the attach screen shot 

![Screenshot 2020-08-25 at 1 35 20 PM](https://user-images.githubusercontent.com/13327379/91149308-8b24f180-e6d8-11ea-861e-dcdae684022a.png)

royjit:
Hi @shaikhshoeb Can you please check if the error happens when we pass "openid" in the list of scopes while you sign in? I think the `idToken` is not returned if we do not pass the right scopes to Cognito.

```swift
[...]

PR #3090, opened 2020-10-15
fix(AWSMobileClient): Gracefully error out if idToken is not returned from service

royjit:
*Issue #, if available:* https://github.com/aws-amplify/aws-sdk-ios/issues/2961

Id token is not returned if you do not pass 'openId' in the scope, doc - https://docs.aws.amazon.com/cognito/latest/developerguide/authorization-endpoint.html

*Description of changes:*

### DIFF
--- a/AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClientExtensions.swift
+++ b/AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClientExtensions.swift
@@ -89,7 +89,7 @@ extension AWSMobileClient: AWSIdentityProviderManager {
             
             let userPoolsTokenTask: AWSTaskCompletionSource<NSDictionary> = AWSTaskCompletionSource.init()
             self.getTokens { [weak self] (tokens, error) in
-                self?.setLoginMapTaskUsing(tokens: tokens, error: error, to: userPoolsTokenTask)
+                self?.setLoginMap(using: tokens, and: error, to: userPoolsTokenTask)
             }
             return userPoolsTokenTask.task
         } else if federationProvider == .oidcFederation {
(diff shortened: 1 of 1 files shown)


AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClientExtensions.swift
```diff
@@ -118,6 +107,23 @@ extension AWSMobileClient: AWSIdentityProviderManager {
         let task = AWSTask.init(result: dict as NSDictionary)
         return task
     }
+    
+    func setLoginMapTaskUsing(tokens: Tokens?, error: Error?, to task: AWSTaskCompletionSource<NSDictionary>) {
+        if let tokens = tokens {
```
palpatim, line 112: nit: guard/let instead of nested if. Also, is it acceptable to `task.set(error: nil)`? since `error` is an optional arg?

AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClientExtensions.swift
```diff
@@ -118,6 +107,23 @@ extension AWSMobileClient: AWSIdentityProviderManager {
         let task = AWSTask.init(result: dict as NSDictionary)
         return task
     }
+    
+    func setLoginMapTaskUsing(tokens: Tokens?, error: Error?, to task: AWSTaskCompletionSource<NSDictionary>) {
```
palpatim, line 111: nit: naming is a bit awkward.
```suggestion
    func setLoginMapTokensResult(on task: AWSTaskCompletionSource<NSDictionary>, tokens: Tokens?, error: Error?) {
```

royjit: renamed

AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClientExtensions.swift
royjit: - Changed to guard/let. 
- Added check for error = nil

palpatim (approved):

PR #3100, opened 2020-10-15
AWSEC2 Update models to latest

awsmobilesdk:
AWSEC2 Update models to latest


palpatim:
Model version conflict. Will reconcile separately.

Issue #2962, opened 2020-08-20
nil transfer utility after successful registration


lmramirez:
**State your question**

I have set up my transfer utility like so:

```
        let credentialProvider = AWSBasicSessionCredentialsProvider(accessKey: uploadCredentials.accessKeyId,
                                                                     secretKey: uploadCredentials.secretAccessKey,
                                                                     sessionToken: uploadCredentials.sessionToken)

        guard let configuration = AWSServiceConfiguration(region: .USWest2, credentialsProvider: credentialProvider) else {
            completion?(.failure(.configurationFailed))
            return
        }
        configuration.sharedContainerIdentifier = config.appGroupId

        AWSS3TransferUtility.register(
            with: configuration,
            transferUtilityConfiguration: defaultTransferUtilityConfiguration,
            forKey: utilityKey
        ) { error in
            if let error = error {
                completion?(.failure(.failedToRegister(error: error)))
                return
            }

            guard let transferUtility = AWSS3TransferUtility.s3TransferUtility(forKey: utilityKey) else {
                // *** This code path should never be hit, but it sometimes is ****
                completion?(.failure(.utilityNotFound))
                return
            }

            let uploadService = VideoFileUploadService(config: config,
                                                       transferUtility: transferUtility,
                                                       videoId: videoId,
                                                       s3Bucket: s3Bucket)

            completion?(.success(uploadService))
        }
```

I've noticed in my application logs however, that on occasion, the `transferUtility` is `nil`. The docs indicate that this should only ever happen when there was an error in the registration, however, we can see that I am checking for the error in the lines preceding the getting of the transfer utility. How is this possible?

**Which AWS Services are you utilizing?**
S3

**Provide code snippets (if applicable)**

**Environment(please complete the following information):**
 - SDK Version: 2.13.4
 - Dependency Manager: cocoapods
 - Swift Version : 5.0

**Device Information (please complete the following information):**
 - Device: iPhone 10.4, but also many others
 - iOS Version: 13.6
 - Specific to simulators: No

lawmicha:
Hi @lmramirez, Could you enable verbose logging?
```
AWSDDLog.sharedInstance.logLevel = .verbose
AWSDDLog.add(AWSDDTTYLogger.sharedInstance)
```

Do you have any relevant logs to share when it successfully registers but then retrieving it returns nil?

related code
[...]

PR #3134, opened 2020-10-17
fix(S3): call completion block after register finishes v2

wooj2:
Second attempt in attempts to resolve:
#2962

In the first attempt to fix: https://github.com/aws-amplify/aws-sdk-ios/pull/3132 
unit tests were breaking.  This is because unit tests were written with:
```
    [AWSS3 registerS3WithConfiguration:configuration forKey:key];
```
... and I suspect a lot of our customers are relying on this code to behave synchronously.

Therefore, to maintain this synchronous behavior, we'll need to introduce a semaphore to ensure that the async task finishes prior to returning with the call.

### DIFF
--- a/AWSS3/AWSS3TransferUtility.m
+++ b/AWSS3/AWSS3TransferUtility.m
@@ -320,12 +320,26 @@ static AWSS3TransferUtility *_defaultS3TransferUtility = nil;
         _serviceClients = [AWSSynchronizedMutableDictionary new];
     });
     
-    AWSS3TransferUtility *s3TransferUtility = [[AWSS3TransferUtility alloc] initWithConfiguration:configuration
+    AWSS3TransferUtility *s3TransferUtility;
+    dispatch_semaphore_t semaphoreCreate = dispatch_semaphore_create(0);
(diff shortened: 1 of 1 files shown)


wooj2:
Hmm re-looking at the code, I am hesitant with this approach because it relies on finishing a download prior to returning control back to the caller..  It seems that there are really two cases here:
1.  Customers who are calling 
```
[AWSS3 registerS3WithConfiguration:configuration forKey:key];
```
who are expecting the configuration to be immediately configured,

2.  Customers who are calling:
```
    [AWSS3TransferUtility registerS3TransferUtilityWithConfiguration:configuration
                                        transferUtilityConfiguration:[AWSS3TransferUtilityConfiguration new]
                                                              forKey:key
                                                   completionHandler:^(NSError * _Nullable error) {
    }];
```
Who are expecting waiting for a completion block to execute to allow them to know when it finishes loading transfers from prior sessions.

I thought this was going to be an easy fix, but I need to go back to the drawing board, and think about a solution to satisfy both cases.

wooj2:
Closing this PR, it does not address using this interface in an synchronous manner.

PR #3135, opened 2020-10-18
fix(S3): register S3TransferUtility prior to async loading

wooj2:
Third approach to resolving: 
https://github.com/aws-amplify/aws-sdk-ios/issues/2962

This approach should support calling `AWSS3TransferUtility.register()` in a synchronous or asynchronous manner.

In the synchronous case, code will look like something like this:
```
AWSS3TransferUtility.register(with: configuration,
    transferUtilityConfiguration: defaultTransferUtilityConfiguration,
    forKey: utilityKey)
guard let transferUtility = AWSS3TransferUtility.s3TransferUtility(forKey: utilityKey) else {
    print("failed to get transfer utility")
    return
}
//transferUtility should be available here
```

In the async case, code will look like this:
```
AWSS3TransferUtility.register(
    with: configuration,
    transferUtilityConfiguration: defaultTransferUtilityConfiguration,
    forKey: utilityKey
) { error in
    if let error = error {
        completion?(.failure(.failedToRegister(error: error)))
        return
    }

    guard let transferUtility = AWSS3TransferUtility.s3TransferUtility(forKey: utilityKey) else {
        completion?(.failure(.utilityNotFound))
        return
    }
    //transferUtility should be available here
}
```


AWSS3/AWSS3TransferUtility.m
```diff
@@ -416,7 +414,7 @@ - (instancetype)initWithConfiguration:(AWSServiceConfiguration *)serviceConfigur
         }
         
         if (identifier) {
-            _sessionIdentifier = identifier;
+            _sessionIdentifier = [NSString stringWithFormat:@"%@.%@", AWSS3TransferUtilityDefaultIdentifier, identifier];
```
wooj2, line 417: This has the potential to break clients who are calling the init function.  Thanks @palpatim

palpatim (approved):

Issue #3086, opened 2020-10-14
iOS SDK EndpointProfileUser store user attributes


yannianliu:
I saw AWS SDK for Android 2.16.11 has began to support storing user attributes, which will persist even after updating the endpoint.
￼Can iOS SDK has the same functionality soon?
Thank you

lawmicha:
Hi @yannianliu, I will mark this as a feature request

related Android, https://github.com/aws-amplify/aws-sdk-android/pull/1535

ruiguoamz:
PR https://github.com/aws-amplify/aws-sdk-ios/pull/3142 has been merged in, pending release.

palpatim:
We released this on https://github.com/aws-amplify/aws-sdk-ios/releases/tag/2.18.1

PR #3142, opened 2020-10-20
feat(Pinpoint): support for adding user attributes to PinpointEndpointProfileUser

ruiguoamz:
*Description of changes:*
Amazon Pinpoint [doc](https://docs.aws.amazon.com/pinpoint/latest/apireference/apps-application-id-users-user-id.html)
Regarding this issue: https://github.com/aws-amplify/aws-sdk-ios/issues/3086
- Added property `userAttributes`

- Added public methods for `userAttributes`: 
```objective-c
- (void)addUserAttribute:(NSArray *)theValue
                  forKey:(NSString *)theKey;
```
,
```objective-c
- (NSDictionary *)allUserAttributes;
```
,
```objective-c
- (NSArray *)userAttributeForKey:(NSString *)theKey;
```
and
```objective-c
- (BOOL)hasUserAttributeForKey:(NSString *)theKey;
```

### DIFF
--- a/AWSPinpoint/AWSPinpointEndpointProfile.h
+++ b/AWSPinpoint/AWSPinpointEndpointProfile.h
@@ -280,16 +280,30 @@ NS_ASSUME_NONNULL_BEGIN
  Returns an Dictionary of all user attributes contained within this AWSPinpointEndpointProfile
  @returns an Dictionary of all user attributes, where the attribute keys are the keys and the attribute values are the values
  */
-- (NSDictionary *) allUserAttributes;
+- (NSDictionary *)allUserAttributes;
 
 /**
-Adds an userAttribute to this AWSPinpointEndpointProfileUser with the specified key. Only 40 attributes/metrics
+Adds an user attribute to this AWSPinpointEndpointProfileUser with the specified key. Only 40 attributes
 are allowed to be added to an AWSPinpointEndpointProfileUser. If 40 attributes already exist on this AWSPinpointEndpointProfileUser, the call is ignored.
-@param theValue The value of the userAttribute. The value will be truncated if it exceeds 200 characters.
-@param theKey The key of the userAttribute. The key will be truncated if it exceeds 50 characters.
+@param theValue The value of the user attribute. The value will be truncated if it exceeds 200 characters.
+@param theKey The key of the user attribute. The key will be truncated if it exceeds 50 characters.
 */
 - (void)addUserAttribute:(NSArray *)theValue
                   forKey:(NSString *)theKey;
+
+/**
+ Returns the value of the user attribute with the specified key.
+ @param theKey The key of the user attribute to return
+ @returns The user attribute with the specified key, or null if user attribute does not exist
+ */
+- (NSArray *)userAttributeForKey:(NSString *)theKey;
+
+/**
+ Determines if this AWSPinpointEndpointProfileUser contains a specific user attribute
+ @param theKey The key of the user attribute
+ @returns YES if this AWSPinpointEndpointProfileUser has an user attribute with the specified key, NO otherwise
+ */
+- (BOOL)hasUserAttributeForKey:(NSString *)theKey;
 
 @end
 
--- a/AWSPinpoint/AWSPinpointEndpointProfile.h
+++ b/AWSPinpoint/AWSPinpointEndpointProfile.h
@@ -104,7 +104,7 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  Adds an attribute to this AWSPinpointEndpointProfile with the specified key. Only 40 attributes/metrics
  are allowed to be added to an AWSPinpointEndpointProfile. If 40 attributes/metrics already exist on this AWSPinpointEndpointProfile, the call is ignored.
- @param theValue The value of the attribute. The value will be truncated if it exceeds 200 characters.
+ @param theValue The value of the attribute. The value will be truncated if it exceeds 100 characters.
  @param theKey The key of the attribute. The key will be truncated if it exceeds 50 characters.
  */
 - (void)addAttribute:(NSArray *)theValue
@@ -277,17 +277,16 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, strong) NSString *_Nullable userId;
 
 /**
- Returns an Dictionary of all user attributes contained within this AWSPinpointEndpointProfileUser
- @returns an Dictionary of all user attributes, where the user attribute keys are the keys and the user attribute values are the values
+ Returns an NSDictionary of all user attributes contained within this AWSPinpointEndpointProfileUser
+ @returns an NSDictionary of all user attributes, where the user attribute keys are the keys and the user attribute values are the values
  */
 - (NSDictionary *)allUserAttributes;
 
 /**
-Adds an user attribute to this AWSPinpointEndpointProfileUser with the specified key. Only 40 attributes
-are allowed to be added to an AWSPinpointEndpointProfileUser. If 40 user attributes already exist on this AWSPinpointEndpointProfileUser, the call is ignored.
-@param theValue The value of the user attribute. The value will be truncated if it exceeds 200 characters.
+Adds an user attribute to this AWSPinpointEndpointProfileUser with the specified key.
+@param theValue The value of the user attribute. The value will be truncated if it exceeds 100 characters.
 @param theKey The key of the user attribute. The key will be truncated if it exceeds 50 characters.
-*/
+ */
 - (void)addUserAttribute:(NSArray *)theValue
                   forKey:(NSString *)theKey;
 
--- a/AWSPinpoint/AWSPinpointEndpointProfile.h
+++ b/AWSPinpoint/AWSPinpointEndpointProfile.h
@@ -277,14 +277,14 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, strong) NSString *_Nullable userId;
 
 /**
- Returns an Dictionary of all user attributes contained within this AWSPinpointEndpointProfile
- @returns an Dictionary of all user attributes, where the attribute keys are the keys and the attribute values are the values
+ Returns an Dictionary of all user attributes contained within this AWSPinpointEndpointProfileUser
+ @returns an Dictionary of all user attributes, where the user attribute keys are the keys and the user attribute values are the values
  */
 - (NSDictionary *)allUserAttributes;
 
 /**
 Adds an user attribute to this AWSPinpointEndpointProfileUser with the specified key. Only 40 attributes
-are allowed to be added to an AWSPinpointEndpointProfileUser. If 40 attributes already exist on this AWSPinpointEndpointProfileUser, the call is ignored.
+are allowed to be added to an AWSPinpointEndpointProfileUser. If 40 user attributes already exist on this AWSPinpointEndpointProfileUser, the call is ignored.
 @param theValue The value of the user attribute. The value will be truncated if it exceeds 200 characters.
 @param theKey The key of the user attribute. The key will be truncated if it exceeds 50 characters.
 */
--- a/AWSPinpoint/AWSPinpointEndpointProfile.h
+++ b/AWSPinpoint/AWSPinpointEndpointProfile.h
@@ -128,7 +128,7 @@ NS_ASSUME_NONNULL_BEGIN
  Returns an Dictionary of all attributes contained within this AWSPinpointEndpointProfile
  @returns an Dictionary of all attributes, where the attribute keys are the keys and the attribute values are the values
  */
-- (NSDictionary*) allAttributes;
+- (NSDictionary *)allAttributes;
 
 /**
  Adds a metric to this AWSPinpointEndpointProfile with the specified key. Only 40 attributes/metrics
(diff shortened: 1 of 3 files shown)


AWSPinpoint/AWSPinpointEndpointProfile.m
```diff
@@ -546,4 +561,83 @@ - (void)encodeWithCoder:(NSCoder *)encoder {
...
+}
+
+- (NSArray *)userAttributeForKey:(NSString *)theKey {
+    @synchronized(self) {
+        return [self.userAttributes objectForKey:theKey];
+    }
+}
+
++ (NSArray *)processUserAttributeValues:(NSArray*) values {
+    NSMutableArray *trimmedValues = [NSMutableArray arrayWithCapacity:MAX_ENDPOINT_ATTRIBUTE_VALUES];
```
wooj2, line 608: Android does not seem to reference MAX_ENDPOINT_ATTRIBUTE_VALUES.. does android need to add this too?

AWSPinpoint/AWSPinpointEndpointProfile.m
```diff
@@ -546,4 +561,83 @@ - (void)encodeWithCoder:(NSCoder *)encoder {
...
+    int valuesCount = 0;
+    for (NSString *val in values) {
+        [trimmedValues addObject:[AWSPinpointEndpointProfileUser trimValue:val]];
+        if (++valuesCount >= MAX_ENDPOINT_ATTRIBUTE_VALUES) {
+            AWSDDLogWarn(@"Only %d attributes values are allowed, attribute values has been reduced to %d values.", MAX_ENDPOINT_ATTRIBUTE_VALUES, MAX_ENDPOINT_ATTRIBUTE_VALUES);
+            break;
+        }
+    }
+    return trimmedValues;
```
wooj2, line 617: I think we should come up with a different algorithm.  This will lead to non-deterministic behavior since the order or values can vary.

AWSPinpoint/AWSPinpointEndpointProfile.m
```diff
@@ -546,4 +561,83 @@ - (void)encodeWithCoder:(NSCoder *)encoder {
...
+}
+
++ (NSString *)trimKey:(NSString*)theKey
+              forType:(NSString*)theType {
+    NSString* trimmedKey = [AWSPinpointStringUtils clipString:theKey
+                                                   toMaxChars:MAX_ENDPOINT_ATTRIBUTE_METRIC_KEY_LENGTH
```
wooj2, line 623: MAX_ENDPOINT_ATTRIBUTE_METRIC_KEY_LENGTH does not seem to be referenced in the android codebase, does android need this too?

AWSPinpoint/AWSPinpointEndpointProfile.m
```diff
@@ -546,4 +561,83 @@ - (void)encodeWithCoder:(NSCoder *)encoder {
...
+    @synchronized(self) {
+        if (theKey != nil && theValue != nil && ![self hasUserAttributeForKey:theKey]) {
+            if (self.currentNumOfUserAttributes < MAX_NUM_OF_METRICS_AND_ATTRIBUTES) {
+                [self.userAttributes setValue:[AWSPinpointEndpointProfileUser processUserAttributeValues:theValue]
+                                       forKey:[AWSPinpointEndpointProfileUser trimKey:theKey
+                                                                              forType:@"userAttribute"]];
+            } else {
+                AWSDDLogWarn(@"Max number of userAttributes/metrics reached, dropping attribute with key: %@", theKey);
+            }
+        } else if ([self hasUserAttributeForKey:theKey]) {
```
wooj2, line 576: seems like this could be an if/else rather than calling `hasUserAttributeForKey` twice?

AWSPinpoint/AWSPinpointEndpointProfile.m
```diff
@@ -546,4 +561,83 @@ - (void)encodeWithCoder:(NSCoder *)encoder {
     [encoder encodeObject:_userId forKey:@"userId"];
```
wooj2, line 561: Now that we've added userAttributes to this class, seems like we would need to add it to this function `encodeWithCoder:` ?

AWSPinpoint/AWSPinpointEndpointProfile.m
```diff
@@ -520,11 +526,19 @@ - (void)encodeWithCoder:(NSCoder *)encoder {
...
 
-- (NSString*) quotedString:(NSString*) input {
+- (instancetype) init {
+    if (self = [super init]) {
+        _userId = @"";
+        _userAttributes = [NSMutableDictionary dictionary];
+    }
+    return self;
+}
+
```
wooj2, line 536: I don't think we should be adding an init function here?  why do we need this?

AWSPinpoint/AWSPinpointEndpointProfile.h
```diff
@@ -276,7 +276,35 @@ NS_ASSUME_NONNULL_BEGIN
...
+ @returns an Dictionary of all user attributes, where the user attribute keys are the keys and the user attribute values are the values
+ */
+- (NSDictionary *)allUserAttributes;
+
+/**
+Adds an user attribute to this AWSPinpointEndpointProfileUser with the specified key. Only 40 attributes
+are allowed to be added to an AWSPinpointEndpointProfileUser. If 40 user attributes already exist on this AWSPinpointEndpointProfileUser, the call is ignored.
+@param theValue The value of the user attribute. The value will be truncated if it exceeds 200 characters.
+@param theKey The key of the user attribute. The key will be truncated if it exceeds 50 characters.
+*/
```
wooj2, line 290: Whitespace seems off here.

AWSPinpoint/AWSPinpointEndpointProfile.h
```diff
@@ -276,7 +276,35 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic, strong) NSString *_Nullable userId;
 
+/**
+ Returns an Dictionary of all user attributes contained within this AWSPinpointEndpointProfileUser
+ @returns an Dictionary of all user attributes, where the user attribute keys are the keys and the user attribute values are the values
```
wooj2, line 281: I think we should either change `Dictionary` to NSDictionary or dictionary

AWSPinpoint/AWSPinpointEndpointProfile.h
```diff
@@ -276,7 +276,35 @@ NS_ASSUME_NONNULL_BEGIN
...
+ */
+- (NSDictionary *)allUserAttributes;
+
+/**
+Adds an user attribute to this AWSPinpointEndpointProfileUser with the specified key. Only 40 attributes
+are allowed to be added to an AWSPinpointEndpointProfileUser. If 40 user attributes already exist on this AWSPinpointEndpointProfileUser, the call is ignored.
```
wooj2, line 287: How did we decide that we should ignore the call?

AWSPinpoint/AWSPinpointEndpointProfile.h
```diff
@@ -276,7 +276,35 @@ NS_ASSUME_NONNULL_BEGIN
...
+@param theValue The value of the user attribute. The value will be truncated if it exceeds 200 characters.
+@param theKey The key of the user attribute. The key will be truncated if it exceeds 50 characters.
```
wooj2, line 289: Is android also truncating these as well?

AWSPinpoint/AWSPinpointEndpointProfile.m
wooj2: I am hesitant to put any of this error checking/boundary logic into this code (that we would need to maintain).  I understand that we already have code like this in our codebase, but I am against continuing to add additional logic in our SDK that has this type of restricting logic, because I think the service should dictate what values it is allowed to accept -- not the client.  I actually tried adding 100 items to a single attribute and i got back the following error:

```
2020-10-24 15:01:54:733 SDKIssue3086[58900:645291] Response body:
{"Message":"Too many attribute values for CustomUserAttribute"}
2020-10-24 15:01:54:733 SDKIssue3086[58900:645291] Unable to successfully update endpoint. Error Message:Error Domain=com.amazonaws.AWSPinpointTargetingErrorDomain Code=1 "BadRequestException" UserInfo={responseStatusCode=400, responseHeaders=<CFBasicHash 0x600002a54440 [0x7fff8062d750]>{type = immutable dict, count = 12,
entries =>
```

AWSPinpoint/AWSPinpointEndpointProfile.h
```diff
@@ -276,7 +276,35 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic, strong) NSString *_Nullable userId;
 
+/**
```
wooj2, line 279: Instead of exposing all of these functions (which do error checking, and limit mutability, etc), I think it would be easier to just expose a single dictionary that customers can update:
```
@property (nonatomic, strong) NSDictionary<NSString*, NSArray*> *userAttributes;
```

AWSPinpoint/AWSPinpointEndpointProfile.m
wooj2: Please remove the init.  It is not needed, and since `AWSPinpointEndpointProfileUser` conforms to NSSecureCoding, adding this init actually breaks conformance with that protocol

AWSPinpoint/AWSPinpointEndpointProfile.m
```diff
@@ -520,11 +526,19 @@ - (void)encodeWithCoder:(NSCoder *)encoder {
...
+- (NSString *)quotedString:(NSString*) input {
     return [NSString stringWithFormat:@"\"%@\"", input];
 }
 
-- (NSString*) description {
+- (NSString *)description {
     return [NSString stringWithFormat:
             @"{"
```
wooj2, line 543: We will need to add printing the new `userAttributes` dictionary.

AWSPinpoint/AWSPinpointEndpointProfile.m
```diff
@@ -546,4 +561,83 @@ - (void)encodeWithCoder:(NSCoder *)encoder {
     [encoder encodeObject:_userId forKey:@"userId"];
 }
 
+- (void)addUserAttribute:(NSArray *)theValue
+                  forKey:(NSString *)theKey {
+    if (!theKey) return;
+    @synchronized(self) {
```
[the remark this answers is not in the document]
ruiguoamz: Just wanted to stay parallel with the existing methods. https://github.com/aws-amplify/aws-sdk-ios/blob/main/AWSPinpoint/AWSPinpointEndpointProfile.m#L195-L213

AWSPinpoint/AWSPinpointEndpointProfile.h
ruiguoamz: Sorry I forgot to put this link into PR description: See [here](https://docs.aws.amazon.com/pinpoint/latest/apireference/apps-application-id-users-user-id.html)

And here is a little mistake I will change. Should be `The value will be truncated if it exceeds 100 characters.`

AWSPinpoint/AWSPinpointEndpointProfile.h
ruiguoamz: Also I am trying to keep parallel with the [existing methods](https://github.com/aws-amplify/aws-sdk-ios/blob/main/AWSPinpoint/AWSPinpointEndpointProfile.h#L104-L111). I don't find anywhere saying `Only 40 attributes
are allowed to be added to an AWSPinpointEndpointProfile`

AWSPinpoint/AWSPinpointEndpointProfile.m
```diff
@@ -38,6 +38,12 @@ @interface AWSPinpointEndpointProfile()
 
 @end
 
+@interface AWSPinpointEndpointProfileUser()
+@property (nonatomic, readwrite) NSDictionary<NSString*,NSArray*> *userAttributes;
+@property (atomic, readonly) int currentNumOfUserAttributes;
```
wooj2, line 43: Seems like we can remove this property

AWSPinpoint/AWSPinpointEndpointProfile.m
```diff
@@ -520,15 +525,27 @@ - (void)encodeWithCoder:(NSCoder *)encoder {
...
 @implementation AWSPinpointEndpointProfileUser
 
-- (NSString*) quotedString:(NSString*) input {
+- (NSString *)quotedString:(NSString*) input {
     return [NSString stringWithFormat:@"\"%@\"", input];
 }
 
-- (NSString*) description {
+- (NSString *)description {
+    if (!self.userAttributes) {
```
ruiguoamz, line 533: Nil `userAttributes` is going to fail the [unit test](https://github.com/aws-amplify/aws-sdk-ios/blob/main/AWSPinpointUnitTests/AWSPinpointNSSecureCodingTests.m):

wooj2: That's interesting that this is failing for you... it is not failing for me.  When `self.userAttributes =nil`, i'm seeing this function return a `NSString` with a value of something like:
```
....."OptOut" : "ALL","Attributes":{},"Metrics":{}, "User":{"UserId" : "(null)","Attributes":(null)}}
```

AWSPinpoint/AWSPinpointEndpointProfile.m
```diff
@@ -538,12 +555,78 @@ + (BOOL)supportsSecureCoding {
...
+                                                                  forType:@"userAttribute"]];
+}
+
+- (NSDictionary *)allUserAttributes {
+    return [NSDictionary dictionaryWithDictionary:self.userAttributes];
+}
+
+- (BOOL)hasUserAttributeForKey:(NSString *)theKey {
+    return [self.userAttributes objectForKey:theKey] != nil;
+}
```
wooj2, line 590: Technically, i think this would be more correct:
```
    NSArray *items = [self.userAttributes objectForKey:theKey];
    return items.count;
```
because customers could call:
```
        user.addAttribute([], forKey: "sampleKey")
```
but... hm lets keep it as is to be consistent with the other `addAttribute`

wooj2 (approved):
lgtm! thanks for doing this!

PR #3146, opened 2020-10-21

awsmobilesdk:


palpatim:
Will pick up in combined IoT update

PR #3147, opened 2020-10-21
AWSIoTData Update models to latest

awsmobilesdk:
AWSIoTData Update models to latest


palpatim:
Will pick up in combined IoT update

PR #3182, opened 2020-10-30
fix: consider timestamp format for JSON protocol serialization

lawmicha:
*Issue #, if available:*
The JSON protocol (Rest-JSON, JSON 1.1) serialization ignores the timestampFormat trait of timestamp shapes in c2j model which causes it to default to unixTimetsamp format instead of the prescribed format.

*Description of changes:*
These changes add support for timestampFormat trait for JSON protocols. It also abstracts out the handling of timestamp shape serialization for different serialization protocols as they tend to share similar implementations.
Added tests around these changes for various protocols.


phani-srikar (approved):
Most of these changes are already reviewed in the staging repo. A fix was applied on top of those changes for deserialization in `RestJSON` protocol. These changes are tested for `RestJSON` serialization/deserialization through a sample app. Other protocols were only unit tested.

PR #3183, opened 2020-10-30
feat: use hostPrefix when available to prepend to endpoint

lawmicha:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/AWSCore/Serialization/AWSURLRequestSerialization.m
+++ b/AWSCore/Serialization/AWSURLRequestSerialization.m
@@ -93,12 +93,16 @@
     NSDictionary *actionHTTPRule = [actionRules objectForKey:@"http"];
     NSString *ruleURIStr = [actionHTTPRule objectForKey:@"requestUri"];
 
+    NSDictionary *actionEndpoint = [actionRules objectForKey:@"endpoint"];
+    NSString *endpointHostPrefix = [actionEndpoint objectForKey:@"hostPrefix"];
+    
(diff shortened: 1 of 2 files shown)


AWSCore/Serialization/AWSURLRequestSerialization.m
```diff
@@ -93,12 +93,16 @@ - (AWSTask *)serializeRequest:(NSMutableURLRequest *)request
     NSDictionary *actionHTTPRule = [actionRules objectForKey:@"http"];
     NSString *ruleURIStr = [actionHTTPRule objectForKey:@"requestUri"];
 
+    NSDictionary *actionEndpoint = [actionRules objectForKey:@"endpoint"];
+    NSString *endpointHostPrefix = [actionEndpoint objectForKey:@"hostPrefix"];
```
phani-srikar, line 97: I hope this does not crash if `actionEndpoint` is `nil`

phani-srikar (approved):
+1

AWSCore/Serialization/AWSURLRequestSerialization.m
wooj2: I wouldn't expect a crash, but i would expect that sending `objectForKey:` to `actionEndpoint`  when `actionEndpoint` is `nil` will result in `endpointHostPrefix` being nil

wooj2 (approved):

PR #3208, opened 2020-11-06
chore(IoT): optimistic workaround for cleaning up nsoutputstream

wooj2:
Currently do not have a good way to test this.

The code we currently have deployed to production follows the sample here:
https://developer.apple.com/library/archive/samplecode/SimpleURLConnections/Introduction/Intro.html#//apple_ref/doc/uid/DTS40009245

But just found that the documentation seems to explicitly state ordering of closing then removing from the runloop:
https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Streams/Articles/WritingOutputStreams.html

### DIFF
--- a/AWSIoT/Internal/AWSIoTMQTTClient.m
+++ b/AWSIoT/Internal/AWSIoTMQTTClient.m
@@ -647,8 +647,8 @@
 
 - (void)cleanUpToDecoderStream {
     self.toDecoderStream.delegate = nil;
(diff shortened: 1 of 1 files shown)


palpatim (approved):

palpatim:
Merged this in as an optimistic fix for https://github.com/aws-amplify/aws-sdk-ios/issues/2770. Will go out on next release

PR #3220, opened 2020-11-10
fix(Cognito): Refresh Cognito user pool token if id token expired

royjit:
*Issue #, if available:* https://github.com/aws-amplify/amplify-ios/issues/869

*Description of changes:* Cognito SDK does automatic token refresh only if the access token is expired. Now since Cognito added a new feature to change the expiry of id token as well, we need to check the validity of id token before returning from getSession call. 

Also reduced the time window to check for expiry from 5 min to 2 min. This avoids refreshing the token more often if the token expiry is set to 5 min. 

### DIFF
--- a/AWSCognitoIdentityProvider/AWSCognitoIdentityUser.m
+++ b/AWSCognitoIdentityProvider/AWSCognitoIdentityUser.m
@@ -243,11 +243,9 @@ static const NSString * AWSCognitoIdentityUserUserAttributePrefix = @"userAttrib
                                                                 refreshToken:refreshToken
                                                               expirationTime:expiration];
         }
-        
-        
+
         // If the session expires > 2 minutes return it. We need to check both accessToken and id Token expiry
-        // since user can change both of them in Cognito console. We are also checking `expiration` time returned
-        // from Cognito auth result to keep the logic similar to before.
+        // since user can change both of them in Cognito console.
         if(session
            && [self isSessionValid:session]
            && [expiration compare:[NSDate dateWithTimeIntervalSinceNow:2 * 60]] == NSOrderedDescending) {
--- a/AWSCognitoIdentityProvider/AWSCognitoIdentityUser.m
+++ b/AWSCognitoIdentityProvider/AWSCognitoIdentityUser.m
@@ -185,6 +185,32 @@ static const NSString * AWSCognitoIdentityUserUserAttributePrefix = @"userAttrib
     }];
 }
 
+- (BOOL) isSessionValid:(AWSCognitoIdentityUserSession * _Nonnull)session {
+    // If id token is not present we only need to check the accessToken to determine the validity of the token.
+    if (!session.idToken) {
+        return [self isTokenValid:session.accessToken];
+    } else {
+        return [self isTokenValid:session.accessToken] && [self isTokenValid:session.idToken];
+    }
+    return false;
+}
+
+// Check if the token is valid or not. Returns true if the token is valid.
+//
+// The token is consider invalid if the token expiry is less than or equal to 2 min. This 2 minute buffer is
+// given so that we do not hand over a token to the user which will get expired immediately. This guarrantees that the
+// returned session tokens are valid for atleast 2 min.
+- (BOOL) isTokenValid:(AWSCognitoIdentityUserSessionToken * _Nonnull)token {
+    if ([token.tokenClaims valueForKey:@"exp"]) {
+        int expiryWindow = 2 * 60;
+        NSTimeInterval expiryInterval = [[token.tokenClaims valueForKey:@"exp"] doubleValue];
+        NSDate *tokenExpiration =  [NSDate dateWithTimeIntervalSince1970:expiryInterval];
+        return (tokenExpiration &&
+                [tokenExpiration compare:[NSDate dateWithTimeIntervalSinceNow:expiryWindow]] == NSOrderedDescending);
+    }
+    return false;
+}
+
 /**
  Get a session
  */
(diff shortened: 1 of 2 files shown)


AWSCognitoIdentityProvider/AWSCognitoIdentityUser.m
```diff
@@ -203,13 +229,28 @@ -(instancetype) initWithUsername: (NSString *)username pool:(AWSCognitoIdentityU
...
+            session = [[AWSCognitoIdentityUserSession alloc] initWithIdToken:idToken
+                                                                 accessToken:accessToken
+                                                                refreshToken:refreshToken
+                                                              expirationTime:expiration];
+        }
+        
+        
+        // If the session expires > 2 minutes return it. We need to check both accessToken and id Token expiry
+        // since user can change both of them in Cognito console. We are also checking `expiration` time returned
+        // from Cognito auth result to keep the logic similar to before.
```
palpatim, line 250: The comparison of the logic "to before" won't be meaningful after this PR is merged. :) 
```suggestion
        // If the session expires > 2 minutes return it. We need to check both accessToken and id Token expiry
        // since customer can change both of them in Cognito console.
```

AWSCognitoIdentityProvider/AWSCognitoIdentityUser.m
```diff
@@ -203,13 +229,28 @@ -(instancetype) initWithUsername: (NSString *)username pool:(AWSCognitoIdentityU
...
         NSString * accessToken = self.pool.keychain[accessTokenKey];
-        //if the session expires > 5 minutes return it and there is at least an accessToken.
-        if(expiration && [expiration compare:[NSDate dateWithTimeIntervalSinceNow:5 * 60]] == NSOrderedDescending && accessToken){
-            NSString * idTokenKey = [self keyChainKey:keyChainNamespace key:AWSCognitoIdentityUserIdToken];
-            AWSCognitoIdentityUserSession * session = [[AWSCognitoIdentityUserSession alloc] initWithIdToken:self.pool.keychain[idTokenKey] accessToken:accessToken refreshToken:refreshToken expirationTime:expiration];
         
-            session.expirationTime = expiration;
+        AWSCognitoIdentityUserSession * session;
+        
+        // Session is available if we have expiration and accessToken.
```
palpatim, line 239: I don't know the API, but you're passing refresh token without checking it first. Any cases where that's going to bite us?

AWSCognitoIdentityProvider/AWSCognitoIdentityUser.m
```diff
@@ -203,13 +229,28 @@ -(instancetype) initWithUsername: (NSString *)username pool:(AWSCognitoIdentityU
...
+        if (expiration && accessToken) {
+            session = [[AWSCognitoIdentityUserSession alloc] initWithIdToken:idToken
+                                                                 accessToken:accessToken
+                                                                refreshToken:refreshToken
+                                                              expirationTime:expiration];
+        }
+        
+
```
palpatim, line 247: nit: whitespace
```suggestion
```

AWSCognitoIdentityProvider/AWSCognitoIdentityUser.m
```diff
@@ -203,13 +229,28 @@ -(instancetype) initWithUsername: (NSString *)username pool:(AWSCognitoIdentityU
...
         
-            session.expirationTime = expiration;
+        AWSCognitoIdentityUserSession * session;
+        
+        // Session is available if we have expiration and accessToken.
+        if (expiration && accessToken) {
+            session = [[AWSCognitoIdentityUserSession alloc] initWithIdToken:idToken
+                                                                 accessToken:accessToken
+                                                                refreshToken:refreshToken
+                                                              expirationTime:expiration];
```
palpatim, line 244: We used to set the `session.expirationTime` explicitly after initializing the new session, but now we don't. I assume that was redundant?

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 -Features for next release
 
+### Bug fixes
+
+- Fixed an issue where id token is not refreshed automatically (See [Issue #869](https://github.com/aws-amplify/amplify-ios/issues/869), [PR #3220](https://github.com/aws-amplify/aws-sdk-ios/pull/3220/))
```
palpatim, line 9: To provide a bit more context, would it be accurate to refine this by saying, "not refreshed automatically for customers who set short custom expiration intervals"?

AWSCognitoIdentityProvider/AWSCognitoIdentityUser.m
royjit: I think it should be okay, AWSCognitoIdentityUserSession takes a null refresh token.

AWSCognitoIdentityProvider/AWSCognitoIdentityUser.m
royjit: Yes, the init is doing the same. I am not sure why we had it again, so removed.

AWSCognitoIdentityProvider/AWSCognitoIdentityUser.m
royjit: Removed

AWSCognitoIdentityProvider/AWSCognitoIdentityUser.m
royjit: Changed

CHANGELOG.md
royjit: Yes, changed

palpatim (approved):

Issue #3225, opened 2020-11-13
The kSecUseNoAuthenticationUI is deprecated WARNING in AWSUICKeyChainStore.m


1mash0:
**Describe the bug**
Adding and building AWSConprehend libs causes a kSecUseNoAuthenticationUI deprecated WARNING on line 541 of AWSUICKeyChainStore.m.

Deprecations:
'kSecUseNoAuthenticationUI' is deprecated: first deprecated in iOS 9.0 - Use kSecUseAuthenticationUI instead.

**To Reproduce**
1. create on iOS project
2. add AWSComprehend libs
3. build

**Expected Behavior**
No Deprecated warning

**Code Snippet**
`query[(__bridge __strong id)kSecUseNoAuthenticationUI] = (__bridge id)kCFBooleanTrue;`

**Environment(please complete the following information):**
 - SDK Version: 2.19.0
 - Dependency Manager: Cocoapods
 - Swift Version : 5.0
 - Xcode Version: 12.1

**Device Information (please complete the following information):**
 - Device: n/a
 - iOS Version: n/a
 - Specific to simulators: n/a

royjit:
The PR is merged in and released in 2.21.0 - https://github.com/aws-amplify/aws-sdk-ios/releases/tag/2.21.0 Closing this issue.

PR #3245, opened 2020-11-20
AWSUICKeyChainStore fix 'kSecUseNoAuthenticationUI' deprecation warning

1mash0:
Issue #3225

Description of changes:
'kSecUseNoAuthenticationUI' deprecation warning.
Deployment Target is iOS 9.0, so I removed the iOS8+ process.

Sorry, I closed the PR(#3226 ) myself by mistake.
The content is the same as the PR(#3226 ).

### DIFF
--- a/AWSCore/UICKeyChainStore/AWSUICKeyChainStore.m
+++ b/AWSCore/UICKeyChainStore/AWSUICKeyChainStore.m
@@ -535,11 +535,7 @@ static NSString *_defaultService;
 #if TARGET_OS_IOS
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wunguarded-availability"
-    if (floor(NSFoundationVersionNumber) > floor(1144.17)) { // iOS 9+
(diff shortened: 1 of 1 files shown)


1mash0:
@awsmobilesdk 
please review and thx

lawmicha:
Hi @1mash0, thanks for sending our this PR. The change looks good. Was wondering about the impact of this change, is blocking your development in any way?

1mash0:
@lawmicha 
I don't think there is a problem because the library DeploymentTarget is iOS9.0 and does not enter into the process of using 'kSecUseNoAuthenticationUI' of iOS8+.

palpatim:
Looks like this file has the same availability even the [latest version of this library](https://github.com/kishikawakatsumi/UICKeyChainStore/blob/master/Lib/UICKeyChainStore/UICKeyChainStore.m) (which supports back to 8.0). Since we've not adopted the new version, I'm OK applying this to our forked source.

Thanks for the PR!

Issue #3173, opened 2020-10-28
Multipart Uploads not completing when app is terminated by the system in the background


thezonie:
**Describe the bug**
Multipart uploads appear to consist of an InitiateMultipartUpload HTTP POST, followed by a series of 5MB chunk PUT requests, finished off by a CompleteMultipartUpload POST containing the chunk info. Once the CompleteMultipartUpload is successful, the uploaded is complete, and the file appears in the S3 bucket.

If the app is terminated by the system while in the background during an upload, all of the 5MB chunk background transfers complete, but there is no final CompleteMultipartUpload POST to finish the transfer. Additionally, the 5MB chunks are re-uploaded periodically, as the system attempts to finish the upload, but it never sends the final CompleteMultipartUpload.

**To Reproduce**
1. Launch the app using Xcode on an iOS device, and start a multipart upload.
2. Once the upload starts, background the app by returning to the home screen, and then stop the app using Xcode. This simulates the app being terminated in the background by the system, so the background transfers will continue.

**Observed Behavior**
Using Charles Proxy or similar, you can see the 5MB chunk uploads complete, but no final CompleteMultipartUpload. The uploaded file does not appear in the S3 bucket.

**Expected Behavior**
Once all of the 5MB chunk uploads are complete, the final CompleteMultipartUpload request should be made to complete the upload, and the file should appear in the S3 bucket. 

If the app is left in the foreground or is sent to the background with an upload running that completes before the system terminates it, the upload completes as expected. 

**Areas of the SDK you are using (AWSMobileClient, Cognito, Pinpoint, IoT, etc)?**
AWSS3TransferUtility

**Environment(please complete the following information):**
 - SDK Version: 2.18.0
 - Dependency Manager: CocoaPods
 - Swift Version : 5
 - Xcode Version: 11.7

**Device Information (please complete the following information):**
 - Device: iPhone X
 - iOS Version: iOS 13.7

ruiguoamz:
Hi, @thezonie 

There is one behavior you described I disagree with: `If the app is terminated by the system while in the background during an upload, all of the 5MB chunk background transfers complete`. I don't think this is correct.

`S3TransferUtility` does support background upload, the upload will finish if you turn App into background. But the App terminated by system is another discussion, everything will be wiped out from memory completely. The App running in background and the App terminated is not the same thing. 

Correct me if I am wrong. Thanks for reaching out.

thezonie:
[...]

PR #3269, opened 2020-11-30
Refs #3173 - Multipart Uploads Not Completing

ptwobrussell:
*Issue #3173*

*Description of changes:*

Multi-part uploads do not complete in at least one situation when the app is forced to terminate and then restart, because the keys in the `completedPartsDictionary` are getting clobbered and failing checks cause the transfer to abort.

Steps to reproduce to bug:
* Enable debug console logging and test on a physical device that's not attached to an XCode debug session
* With a a 17mb file, initiate a multi-part upload with default settings. A 4 part upload (parts 1-4) will be initiated with chunk sizes (5, 5, 5, 2)
* Watch the console logging and force the app to terminate (e.g. call `abort` via a "kill switch" button in the app for testing purposes to simulate app termination by iOS) after one of the parts completes (likely part 4, the smaller 2MB chunk)
* Observe that the remaining three tasks are handled by the URLSession delegate but instead of trying using them, three new tasks are spun up. The task identifiers are observed to be parts 4-6 because the implementation defers to URLSession for these identifiers. The three tasks that were previously unhandled by URLSessionDelegate account for tasks 1-3, so the next three tasks are assigned 4-6.
* Wait a couple of minutes
* Open the app
* Observe in the logs that URLSession receives callbacks for tasks 4-6 and a request is sent to S3, but it's not a request to complete the upload (the logging isn't helpful here, but tracing the code shows that the task is aborted because not all parts were accounted for.)

Step to test the patch:
* Apply the patch
* Do the same steps to reproduce the bug
* Observe that tasks 4-6 complete successfully (because the patch prevents task 4 from being clobbered)

Discussion here: https://github.com/aws-amplify/aws-sdk-ios/issues/3173#issuecomment-735038092 - refs #3173 

### DIFF
--- a/AWSS3/AWSS3TransferUtility.m
+++ b/AWSS3/AWSS3TransferUtility.m
@@ -1937,6 +1937,13 @@ internalDictionaryToAddSubTaskTo: (NSMutableDictionary *) internalDictionaryToAd
     return tasks;
 }
 
+-(int)getLargeRandomNumber {
+
+    int from = 1e5;
+    int to = 1e6;
+    return (int)from + arc4random() % (to-from+1);
+}
+
 #pragma mark - Internal helper methods
 
 - (AWSTask *)callFinishMultiPartForUploadTask:(AWSS3TransferUtilityMultiPartUploadTask *)uploadTask {
@@ -2226,7 +2233,18 @@ didCompleteWithError:(NSError *)error {
             subTask.eTag = (NSString *) HTTPResponse.allHeaderFields[@"ETAG"];
             
             //Add it to completed parts and remove it from remaining parts.
-            [transferUtilityMultiPartUploadTask.completedPartsDictionary setObject:subTask forKey:@(subTask.taskIdentifier)];
+
+            // Don't clobber existing keys...
+            if ([transferUtilityMultiPartUploadTask.completedPartsDictionary objectForKey:@(subTask.taskIdentifier)]) {
+                long newKey = (long)subTask.taskIdentifier + [self getLargeRandomNumber];
+                [transferUtilityMultiPartUploadTask.completedPartsDictionary setObject:subTask forKey:@(newKey)];
+                AWSDDLogDebug(@"setObject for completedPartsDictionary key = %@ would clobber existing key!. Using key = %@ instetad", @(subTask.taskIdentifier), @(newKey));
+
+            } else {
+                AWSDDLogDebug(@"setObject for completedPartsDictionary key = %@", @(subTask.taskIdentifier));
+                [transferUtilityMultiPartUploadTask.completedPartsDictionary setObject:subTask forKey:@(subTask.taskIdentifier)];
+            }
+
             [transferUtilityMultiPartUploadTask.inProgressPartsDictionary removeObjectForKey:@(subTask.taskIdentifier)];
             //Update progress
             transferUtilityMultiPartUploadTask.progress.completedUnitCount = transferUtilityMultiPartUploadTask.progress.completedUnitCount - subTask.totalBytesSent + subTask.totalBytesExpectedToSend;
@@ -2277,6 +2295,9 @@ didCompleteWithError:(NSError *)error {
                     NSString *errorMessage = [NSString stringWithFormat:@"Expected to send [%@], but sent [%@] and there are no remaining parts. Failing transfer ",
                                               transferUtilityMultiPartUploadTask.contentLength, @(totalBytesSent)];
                     
+                    AWSDDLogError(@"Expected to send [%@], but sent [%@] and there are no remaining parts. Failing transfer",
+                                  transferUtilityMultiPartUploadTask.contentLength, @(totalBytesSent));
+
                     NSDictionary *userInfo = [NSDictionary dictionaryWithObject:errorMessage
                                                                          forKey:@"Message"];
                     


AWSS3/AWSS3TransferUtility.m
```diff
@@ -2226,7 +2233,18 @@ - (void)URLSession:(NSURLSession *)session
             subTask.eTag = (NSString *) HTTPResponse.allHeaderFields[@"ETAG"];
             
             //Add it to completed parts and remove it from remaining parts.
-            [transferUtilityMultiPartUploadTask.completedPartsDictionary setObject:subTask forKey:@(subTask.taskIdentifier)];
+
+            // Don't clobber existing keys...
+            if ([transferUtilityMultiPartUploadTask.completedPartsDictionary objectForKey:@(subTask.taskIdentifier)]) {
```
lawmicha, line 2238: The change looks like it will fix the issue, however my concern is for future readers to understand why there would be an existing key in the `completedPartsDictionary` without going over this PR to understand that taskIdentifiers from NSURLSessionUploadTask is not unique in some scenarios. 

My question is then can we use the `subtask.partNumber` to uniquely identify the subtasks? This seems like a logical change at first and would apply to updating the structures of `waitingPartsDictionary`, `completedPartsDictionary`, and `inProgressPartsDictionary` but I'm afraid of a regression without diving into every interaction of those three structures.

So I think a more isolated change, that will still keep the door open for possibly using partNumber as the key for `inProgressPartsDictionary` and `waitingPartsDictionary` is to only change the strcture of `completedPartsDictionary` to
```
@property (strong, nonatomic) NSMutableSet <AWSS3TransferUtilityUploadSubTask *> *completedPartsSet;
```

Running over some of the uses of the set

Hydrating from DB should add it to the set uniquely based on the object which includes the taskIdentifier,  partNumber, and rest of the fields
```objective-c
//Check if the subTask is is already completed. If it is, add it to the completed parts list, update the progress object and go to the next iteration of the loop
            if (subTask.status== AWSS3TransferUtilityTransferStatusCompleted ) {
                [multiPartUploadTask.completedPartsSet addObject:subTask];
                continue;
            }
```

Calling finish will construct the AWSS3CompletedPart in ascending order of partNumber
```objective-c
//Create a new Dictionary with the partNumber as the Key
    for(AWSS3TransferUtilityUploadSubTask *subTask in uploadTask.completedPartsSet) {
        [tempDictionary setObject:subTask forKey:subTask.partNumber];
    }
    
    //Compose the request.
    for(int i = 1; i <= [uploadTask.completedPartsSet count]; i++) {
        AWSS3TransferUtilityUploadSubTask *subTask = [tempDictionary objectForKey: [NSNumber numberWithInt:i]];
        AWSS3CompletedPart *completedPart = [AWSS3CompletedPart new];
        completedPart.partNumber = subTask.partNumber;
        completedPart.ETag = subTask.eTag;
        [completedParts addObject:completedPart];
    }
```

OnComplete (the code at this line) will just add it to the set and be deduped to a new element in the set even when the taskIdentifer is the same (assuming partNumber if the unique identifier)
```objective-c
//Add it to completed parts and remove it from remaining parts.
            [transferUtilityMultiPartUploadTask.completedPartsSet addObject:subTask];
            [transferUtilityMultiPartUploadTask.inProgressPartsDictionary removeObjectForKey:@(subTask.taskIdentifier)];
```

lawmicha:
I have some rough code that performs the mulit-part upload of 17 MB's ie.
```swift
Button("Upload", action: {
                let largeDataObject = Data(repeating: 0xff, count: 1_024 * 1_024 * 17) // 17 MB
                writeToFile(data: largeDataObject, fileName: "\(UUID().uuidString).txt")
            })
```

```swift
func writeToFile(data: Data, fileName: String){
    // get path of directory
    guard let directory = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).last else {
        return
    }
    // create file url
    let fileurl =  directory.appendingPathComponent("\(fileName).txt")
    // if file exists then write data
    if FileManager.default.fileExists(atPath: fileurl.path) {
        if let fileHandle = FileHandle(forWritingAtPath: fileurl.path) {
            // seekToEndOfFile, writes data at the last of file(appends not override)
            fileHandle.seekToEndOfFile()
            fileHandle.write(data)
            fileHandle.closeFile()
        }
        else {
            print("Can't open file to write.")
        }
    }
    else {
        // if file does not exist write data for the first time
        do{
            try data.write(to: fileurl, options: .atomic)
        }catch {
            print("Unable to write in new file.")
        }
    }
    
    uploadData(imageUrl: fileurl)
}
func uploadData(imageUrl: URL) {
        let expression = AWSS3TransferUtilityMultiPartUploadExpression()
        expression.progressBlock = {(task, progress) in
            DispatchQueue.main.async(execute: {
                // Do something e.g. Update a progress bar.
                self.percent = progress.fractionCompleted
                print("Progress: \(progress)")
            })
        }
        
        var completionHandler: AWSS3TransferUtilityMultiPartUploadCompletionHandlerBlock?
        completionHandler = { (task, error) -> Void in
            DispatchQueue.main.async(execute: {
                // Do something e.g. Alert a user for transfer completion.
                // On failed uploads, `error` contains the error object.
            })
        }
        
        let transferUtility = AWSS3TransferUtility.default()
        print("imageUrl is: \(imageUrl)")
        
        transferUtility
            .uploadUsingMultiPart(fileURL: imageUrl,
                                  bucket: "[BUCKET]",
                                  key: "public/image.png",
                                  contentType: "png",
                                  expression: expression,
                                  completionHandler: completionHandler)
            .continueWith { (task) -> AnyObject? in
                if let error = task.error {
                    print("Error: \(error.localizedDescription)")
                }

                if let result = task.result {
                    // Do something with uploadTask.
                    print(result)
                }
                return nil;
            }
    }

```

lawmicha:
Hi @ptwobrussell, could you elaborate on your steps for this? Specifically "enable debug console logging"
> Enable debug console logging and test on a physical device that's not attached to an XCode debug session

I've got my app set up as
```swift
AWSDDLog.sharedInstance.logLevel = .verbose
AWSDDLog.add(AWSDDTTYLogger.sharedInstance)
```
I am able to select my iOS device from the list in the Console app, filter on the app Process, clicked on Action-> Include Debug Messages, but still don't see the visibility of logs that I expected to see.

AWSS3/AWSS3TransferUtility.m
lawmicha: Will wait for your feedback on my comments regarding the use of a Set instead of generating a random number, here's the PR with my proposed change https://github.com/aws-amplify/aws-sdk-ios/pull/3282/files Since I couldn't reproduce  yet, wanted to still send this out incase you get a chance verify the fix on your end

ptwobrussell:
@lawmicha I had some trouble with that myself. Not at my computer at the moment but try this:

```
AWSDDLog.sharedInstance.logLevel = .verbose
AWSDDLog.add(AWSDDOSLogger.sharedInstance)
```

I didn't spend too much time on it, but what I observed while working on this was that I was unable to use the TTY logger and get what I needed to see in the Console.app system logs. It seemed as though the logs were not being propagated properly, or at least as I would've expected them to be, when both loggers (TTY and OS) were in use so make sure you are only using the OS one.

As I recall, the logging issue may have only been an issue when I was simultaneously connected to an XCode bugging session. 

However, I found success when I killed my XCode debugging session and ran the app on my hardware device while it was connected to my machine with a cable. Then I could observe the Console app logging I needed at the debug and verbose levels.

ptwobrussell:
@lawmicha - Just getting a chance to catch up here. What can I do at this point to support you and help move this forward?

lawmicha:
> @lawmicha - Just getting a chance to catch up here. What can I do at this point to support you and help move this forward?

As I had some troubles reproducing the issue, I can only test the success case. Could you try taking my PR and see if it fixes the issue for you? https://github.com/aws-amplify/aws-sdk-ios/pull/3282/files  the PR contains the proposed change (see [comment](https://github.com/aws-amplify/aws-sdk-ios/pull/3269#discussion_r537124364)) that doesn't require the need to generate a random large number. I'll try again and report back as well if i am able to reproduce the issue

ptwobrussell:
@lawmicha - I just tested your patch several times, and it does appear to fix the issue for me. Nice work! I like your approach. I may leave a comment or two on your patch just to help clarify some things in my own mind if that's alright since I'm so intellectually vested at this point that I want to follow it the whole way through...

ptwobrussell:
@lawmicha - Also, no idea if it would be useful to you, but I'd be happy to share my sample project that I've been using for testing and help you get your logging setup anytime over a screen share. (As I'm sure you know, it took quite a bit of work to even get setup to troubleshoot this issue, and I want to pay that forward however I can...)

PR #3282, opened 2020-12-06
fix: S3TransferUtility unique identify completedParts

lawmicha:
*Issue #, if available:*
proposal for changes from https://github.com/aws-amplify/aws-sdk-ios/pull/3269
*Description of changes:*

### DIFF
--- a/AWSS3/AWSS3TransferUtility.m
+++ b/AWSS3/AWSS3TransferUtility.m
@@ -107,7 +107,7 @@ static int const AWSS3TransferUtilityMultiPartDefaultConcurrencyLimit = 5;
 @property BOOL cancelled;
 @property BOOL temporaryFileCreated;
 @property NSMutableDictionary <NSNumber *, AWSS3TransferUtilityUploadSubTask *> *waitingPartsDictionary;
-@property (strong, nonatomic) NSMutableDictionary <NSNumber *, AWSS3TransferUtilityUploadSubTask *> *completedPartsDictionary;
+@property (strong, nonatomic) NSMutableSet <AWSS3TransferUtilityUploadSubTask *> *completedPartsSet;
 @property (strong, nonatomic) NSMutableDictionary <NSNumber *, AWSS3TransferUtilityUploadSubTask *> *inProgressPartsDictionary;
 @property int retryCount;
 @property int partNumber;
@@ -582,7 +582,7 @@ static AWSS3TransferUtility *_defaultS3TransferUtility = nil;
             }
             //Check if the subTask is is already completed. If it is, add it to the completed parts list, update the progress object and go to the next iteration of the loop
             if (subTask.status== AWSS3TransferUtilityTransferStatusCompleted ) {
-                [multiPartUploadTask.completedPartsDictionary setObject:subTask forKey:@(sessionTaskID)];
+                [multiPartUploadTask.completedPartsSet addObject:subTask];
                 continue;
             }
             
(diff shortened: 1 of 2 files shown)


AWSS3/AWSS3TransferUtility.m
```diff
@@ -2433,7 +2432,7 @@ - (void)URLSession:(NSURLSession *)session
     
         //Calculate the total sent so far
         int64_t totalSentSoFar = 0;
-        for (AWSS3TransferUtilityUploadSubTask *aSubTask in [transferUtilityMultiPartUploadTask.completedPartsDictionary allValues]) {
+        for (AWSS3TransferUtilityUploadSubTask *aSubTask in transferUtilityMultiPartUploadTask.completedPartsSet) {
```
ptwobrussell, line 2435: @lawmicha - This is the line of the patch that I still don't fully understand. If we are unconditionally adding `AWSS3TransferUtilityUploadSubTask` objects (of type `id`?) to `completedPartsSet` when we hydrate from the database, how do we not end up with duplicates here ini the accounting for the arithmetic in the for loop?

AWSS3/AWSS3TransferUtility.m
```diff
@@ -1941,17 +1941,16 @@ - (NSMutableArray *) getTasksHelper:(AWSSynchronizedMutableDictionary *)dictiona
 
 - (AWSTask *)callFinishMultiPartForUploadTask:(AWSS3TransferUtilityMultiPartUploadTask *)uploadTask {
     
-    NSMutableArray *completedParts = [NSMutableArray arrayWithCapacity:[uploadTask.completedPartsDictionary count]];
+    NSMutableArray *completedParts = [NSMutableArray arrayWithCapacity:[uploadTask.completedPartsSet count]];
     NSMutableDictionary *tempDictionary = [NSMutableDictionary new];
     
     //Create a new Dictionary with the partNumber as the Key
```
ptwobrussell, line 1947: This is now a bad comment since we're not using the dictionary.

ptwobrussell (commented):
FWIW, I found the comment on lines 705-706 where `handleUnlinkedTransfers` is called to be incredibly misleading and confusing. A little clarity on how the multi-part uploads are "linked" would really help future readers of this code.

Also, if the string that's put into the userInfo dict around 2276 (`@"Expected to send [%@], but sent [%@] and there are no remaining parts. Failing transfer "`) were logged, it would have been incredibly helpful in tracking down the issue. 90% of my effort was understanding everything to get to that point. If I had seen that message in the console and worked backwards, it would have really simplified my experience determining the root cause.

AWSS3/AWSS3TransferUtility.m
lawmicha: good point, I'm not sure what is persisted in the database so there might be rows which have the same part numbers. This is more reason to have a dictionary based on partNumber. the defensive thing we can do here is similar to the logic in `callFinishMultiPartForUploadTask` by dedupping the completedPartsSet to a dictionary based on part numbers

AWSS3/AWSS3TransferUtility.m
```diff
@@ -2433,7 +2432,12 @@ - (void)URLSession:(NSURLSession *)session
     
         //Calculate the total sent so far
         int64_t totalSentSoFar = 0;
-        for (AWSS3TransferUtilityUploadSubTask *aSubTask in [transferUtilityMultiPartUploadTask.completedPartsDictionary allValues]) {
+        //Create a new Dictionary with the partNumber as the Key
+        NSMutableDictionary *completedPartsByPartNumber = [NSMutableDictionary new];
+        for(AWSS3TransferUtilityUploadSubTask *subTask in transferUtilityMultiPartUploadTask.completedPartsSet) {
+            [completedPartsByPartNumber setObject:subTask forKey:subTask.partNumber];
+        }
+        for (AWSS3TransferUtilityUploadSubTask *aSubTask in [completedPartsByPartNumber allValues]) {
```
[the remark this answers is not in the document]
lawmicha: i'm being defensive here in case the `completedPartsSet` contains duplicate partNumber, which is populated when the app calls `recovery` and hydrate the data from DB. I don't know exactly if there would be, but best be defensive about this, considering we are fixing a similar bug with this change. Another consideration was to turn the completedParts dictionary to key off of the partNumber. But with this solution with the Set, we should make sure to handle any items with the same partNumber

royjit (approved):
LGTM

PR #3334, opened 2020-12-18
chore(auth): Create a new target named AWSMobileClientXCF to support xcframework

royjit:
*Issue #, if available:*

*Description of changes:*


.swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
palpatim, line 1: Should we be committing the .swiftpm dir? I don't think there's a consensus on this, but I'd rather not commit unless we can prove it's necessary.

AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClientXCF.plist
```diff
@@ -0,0 +1,24 @@
...
+	<key>CFBundleIdentifier</key>
+	<string>$(PRODUCT_BUNDLE_IDENTIFIER)</string>
+	<key>CFBundleInfoDictionaryVersion</key>
+	<string>6.0</string>
+	<key>CFBundleName</key>
+	<string>$(PRODUCT_NAME)</string>
+	<key>CFBundlePackageType</key>
+	<string>FMWK</string>
+	<key>CFBundleShortVersionString</key>
+	<string>2.19.1</string>
```
palpatim, line 18: Make sure this gets added to our bump version script

AWSAuthSDK/Sources/AWSMobileClient/Internal/AWSCognitoAuth+Extensions.h
```diff
@@ -13,9 +13,8 @@
 // permissions and limitations under the License.
 //
 
-#import <AWSMobileClient/AWSCognitoAuth.h>
```
palpatim, line 16: What prompted this change? I think I agree with it, but am curious to know what the side effects are.

AWSAuthSDK/Sources/AWSMobileClient/Internal/AWSMobileClientXCF-Mixed-Swift.h
```diff
@@ -0,0 +1,30 @@
...
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+//
+
+// #import this header file instead of <AWSMobileClient/AWSMobileClient-Swift.h>
+// This workaround allows for this package to be compiled as a static lib
+// or dynamic framework
+// https://github.com/CocoaPods/CocoaPods/issues/7594
+#if __has_include("AWSMobileClient-Swift.h")
```
palpatim, line 20: Should we reverse this test to explicitly check for the XCF flavor and default to the non-XCF version?

.swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
royjit: Yes, this might have created  due to my testing of spm. Removed.

AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClientXCF.plist
royjit: Added

AWSAuthSDK/Sources/AWSMobileClient/Internal/AWSCognitoAuth+Extensions.h
royjit: This file will be used by both target AWSMobileClient and AWSMobileClientXCF, so the import doesnot work. Looking at the content below we donot need the whole file imported and just `AWSCognitoAuthConfiguration` works.

I do not see any side effect because this is internally used by AWSMobileClient class where the header file is already imported in `AWSMobileClient.h` and `AWSMobileClientXCF.h` file.

AWSAuthSDK/Sources/AWSMobileClient/Internal/AWSMobileClientXCF-Mixed-Swift.h
royjit: Actually the import statement should be changed, updated.

palpatim (approved):
LGTM--I think this lays a solid foundation for building xcframeworks across the entire SDK for SPM support. Before merging into develop, make sure this works as expected in build steps and sample apps that use both Carthage and binary frameworks.

danthorpe:
I'm just wondering, what is the expectation for consuming this via SwiftPM? In particular, I'm interested in how to setup a test target in a package for testing a target which uses this as a dependency. 

For example, I would have thought that the following is enough:

```swift
let package = Package(
    name: "MyPackage",
    platforms: [.iOS(.v14)],
    products: [
        .library(name: "MyPackage", targets: ["MyPackage"])
    ],
    dependencies: [
        .package(name: "AWSiOSSDKV2", url: "https://github.com/aws-amplify/aws-sdk-ios-spm", from: "2.24.0"),
    ],
    targets: [
        .target(
            name: "MyPackage",
            dependencies: [
                .product(name: "AWSCore", package: "AWSiOSSDKV2"),
                .product(name: "AWSAuthCore", package: "AWSiOSSDKV2"),
                .product(name: "AWSCognitoIdentityProvider", package: "AWSiOSSDKV2"),
                .product(name: "AWSMobileClientXCF", package: "AWSiOSSDKV2")
            ]
        ),
        .testTarget(
            name: "MyPackageTests",
            dependencies: [
                "MyPackage",
            ]
        )
    ]
)
```

But running tests for `MyPackage` fails with an error of:

> Library not loaded: @rpath/AWSCognitoIdentityProviderASF.framework/AWSCognitoIdentityProviderASF

Issue #3362, opened 2021-01-07
[AWSCore] API request signature uses uncoded query parameters


edwinveger:
This issue is related to / a duplicate of [issue 977 in the Amplify repository](https://github.com/aws-amplify/amplify-ios/issues/977). I have traced the issue to AWSCore.

**Describe the bug**
Signing takes place prior to URL query encoding. This fails the request if there are query parameters contains characters that need to be encoded, e.g. `@` or `:`.

**To Reproduce**
Perform a REST request with a query parameters that contains one or more characters that will be encoded by URLSession  when making the request (e.g. `/my/endpoint/path/?email=my@email.address.com` or `/dates/post/?s=2021-06-18T09:00:00Z`).

**Observed Behavior**
In Amplify this fails with a 403 response status code.

```
{
  "error": "The request signature we calculated does not match the signature you provided. Check your AWS Secret Access Key and signing method. Consult the service documentation for details.\n\nThe Canonical String for this request should have been\n'GET\n/my/endpoint/path/\nemail=my%40email.address.com\ncontent-type:application/json\nhost:REDACTED\nuser-agent:amplify-iOS/(Function) iOS/14.3 en\nx-amz-date:20201217T105627Z\nx-amz-security-token:REDACTED\n\ncontent-type;host;user-agent;x-amz-date;x-amz-security-token\REDACTED'\n\nThe String-to-Sign should have been\n'AWS4-HMAC-SHA256\REDACTED\n20201217/us-east-1/execute-api/aws4_request\REDACTED'\n"
}
```

**Expected Behavior**
A successful request since signing should be OK.

**Suggested fix**
I have traced the issue to `AWSSignature.m`. If I add the following lines at line 327 (right after setting `query`), it works.

```swift
NSCharacterSet* querySet = [[NSCharacterSet characterSetWithCharactersInString:@"!*'\();:@+$,%#[] "] invertedSet];
query = [query stringByAddingPercentEncodingWithAllowedCharacters:querySet];
```
This ensures that the proper encoded query string is used for signing.
It seems like a very common use case, and seeing the simplicity of this fix I must say I am quite confused that this issue has not been reported/addressed earlier. Am I missing something?

**Misc**
[...]

Issue #3215, opened 2020-11-10
[AWSSignatureV4Signer generateQueryStringForSignatureV4WithCredentialProvider: ...] doesn't handle spaces (or other characters requiring encoding) correctly


mojodna:
When spaces (or other non-URI-safe characters) are present in the target path, the `keyPath` argument must be percent-encoded (using `.addingPercentEncoding(withAllowedCharacters: .urlPathAllowed)`. However, `urlComponents.percentEncodedPath` effectively negates the encoding step that has occurred. SigV4 requires that [the canonical request](https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html) be URI-encoded **twice** (step 2). In [`AWSSignatureV4Signer generateQueryStringForSignatureV4WithCredentialProvider: ...`](https://github.com/aws-amplify/aws-sdk-ios/blob/0f985322625566a566726e8ca2c24d5a5bc678b8/AWSCore/Authentication/AWSSignature.m#L390), it is effectively URI-encoded **zero** times.

**Observed Behavior**

Requests using signed URLs with spaces in their paths fail with authentication errors at the the API Gateway level.

**Expected Behavior**

Requests using signed URLs with spaces in their paths succeed.

**Code Snippet**

```swift
let url = "/path with spaces"
// URL-encode spaces, etc.
let keyPath = String(url.path.dropFirst())
percentEncodedKeyPath = keyPath.addingPercentEncoding(withAllowedCharacters: .urlPathAllowed)!

let task = AWSSignatureV4Signer
    .generateQueryStringForSignatureV4(
        withCredentialProvider: credentialsProvider,
        httpMethod: .GET,
        expireDuration: 60,
        endpoint: endpoint!,
        keyPath: percentEncodedKeyPath,
        requestHeaders: requestHeaders,
        requestParameters: .none,
        signBody: true)

task.waitUntilFinished()

if let error = task.error as NSError? {
    print("Error occurred: \(error)")
}

if let result = task.result {
  let signedUrl = result as URL

  // requesting signedUrl results in authentication failures because the canonical requests do not match
}
```

When I attempt to work around this by double-encoding it myself, the signed URL includes the double-encoding rather than matching the original URL:

```swift
let url = "/path with spaces"
// URL-encode spaces, etc.
let keyPath = String(url.path.dropFirst())
percentEncodedKeyPath = keyPath.addingPercentEncoding(withAllowedCharacters: .urlPathAllowed)!

[...]

PR #3378, opened 2021-01-16
Add URL query parameter encoding to AWSSignature

edwinveger:
Issue #3362
Issue #3215 

*Description of changes:*
Apply URL encoding of special characters in URL query parameters for use in signing.
As mentioned in #3362, this PR is missing tests and quite rudimentary overall. Its main purpose is to advance the resolution of above mentioned issues.

### DIFF
--- a/AWSCore/Authentication/AWSSignature.m
+++ b/AWSCore/Authentication/AWSSignature.m
@@ -324,6 +324,11 @@ NSString *const AWSSignatureV4Terminator = @"aws4_request";
         query = [NSString stringWithFormat:@""];
     }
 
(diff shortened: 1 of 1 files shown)


palpatim:
@edwinveger Thank you for the PR here. This is a good candidate for unit tests to ensure a) it fixes the issue at hand; and b) doesn't cause any regressions. If you have time to add tests to this PR, please feel free, otherwise, we'll pick it up and add the tests before we merge as soon as we can.

diegocstn:
@edwinveger thanks for your PR. 
`AWSCore` is responsible for signing requests for both _AWSAPIGateway_ and _Amplify_ but the former is already correctly encoding query parameters per SigV4 rules, so we decided to apply this fix to the API category in _Amplify_ .
I'll close this in favor of a [new PR](https://github.com/aws-amplify/amplify-ios/pull/1080).

PR #3310, opened 2020-12-16
feat(auth): Add support for ASWebAuthenticationSession

royjit:
*Description of changes:* Adds support for `ASWebAuthenticationSession` to `AWSCognitoAuth` and `AWSMobileClient`. Since ASWebAuthenticationSession require a presentation anchor to show the UI, the following new apis are added:

AWSCognitoAuth

```objc

- (void)getSessionWithWebUI:(nonnull ASPresentationAnchor) anchor
                  completion:(nullable AWSCognitoAuthGetSessionBlock) completion API_AVAILABLE(ios(13));

- (void) signOutWithWebUI:(nonnull ASPresentationAnchor) anchor
                completion:(AWSCognitoAuthSignOutBlock)completion API_AVAILABLE(ios(13));
```

AWSMobileClient

```swift

@available(iOS 13, *)
     public func showSignIn(presentationAnchor: ASPresentationAnchor,
                            hostedUIOptions: HostedUIOptions,
                            _ completionHandler: @escaping(UserState?, Error?) -> Void) 

@available(iOS 13, *)
     public func signOut(presentationAnchor: ASPresentationAnchor,
                         options: SignOutOptions = SignOutOptions(),
                         completionHandler: @escaping ((Error?) -> Void))
```

Also a new parameter is also added for private session in `HostedUIOptions` and `AWSCognitoAuthConfiguration`. 

---

The support for `ASWebAuthenticationSession` in the Auth SDK is from 13.0 because `ASPresentationAnchor` support is from 13.0. The following test cases were manually tested:

AWSMobileClient

*Case a: SignIn with SFAuth and SignOut with SFAuth (iOS > 13.0, iOS 12.0)*

1. SignIn using SFAuthentication was successful and successfully retrieved credentials
```swift
AWSMobileClient.default().showSignIn(navigationController: self.navigationController,
                                    hostedUIOptions: hostedUIOptions)
```
2. SignOut using SFAuthentication flow was successful
```swift
AWSMobileClient.default().signOut { (error) in

}
```
Worked in iOS 14.0 and iOS 12.0

*Case b: SignIn With SFAuth and SignOut using ASWebAuth*

1. SignIn using SFAuthentication was successful and successfully retrieved credentials
```swift
AWSMobileClient.default().showSignIn(navigationController: self.navigationController,
                                    hostedUIOptions: hostedUIOptions)
```
2. SignOut using ASWebAuthenticationSession flow was successful
```swift
[...]

PR #3402, opened 2021-01-30
Bug fix(Cognito): Use given vc for signout 

kateinoigakukun:
https://github.com/aws-amplify/aws-sdk-ios/pull/3310 changed `AWSCognitoAuth signOut` to ignore vc parameter given as a method parameter, but the vc should be used to present SFSafariVC.


royjit (approved):

royjit:
Thank you @kateinoigakukun for the fix, this will be available in the next release.

PR #3424, opened 2021-02-07
fix: polly lexicon not found test

lawmicha:
*Issue #, if available:*
The integration tests is sometimes failing with the [error](https://app.circleci.com/pipelines/github/aws-amplify/aws-sdk-ios/4375/workflows/63e6e816-d734-4193-8bfb-13942ddb919d/jobs/56684/steps)
```
  testSynthesize, ((task.error) == nil) failed: "Error Domain=com.amazonaws.AWSPollyErrorDomain Code=11 "Lexicon not found" UserInfo={NSLocalizedDescription=Lexicon not found, NSLocalizedFailureReason=LexiconNotFoundException:http://polly.amazonaws.com/doc/v1}"
```

*Description of changes:*
It appears that the lexicons are being added and deleted on every test, when the test is set up and teared down. According to the docs for [PutLexicon](https://docs.aws.amazon.com/polly/latest/dg/API_PutLexicon.html) there is an eventual consistency which is probably why there is a `sleep(20)` in the test. 

My proposal in this fix is to remove the deletion of the lexicons, and keep them in the region for use across all the tests. We continue to update the lexicons on every test run, which will have no impact on the tests as it will eventually overwrite the previous lexicon of the same name, and this also sets up the tests so that it can be run from a new region or an region with no lexicons.

### DIFF
--- a/AWSPollyTests/AWSPollyTests.m
+++ b/AWSPollyTests/AWSPollyTests.m
@@ -79,7 +79,6 @@ static NSDictionary<NSString *,NSString *> *lexicons;
 
 - (void)tearDown {
     [super tearDown];
-    [self deleteLexicon];
 }
 
 - (void)setUpLexicon {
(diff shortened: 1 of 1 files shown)


palpatim (approved):
LGTM. There's a weak argument that we should be covering the `deleteLexicon` use case, but if we want to cover that, we should have a separate, dedicated test for it. I'm not convinced that is an 80% use case for mobile anyway.

PR #3461, opened 2021-03-05
fix: Fix ignored isInitialCustomChallenge parameter in AWSCognitoIdentityUser

johntmcintosh:
*Issue #, if available:*
n/a

*Description of changes:*
I found that one of the method variants of `[AWSCognitoIdentityUser getSession]` took `isInitialCustomChallenge` as an input parameter, but didn't forward that parameter to the method variant that it called, which resulted in a value of `NO` always being used, regardless of what value the API consumer passed in.

*Check points:*

- [x] All unit tests pass
- [x] Updated CHANGELOG.md

FYI - The AWSCognitoIdentityProviderTests fail when I run them with the following exception being thrown:

> *** Terminating app due to uncaught exception 'createUserPool error', reason: 'Error: Error Domain=com.amazonaws.AWSServiceErrorDomain Code=7 "(null)" UserInfo={__type=UnrecognizedClientException, message=The security token included in the request is invalid.}'

but they were also doing this before my change, so I don't think that my change is related.

### DIFF
--- a/AWSCognitoIdentityProvider/AWSCognitoIdentityUser.m
+++ b/AWSCognitoIdentityProvider/AWSCognitoIdentityUser.m
@@ -330,7 +330,7 @@ static const NSString * AWSCognitoIdentityUserUserAttributePrefix = @"userAttrib
                                                password:(NSString *) password
                                          validationData:(NSArray<AWSCognitoIdentityUserAttributeType*>*) validationData
                                isInitialCustomChallenge:(BOOL) isInitialCustomChallenge {
-    return [self getSession:username password:password validationData:validationData clientMetaData:nil];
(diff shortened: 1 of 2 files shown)


palpatim:
@johntmcintosh 

Thanks for the PR, we'll review & comment

> FYI - The AWSCognitoIdentityProviderTests fail when I run them with the following exception being thrown

Those are integration tests that require valid AWS credentials and a generated test config file. We'll make sure those pass with the suggested change; it's safe for you to disregard that failure for now. The unit tests will automatically run in CircleCI, so we'll be able to ensure those pass.

palpatim (approved):

johntmcintosh:
Thanks @palpatim  -- I had a suspicion the error looked like something that was dependent on some config I hadn't setup. Appreciate the quick feedback 🙌

Issue #3303, opened 2020-12-15
Crash on [cleanUpToDecoderStream] function


florentdouine:
**Describe the bug**
Our major crash is coming from AWS SDK and specially the function 'cleanUpToDecoderStream'. The crash occurs 70% of the time in background on every devices and OS versions. Unfortunately, we are not able to reproduce the crash on our side...

**Stack Trace**
We see different stack strace ending up to the same crash :
````
0  CoreFoundation                 0x1a99d3b38 CFArrayGetCount + 16
1  CoreFoundation                 0x1a9a58df0 _CFStreamClose + 264
2  AWSIoT                         0x1076d7590 -[AWSIoTMQTTClient cleanUpToDecoderStream] + 650 (AWSIoTMQTTClient.m:650)
3  AWSIoT                         0x1076dad28 -[AWSIoTMQTTClient webSocket:didFailWithError:] + 1196 (AWSIoTMQTTClient.m:1196)
4  AWSIoT                         0x107745748 __33-[AWSSRWebSocket _failWithError:]_block_invoke_2 + 774 (AWSSRWebSocket.m:774)
5  libdispatch.dylib              0x1a96b724c _dispatch_call_block_and_release + 32
6  libdispatch.dylib              0x1a96b8db0 _dispatch_client_callout + 20
7  libdispatch.dylib              0x1a96c9a68 _dispatch_root_queue_drain + 656
8  libdispatch.dylib              0x1a96ca120 _dispatch_worker_thread2 + 116
9  libsystem_pthread.dylib        0x1f2edc7c8 _pthread_wqthread + 216
10 libsystem_pthread.dylib        0x1f2ee375c start_wqthread + 8
````

````
Crashed: com.apple.root.default-qos
0  CoreFoundation                 0x19fc41394 CFArrayGetCount + 16
1  CoreFoundation                 0x19fcc1360 _CFStreamClose + 260
2  AWSIoT                         0x1075e7590 -[AWSIoTMQTTClient cleanUpToDecoderStream] + 650 (AWSIoTMQTTClient.m:650)
3  AWSIoT                         0x1075eb198 -[AWSIoTMQTTClient webSocket:didCloseWithCode:reason:wasClean:] + 1234 (AWSIoTMQTTClient.m:1234)
4  AWSIoT                         0x107658754 __41-[AWSSRWebSocket safeHandleEvent:stream:]_block_invoke_2 + 1610 (AWSSRWebSocket.m:1610)
[...]

PR #3487, opened 2021-03-30
fix(iot): Make cleanup decoder stream synchronized

royjit:
*Issue #, if available:*

*Description of changes:*

*Check points:*

- [x] All unit tests pass
- [ ] Updated CHANGELOG.md

### DIFF
--- a/AWSIoT/Internal/AWSIoTMQTTClient.m
+++ b/AWSIoT/Internal/AWSIoTMQTTClient.m
@@ -646,10 +646,15 @@
 }
 
 - (void)cleanUpToDecoderStream {
-    self.toDecoderStream.delegate = nil;
-    [self.toDecoderStream close];
-    [self.toDecoderStream removeFromRunLoop:[NSRunLoop currentRunLoop] forMode:NSDefaultRunLoopMode];
-    self.toDecoderStream = nil;
+    @synchronized (self) {
+        if (self.toDecoderStream == nil) {
+            return;
+        }
+        self.toDecoderStream.delegate = nil;
+        [self.toDecoderStream close];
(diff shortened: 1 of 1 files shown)


palpatim:
Have you been able to repro https://github.com/aws-amplify/aws-sdk-ios/issues/3303 and verify that this fixes it, or is it an optimistic fix for that issue? Are there actually cases where the "cleanup" method would be called simultaneously?

royjit:
> Have you been able to repro #3303 and verify that this fixes it, or is it an optimistic fix for that issue? Are there actually cases where the "cleanup" method would be called simultaneously?

This is an optimistic fix, could not repo #3303 . But running the [sample app](https://github.com/awslabs/aws-sdk-ios-samples/tree/main/IoT-Sample/Swift), I could see that `cleanUpToDecoderStream` being called from different thread (though not simultaneously).

palpatim:
> > Have you been able to repro #3303 and verify that this fixes it, or is it an optimistic fix for that issue? Are there actually cases where the "cleanup" method would be called simultaneously?
> 
> This is an optimistic fix, could not repo #3303 . But running the [sample app](https://github.com/awslabs/aws-sdk-ios-samples/tree/main/IoT-Sample/Swift), I could see that `cleanUpToDecoderStream` being called from different thread (though not simultaneously).

Running the sample app with ThreadSanitizer enabled would be another way to check if there are any races associated with that, but I agree with the intent of the code. As long as we're confident that we aren't introducing any deadlocks by synchronizing that call, this seems like a good change. Since we'd only expect this to be called during teardowns, it should be a minimal impact on performance.

PR #3512, opened 2021-04-13
AWSSTS Update models to latest

awsmobilesdk:
AWSSTS Update models to latest


AWSCore/STS/Info.plist
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
royjit, line 1: Why is this created now?

lawmicha: + @phani-srikar, for visibility

lawmicha: moving this to draft, so to remember not to check this in until we are ready

PR #3517, opened 2021-04-14
revises integration testing script to check for failure conditions

brennanMKE:
*Issue #, if available:*

*Description of changes:*

Adds checks for failure conditions in script used for integration testing. The updated script has been tested locally. It does not change the regular behavior of the script.

*Check points:*

- [ ] ~~Added new tests to cover change, if needed~~
- [x] All unit tests pass
- [ ] ~~Updated CHANGELOG.md~~
- [ ] ~~Documentation update for the change if required~~

### DIFF
--- a/Scripts/generate-test-config.sh
+++ b/Scripts/generate-test-config.sh
@@ -198,6 +198,7 @@ cmd_quiet_flag="--quiet"
 [[ -n $region ]] || die "'region' not specified"
 export AWS_DEFAULT_REGION=${region}
 
+[[ ! -z $(which jq) ]] || die "jq is missing: brew install jq"
 
 ##################################################
 # LOCAL FUNCTIONS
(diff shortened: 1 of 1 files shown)


Scripts/generate-test-config.sh
```diff
@@ -198,6 +198,7 @@ cmd_quiet_flag="--quiet"
 [[ -n $region ]] || die "'region' not specified"
 export AWS_DEFAULT_REGION=${region}
 
+[[ ! -z $(which jq) ]] || die "jq is missing: brew install jq"
```
palpatim, line 201: The [`validate_dependencies` function](https://github.com/aws-amplify/aws-sdk-ios/pull/3517/files#diff-02697d3f7e59bcd8b9d7b6c10f286fbd108a92cb81d544a200f743df66577d3cL205) should already validate this. It's only necessary for the assume-role use case (which you should be using to generate the test config locally), so it isn't appropriate to validate every time. `validate_dependencies` is invoked as the [first step in `main`](https://github.com/aws-amplify/aws-sdk-ios/pull/3517/files#diff-02697d3f7e59bcd8b9d7b6c10f286fbd108a92cb81d544a200f743df66577d3cL298). If we're seeing failures, let's fix that invocation or function logic.

brennanMKE: Yes, it failed for me so I will look at that function to update it.

brennanMKE: I've determined this check will not be needed and removed it in the next commit.

Scripts/generate-test-config.sh
```diff
@@ -253,14 +252,26 @@ function resolve_credentials {
...
-    circleci_execution_role_arn=$(aws ssm get-parameter --name "/mobile-sdk/${platform}/common/circleci_execution_role" | jq -r .Parameter.Value)
+    circleci_execution_role_arn=$(echo "${circleci_execution_role}" | jq -r .Parameter.Value)
     readonly circleci_execution_role_arn
 
     local assume_role_creds
     assume_role_creds=$(aws sts assume-role --duration-seconds 3600 --role-arn "${circleci_execution_role_arn}" --role-session-name "IntegTest-$(date +%Y%m%d%H%M%S)")
     readonly assume_role_creds
 
+    if [ -z "${assume_role_creds}" ]; then
+      die "Failed to assume role credentials for Circle CI"
```
palpatim, line 272: nit: error clarification
```suggestion
      die "Failed to assume Circle CI execution role"
```

palpatim (approved):
Minor error message update, otherwise LGTM

PR #3533, opened 2021-04-16
fix: Fixed base64 decode failed contains special character such as Japanese chars

NivisUnder7:
*Issue #, if available:*
n/a
*Description of changes:*

I found that decoded failed when special character decoded  by `initWithBase64EncodedString:claims` such as Japanese characters. 
Probably same issue #2928 
In other words, current code to get claims from a JWT Token works well as long as there are no special characters in encoded in the base64.
So, that code need to replace _ and - (which are invalid characters in a base64 string) with / and +, respectively.


palpatim:
@NivisUnder7 Thanks for the PR! Can you confirm that you see this, for example, if the username you're using contains encoded chars? Is this tied to an open issue so we can validate that your fix corrects it? If not, can you list some repro steps so we can add some tests (or at least manually test to validate the fix)?

palpatim:
@royjit confirmed that the changes he added in #3623 fail without the change, but pass with this change. Merging

palpatim (approved):

PR #3543, opened 2021-04-20
test(AWSMobileClient): Custom endpoint integ tests

palpatim:
*Check points:*

- [X] Added new tests to cover change, if needed
- [X] All unit tests pass
- [X] All integration tests pass
- [X] Updated CHANGELOG.md
- [X] Documentation update for the change if required - Docs coming on a separate PR in docs repo. CHANGELOG includes usage instructions as well

### DIFF
--- a/AWSAuthSDK/Tests/AWSMobileClientTests/AWSMobileClientTestBase.swift
+++ b/AWSAuthSDK/Tests/AWSMobileClientTests/AWSMobileClientTestBase.swift
@@ -218,19 +218,36 @@ class AWSMobileClientTestBase: XCTestCase {
             return nil
         }.waitUntilFinished()
     }
-    
-    func invalidateSession(username: String) {
+
+    /// Removes the access token's expiration key from the keychain, which forces the SDK into the
+    /// "refresh token expired" flow
+    /// - Parameter username: the username for which to invalidate the token
+    func invalidateRefreshToken(username: String) {
         let bundleID = Bundle.main.bundleIdentifier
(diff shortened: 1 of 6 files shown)


AWSAuthSDK/Tests/AWSMobileClientTests/AWSMobileClientTestBase.swift
```diff
@@ -224,7 +224,11 @@ class AWSMobileClientTestBase: XCTestCase {
         let keychain = AWSUICKeyChainStore(service: "\(bundleID!).\(AWSCognitoIdentityUserPool.self)")
         let namespace = "\(AWSMobileClient.default().userPoolClient!.userPoolConfiguration.clientId).\(username)"
         let key = "\(namespace).tokenExpiration"
-        keychain.removeItem(forKey: key)
+
```
royjit, line 227: I think this invalidation is different, from the existing one. The existing invalidation removes the expiration date so that it falls to the refresh token expire flow. I think what you are trying to do is accesstoken expiry flow. 

Can we split the above method into two:

```
func mockRefreshTokenExpired() -> same as existing code

func mockAccessTokenExpired() -> your change as per this PR

```

palpatim: Ah, you're right, and I should have dug into this more deeply before making the change. Will update.

royjit (approved):
LGTM

PR #3594, opened 2021-05-26
test: UI test for hostedUI refresh token follow

royjit:
*Issue #, if available:*

*Description of changes:*

*Check points:*

- [x] All unit tests pass
- [ ] ~Updated CHANGELOG.md~

### DIFF
--- a/AWSAuthSDK/AWSAuthSDKTestApp/HostedUIUserPoolViewController.swift
+++ b/AWSAuthSDK/AWSAuthSDKTestApp/HostedUIUserPoolViewController.swift
@@ -46,6 +46,12 @@ class HostedUIUserPoolViewController: UIViewController {
         invalidateRefreshToken()
     }
 
+    /// Invalidates the refresh token, the next call to getToken or credentials will return `signedOutUserPoolsTokenInvalid`
+    ///
+    /// AWSCognitoAuth uses refresh token only when the accessToken is expired. So to invalidate refresh token we first need to
+    /// invalidate the access token. After that we assign a invalid dummy value to the refresh token keychain entry. This is done
+    /// to make the refresh token call to Cognito to fail with invalid refresh token error. This invalid refresh token error will trigger
+    /// `signedOutUserPoolsTokenInvalid` status in AWSMobileClient.
     private func invalidateRefreshToken() {
         // Invalidate the access token first, so that the logic will move to refreshing the token
         invalidateAccessToken()
--- a/AWSAuthSDK/AWSAuthSDKTestApp/HostedUIUserPoolViewController.swift
+++ b/AWSAuthSDK/AWSAuthSDKTestApp/HostedUIUserPoolViewController.swift
@@ -47,12 +47,18 @@ class HostedUIUserPoolViewController: UIViewController {
     }
 
     private func invalidateRefreshToken() {
-        let key = getTokenKeychain()
-        getKeychain().removeItem(forKey: key)
+        // Invalidate the access token first, so that the logic will move to refreshing the token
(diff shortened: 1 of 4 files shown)


AWSAuthSDK/AWSAuthSDKTestApp/HostedUIUserPoolViewController.swift
```diff
@@ -47,12 +47,18 @@ class HostedUIUserPoolViewController: UIViewController {
     }
 
     private func invalidateRefreshToken() {
```
lawmicha, line 49: can we add comment to briefly explain this? can link to the SDK code that actually did the check against the expiration of the access token before using the refresh token

AWSAuthSDK/AWSAuthSDKTestApp/UserDetailsViewController.swift
```diff
@@ -43,6 +44,16 @@ class UserDetailsViewController: UIViewController {
         updateUserAttributes()
     }
     
+    func listenUserState() {
+        AWSMobileClient.default().addUserStateListener(self) { (state, additionalInfo) in
+
+            if (state == .signedOutUserPoolsTokenInvalid) {
```
lawmicha, line 50: what if it's another state? could we log the state as an info level log if it is expected to listen to multiple state changes, or log an error level if it's unexpected?

AWSAuthSDK/AWSAuthSDKTestAppUITests/AWSHostedUIUserPoolTests.swift
```diff
@@ -14,7 +14,88 @@ class AWSHostedUIUserPoolTests: AWSAuthSDKUITestBase {
...
+        app.buttons["SignIn User"].tap()
+        signInUserpool(application: app, username: username)
+        
+        // Check if successfully signed in
+        let userPoolSignInStateLabelElement = app.staticTexts["userPoolSignInStateLabel"]
+        let predicate = NSPredicate(format: "label CONTAINS[c] %@", "signedIn")
+        let signInExpectation = expectation(for: predicate,
+                                            evaluatedWith: userPoolSignInStateLabelElement,
+                                            handler: nil)
+        wait(for: [signInExpectation], timeout: 5)
```
lawmicha, line 79: [low] since it's duplicated, can you just call `testHostedUISignIn()`? or probably need to introduce a new private method that returns whatever you need to perform the rest of the test, or pass in what you use later, ie. the username. but if there aren't other scenarios to test and the new private method is just reused once, then can also just keep it the way it is

AWSAuthSDK/AWSAuthSDKTestApp/HostedUIUserPoolViewController.swift
royjit: Added comments

AWSAuthSDK/AWSAuthSDKTestApp/UserDetailsViewController.swift
royjit: This is a hosting app, so doesnot have different log levels. I added a print to log the state so that we can see this in terminal while running the test.

royjit (commented):
Moved the duplicated logic to a private method.

AWSAuthSDK/AWSAuthSDKTestAppUITests/AWSHostedUIUserPoolTests.swift
royjit: Moved the duplicate code

lawmicha (approved):

PR #3638, opened 2021-06-16
feat(Auth): Sign out with revoke token

lawmicha:
*Issue #, if available:*

*Description of changes:*

**AWSCognitoIdentityProvider**
1. Get the AccessToken from the keychain
2. Check if there is a `origin_jti` claim from the Access Token
3. Call revoke token with the refresh token, only if (2) contains the claim
4. Call (3) in an retry expoential back-off 

**AWSMobileClient sign out API**
- HostedUI Sign out API will call revoke token API before performing hosted ui sign out
- UserPool sign out API will call revoke the token. note: It does not care about the invalidateToken sign out option
- no changes to global sign out

**Behavior comparison**
- Android: https://github.com/aws-amplify/aws-sdk-android/pull/2415
- JS: https://github.com/aws-amplify/amplify-js/pull/8418

**Retryability**
In JS, the retry mechanism takes on a similar structure to AppSync's retry logic as
```
// pseudocode
Retryability {
   let baseTime = 100ms
   let maxTime = 5mins
   let jitter = 100ms
   func shouldRetry(statusCode: Int, totalBackOffTime: Int) {
       return statusCode == 5xx || 429 && totalBackoffTime < maxTime
   }
   func backoffAmount(retryCount) -> Double {
     return 2^retryCount * baseTime + jitterAmount()
   }
   private func jitterAmount() {
     return 100ms * random(0, 1)
   }
}
```

iOS also has built in retry logic at https://github.com/aws-amplify/aws-sdk-ios/blob/main/AWSCore/Serialization/AWSURLRequestRetryHandler.m#L53

**Cognito Notes**
[Cognito revoke token docs](https://docs.aws.amazon.com/cognito/latest/developerguide/token-revocation.html)
Checking revoke token functionality is enabled
With your cient id and user pool id (from `awsconfiguration.json` as `AppClientId` and `PoolId`)
```
aws cognito-idp describe-user-pool-client \
    --client-id xxxxxxxxxxxxxxxx \
    --user-pool-id [us-east-1_xxxxxx]
```
should display *EnableTokenRevocation* to true
```
{
    "UserPoolClient": {
        "UserPoolId": "us-east-1_xxx",
        "ClientName": "xxx",
        "ClientId": "xxx",
        "LastModifiedDate": "2021-06-30T12:36:36.347000-07:00",
        "CreationDate": "2021-06-23T10:29:40.936000-07:00",
        "RefreshTokenValidity": 30,
        "TokenValidityUnits": {},
        "AllowedOAuthFlowsUserPoolClient": false,
        "EnableTokenRevocation": true
    }
}
```
If it doesn't and you have an previous provision Cognito resource, enable using
```
aws cognito-idp update-user-pool-client \
    --client-id xxxxx \
    --user-pool-id [xx-xxxx-xxxx] \
    --enable-token-revocation \
```

- [ ] add additional check on asserting that token cannot be reused, ie. calling updateAttribtes or global sign out

**Testing:**
- [x] Update integration test provisioning with revoke token enable
- [x] re-run existing tests and new tests added which should have test coverage 
- [x] run integ test using pipeline: https://github.com/aws-amplify/aws-sdk-ios/pull/3679
- [x] Sample app testing: HostedUI
- [x] Sample app testing: User Pools
- [x] Sample app testing: Federated Sign in

*Check points:*

- [x] All unit tests pass
- [x] Updated CHANGELOG.md (https://github.com/aws-amplify/aws-sdk-ios/pull/3680)
- [x] Documentation update for the change if required (https://github.com/aws-amplify/docs/pull/3379)

### DIFF
--- a/AWSCognitoIdentityProvider/AWSCognitoIdentityUser.m
+++ b/AWSCognitoIdentityProvider/AWSCognitoIdentityUser.m
@@ -1488,6 +1488,28 @@ static const NSString * AWSCognitoIdentityUserUserAttributePrefix = @"userAttrib
     return refreshToken != nil;
 }
 
+- (AWSTask<AWSCognitoIdentityProviderRevokeTokenResponse *> *) revokeToken {
+    __block NSString * keyChainNamespace = [self keyChainNamespaceClientId];
+    NSString * accessTokenKey = [self keyChainKey:keyChainNamespace key:AWSCognitoIdentityUserAccessToken];
+    NSString * accessTokenString = self.pool.keychain[accessTokenKey];
+    AWSCognitoIdentityUserSessionToken * accessToken = [[AWSCognitoIdentityUserSessionToken alloc] initWithToken:accessTokenString];
+    if ([accessToken.tokenClaims objectForKey:@"origin_jti"] == nil) {
+        AWSDDLogVerbose(@"Access Token does not contain `origin_jti` claim. Skip revoking tokens.");
+        return [AWSTask taskWithResult:nil];
+    }
+    
+    AWSCognitoIdentityProviderRevokeTokenRequest *request = [AWSCognitoIdentityProviderRevokeTokenRequest new];
+    NSString * refreshToken = [self refreshTokenFromKeyChain:keyChainNamespace];
+    request.token = refreshToken;
+    request.clientId = self.pool.userPoolConfiguration.clientId;
+    request.clientSecret = self.pool.userPoolConfiguration.clientSecret;
+    
+    return [[self.pool.client revokeToken:request] continueWithSuccessBlock:^id _Nullable(AWSTask<AWSCognitoIdentityProviderRevokeTokenResponse *> * _Nonnull task) {
+        AWSDDLogVerbose(@"Revoke token completed.");
+        return task;
+    }];
+}
+
 - (AWSTask<AWSCognitoIdentityUserGlobalSignOutResponse *> *) globalSignOut {
     AWSCognitoIdentityProviderGlobalSignOutRequest *request = [AWSCognitoIdentityProviderGlobalSignOutRequest new];
     return [[self getSession] continueWithSuccessBlock:^id _Nullable(AWSTask<AWSCognitoIdentityUserSession *> * _Nonnull task) {
--- a/AWSCognitoIdentityProvider/AWSCognitoIdentityUser.m
+++ b/AWSCognitoIdentityProvider/AWSCognitoIdentityUser.m
@@ -1488,26 +1488,22 @@ static const NSString * AWSCognitoIdentityUserUserAttributePrefix = @"userAttrib
     return refreshToken != nil;
 }
 
-- (AWSTask<AWSCognitoIdentityProviderRevokeTokenResponse *> *) revokeToken {
+-(BOOL) isSessionRevocable {
     __block NSString * keyChainNamespace = [self keyChainNamespaceClientId];
     NSString * accessTokenKey = [self keyChainKey:keyChainNamespace key:AWSCognitoIdentityUserAccessToken];
     NSString * accessTokenString = self.pool.keychain[accessTokenKey];
     AWSCognitoIdentityUserSessionToken * accessToken = [[AWSCognitoIdentityUserSessionToken alloc] initWithToken:accessTokenString];
-    if ([accessToken.tokenClaims objectForKey:@"origin_jti"] == nil) {
-        AWSDDLogVerbose(@"Access Token does not contain `origin_jti` claim. Skip revoking tokens.");
-        return [AWSTask taskWithResult:nil];
-    }
-    
+    return [accessToken.tokenClaims objectForKey:@"origin_jti"];
+}
+
+- (AWSTask<AWSCognitoIdentityProviderRevokeTokenResponse *> *) revokeToken {
+    __block NSString * keyChainNamespace = [self keyChainNamespaceClientId];
     AWSCognitoIdentityProviderRevokeTokenRequest *request = [AWSCognitoIdentityProviderRevokeTokenRequest new];
     NSString * refreshToken = [self refreshTokenFromKeyChain:keyChainNamespace];
     request.token = refreshToken;
     request.clientId = self.pool.userPoolConfiguration.clientId;
     request.clientSecret = self.pool.userPoolConfiguration.clientSecret;
-    
-    return [[self.pool.client revokeToken:request] continueWithSuccessBlock:^id _Nullable(AWSTask<AWSCognitoIdentityProviderRevokeTokenResponse *> * _Nonnull task) {
-        AWSDDLogVerbose(@"Revoke token completed.");
-        return task;
-    }];
+    return [self.pool.client revokeToken:request];
 }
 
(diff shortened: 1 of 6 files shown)


AWSCognitoIdentityProvider/AWSCognitoIdentityUser.m
```diff
@@ -1488,6 +1488,28 @@ -(BOOL) isSignedIn {
     return refreshToken != nil;
 }
 
+- (AWSTask<AWSCognitoIdentityProviderRevokeTokenResponse *> *) revokeToken {
+    __block NSString * keyChainNamespace = [self keyChainNamespaceClientId];
+    NSString * idTokenString = [self keyChainKey:keyChainNamespace key:AWSCognitoIdentityUserIdToken];
+    AWSCognitoIdentityUserSessionToken * idToken = [[AWSCognitoIdentityUserSessionToken alloc] initWithToken:idTokenString];
+    
+    if ([idToken.tokenClaims objectForKey:@"jti"] == nil) {
```
lawmicha, line 1496: "origin_jti"

AWSCognitoIdentityProvider/AWSCognitoIdentityUser.m
```diff
@@ -1488,6 +1488,28 @@ -(BOOL) isSignedIn {
     return refreshToken != nil;
 }
 
+- (AWSTask<AWSCognitoIdentityProviderRevokeTokenResponse *> *) revokeToken {
+    __block NSString * keyChainNamespace = [self keyChainNamespaceClientId];
+    NSString * idTokenString = [self keyChainKey:keyChainNamespace key:AWSCognitoIdentityUserIdToken];
+    AWSCognitoIdentityUserSessionToken * idToken = [[AWSCognitoIdentityUserSessionToken alloc] initWithToken:idTokenString];
```
lawmicha, line 1494: use the accessToken here because hostedUI

div5yesh:
PR for docs content change for token revocation: aws-amplify/docs#3379

AWSCognitoIdentityProvider/AWSCognitoIdentityUser.m
```diff
@@ -1488,6 +1488,28 @@ -(BOOL) isSignedIn {
...
 }
 
+- (AWSTask<AWSCognitoIdentityProviderRevokeTokenResponse *> *) revokeToken {
+    __block NSString * keyChainNamespace = [self keyChainNamespaceClientId];
+    NSString * accessTokenKey = [self keyChainKey:keyChainNamespace key:AWSCognitoIdentityUserAccessToken];
+    NSString * accessTokenString = self.pool.keychain[accessTokenKey];
+    AWSCognitoIdentityUserSessionToken * accessToken = [[AWSCognitoIdentityUserSessionToken alloc] initWithToken:accessTokenString];
+    if ([accessToken.tokenClaims objectForKey:@"origin_jti"] == nil) {
+        AWSDDLogVerbose(@"Access Token does not contain `origin_jti` claim. Skip revoking tokens.");
+        return [AWSTask taskWithResult:nil];
```
royjit, line 1498: Should this be an error? It would be better to send either a success object or an error back from this api. With the current code we are passing three types of response:
1. nil (if origin_jti not present)
2. AWSCognitoIdentityProviderRevokeTokenResponse (if api succeeded)
3. Error (if api failed)

AWSCognitoIdentityProvider/AWSCognitoIdentityUser.m
```diff
@@ -1488,6 +1488,28 @@ -(BOOL) isSignedIn {
...
+    AWSCognitoIdentityProviderRevokeTokenRequest *request = [AWSCognitoIdentityProviderRevokeTokenRequest new];
+    NSString * refreshToken = [self refreshTokenFromKeyChain:keyChainNamespace];
+    request.token = refreshToken;
+    request.clientId = self.pool.userPoolConfiguration.clientId;
+    request.clientSecret = self.pool.userPoolConfiguration.clientSecret;
+    
+    return [[self.pool.client revokeToken:request] continueWithSuccessBlock:^id _Nullable(AWSTask<AWSCognitoIdentityProviderRevokeTokenResponse *> * _Nonnull task) {
+        AWSDDLogVerbose(@"Revoke token completed.");
+        return task;
+    }];
```
royjit, line 1510:
```suggestion
    return [self.pool.client revokeToken:request];
```

Does this work?

AWSCognitoIdentityProvider/AWSCognitoIdentityUser.m
lawmicha: this is client side validation for backwards compatible tokens which do not have the revocation feature enabled, it  is also used to avoid an extra API call to Cognito. If the token doesn't contain "origin_jti" and we continue to call Cognito with revoke token, we'll get an error response indicating that the feature is not enabled. If the whole purpose is to avoid the network call then we can return the same response as if we made the service call.

lawmicha: we can also separate out the API into "shouldRevoke -> Bool" and "revokeToken", which would reduce the response down to AWSCognitoIdentityProviderRevokeTokenResponse and Error.

lawmicha: so we can 
1. keep it as is, return nil on success which indiciates that this revokeToken API did no-op
2. keep as is but return an error, to closely align with what the service would return
3. separate out the APis into "isSessionRevocable" and "revokeToken". this dumbs down this revokeToken as just an API call directly to the LLC revoke token API

AWSCognitoIdentityProvider/AWSCognitoIdentityUser.m
lawmicha: yup this works, the logging seems redundant, taking ur suggestion

AWSCognitoIdentityProvider/AWSCognitoIdentityUser.m
lawmicha: as discusssed, our preference is for 3, and to see if we are happy with the calling code in awsmobileclient

AWSCognitoIdentityProvider/AWSCognitoIdentityUser.m
```diff
@@ -1488,6 +1488,24 @@ -(BOOL) isSignedIn {
...
+-(BOOL) isSessionRevocable {
+    NSString * keyChainNamespace = [self keyChainNamespaceClientId];
+    NSString * accessTokenKey = [self keyChainKey:keyChainNamespace key:AWSCognitoIdentityUserAccessToken];
+    NSString * accessTokenString = self.pool.keychain[accessTokenKey];
+    AWSCognitoIdentityUserSessionToken * accessToken = [[AWSCognitoIdentityUserSessionToken alloc] initWithToken:accessTokenString];
+    return [accessToken.tokenClaims objectForKey:@"origin_jti"];
+}
+
+- (AWSTask<AWSCognitoIdentityProviderRevokeTokenResponse *> *) revokeToken {
+    __block NSString * keyChainNamespace = [self keyChainNamespaceClientId];
```
royjit, line 1500: Can we remove this `__block` as well?

royjit (approved):
LGTM, one minor comment

royjit (approved):

Issue #3663, opened 2021-07-02
Limit change to Blob handling to Location Service


brennanMKE:
**Describe the bug**

A recent change to Blob handling has caused changes which break integration tests.

**To Reproduce**
1. Run integration tests for AWSLambda
2. See failed tests

diegocstn:
Fix released with [v.24.3](https://github.com/aws-amplify/aws-sdk-ios/releases/tag/2.24.3)

PR #3664, opened 2021-07-02
fix: fixes Blob handling by limiting change to Location (#3663)

brennanMKE:
*Issue #, if available:*

#3663

*Description of changes:*

Fixes integration tests which started to fail due to how Blob serialization is handled. Now this change in behavior is limited to Location Service which is a new AWS Service while other services will continue with the same behavior.

*Check points:*

- [x] All unit tests pass
- [x] Updated CHANGELOG.md


lawmicha (approved):

ameter (approved):
I confirmed that this fixed the broken AWSLocation calls.

PR #3732, opened 2021-08-20
fix(AWSMobileClient): Reset completion sources after completion

dnys1:
*Issue #, if available:*

Amplify iOS crashes when confirmSignIn is called twice in one app session (https://github.com/aws-amplify/amplify-ios/issues/1376). The issue is that the old `AWSTaskCompletionSource` is not cleared in `AWSMobileClient` and the second request tries to re-complete the first one.

*Description of changes:*
- Reset completion sources in `AWSMobileClient` upon task completion

*Check points:*

- [ x] Added new tests to cover change, if needed
- [ x] All unit tests pass
- [ x] All integration tests pass
- [ x] Updated CHANGELOG.md
- [ x] Documentation update for the change if required
- [ x] PR title conforms to conventional commit style


AWSAuthSDK/Tests/AWSMobileClientTests/AWSMobileClientSignInTests.swift
```diff
@@ -125,6 +125,66 @@ class AWSMobileClientSignInTests: AWSMobileClientTestBase {
...
+                guard let signInResult = signInResult else {
+                    XCTFail("Confirm sign in failed, signInResult unexpectedly nil")
+                    return
+                }
+                
+                XCTAssert(signInResult.signInState == .signedIn)
+                XCTAssertNil(AWSMobileClient.default().userpoolOpsHelper.currentConfirmSignInHandlerCallback,
+                                "Current confirmsign callback should be nil")
+                XCTAssertNil(AWSMobileClient.default().userpoolOpsHelper.newPasswordRequiredTaskCompletionSource,
+                                "New password completion source should be nil")
```
diegocstn, line 182: instead of just checking the values of the those tasks, can we also verify that calling `AWSMobileClient.default().confirmSignIn` twice doesn't cause a crash anymore?

dnys1: Absolutely, and good thinking. I've updated the test to perform the createUser/confirmSignIn flow twice, which is the use case for which we experienced the crash originally. Thanks!

royjit (dismissed):

royjit (approved):

royjit (approved):

PR #3752, opened 2021-08-31

awsmobilesdk:


lawmicha (changes requested):
contains breaking change

PR #3763, opened 2021-09-06
fix: Handle special characters in AWSSignatureV4Signer

ameter:
*Issue #3215*

*Description of changes:*
Fixes encoding issue in AWSSignatureV4Signer to properly handle special characters in URLs.  SigV4 requires special characters to be double encoded.

*Check points:*

- [x] All unit tests pass
- [x] Updated CHANGELOG.md


AWSCoreUnitTests/SigV4Tests/SigV4TestCases.swift
```diff
@@ -1604,6 +1633,108 @@ signature:
...
+credentialsProvider: SigV4TestCredentials.basicTestCredentials,
+
+shouldSignBody: true,
+shouldSignSecurityToken: true,
+serviceName: "s3",
+
+// When
+originalRequest:
+"""
+GET /a%3Ab/%26%24%40%3D%3B%3A%2B%20%2C%3F.zip HTTP/1.1
```
palpatim, line 1653: So this is expecting the incoming URL to be escaped already, right? I assume that's part of the contract to invoke the SigV4 URL signer/Presigned URL generator? We could double/triple check that by including a PresignedURLBuilder test on a key like `😄.jpg`, if we don't already have one.

AWSCoreUnitTests/SigV4Tests/SigV4TestCase.swift
```diff
@@ -28,6 +28,7 @@ struct SigV4TestCase {
     let credentialsProvider: AWSCredentialsProvider
     let shouldSignBody: Bool
     let shouldSignSecurityToken: Bool
+    let serviceName: String
```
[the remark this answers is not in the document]
ameter: I considered this, but put everything in the individual test cases to be consistent with the previous design (i.e. `shouldSignBody` and `shouldSignSecurityToken` were specified in each test case, even though they were false in most cases).  I changed it to default `serviceName` to `"service"` and also to default `shouldSignBody` and `shouldSignSecurityToken` to `false` for consistency.

AWSCoreUnitTests/SigV4Tests/SigV4TestCases.swift
ameter: No, there is no requirement or expectation that the incoming URL be escaped already.  In order to support it either being previously encoded or not, we strip the encoding (once for s3 and twice for everything else) before we re-encode it (again, once for s3 and twice for everything else).

The reason that it's encoded in the test cases is because of the test suite design, where the passed in request is deserialized from a string.  We could change the testing setup, but in each case the encoding is still stripped and re-applied anyway.

palpatim: OK. Since this is testing the lower-level encoding routines, your explanation of the test structure makes sense.

It looks like we have a [special character test for the presigned URL builder](https://github.com/aws-amplify/aws-sdk-ios/blob/main/AWSS3Tests/AWSS3PreSignedURLBuilderTests.m#L416) but it doesn't include characters outside ASCII range. It would be good to add a case with Unicode/non-ASCII characters to that test to make sure we're testing the whole path. My goal here is to validate that we aren't adding any surprising behavior to the customer-facing APIs, even though you're proving the low-level behavior is right with these tests.

ameter: I added integration test resources for amazon location ([corresponding amplify-ci-support PR](https://github.com/aws-amplify/amplify-ci-support/pull/99)) and integration tests for the map requests that can contain special characters.  I also added a 😀 (non-ascii) character to the S3 testPutObjectWithSpecialCharacters test.  All relevant tests pass.

palpatim (approved):

PR #3681, opened 2021-07-14
feat(AWSIoT): support for retained messages

brennanMKE:
*Issue #, if available:*

Do not review or merge. PR is being used to run integration tests with CircleCI.

*Description of changes:*

*Check points:*

- [x] Added new tests to cover change, if needed
- [x] All unit tests pass
- [x] All integration tests pass
- [x] Updated CHANGELOG.md
- [x] Documentation update for the change if required
- [x] PR title conforms to conventional commit style

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3782, opened 2021-09-21

awsmobilesdk:


hamoody0706 (approved):
AWSIoT/AWSIoTResources.m

lawmicha (changes requested):
Not merging this in until 
1. verify that this change is correct. It was added as part of https://github.com/aws-amplify/aws-sdk-ios/pull/3681/files#diff-30ac9a3837d3485359acd899c68ded472ce29cd1f14dd07a47e77edea6fe8348 @brennanMKE 
2. Incorporate the change to the model update process so that the change will persist and not be overridden here.
3. short term- can we manually fix this model update PR. long term- one 2. is done then new model updates will be fine

Issue #3851, opened 2021-10-29
IoT: Add back Message as a public type


brennanMKE:
This type should be a part of the public IoT API.

lawmicha:
this is related to https://github.com/aws-amplify/aws-sdk-ios/pull/3782

lawmicha:
PR has been merged and pending release

PR #3852, opened 2021-10-29
fix(IoT): adds Message as public type

brennanMKE:
*Issue #, if available:*

#3851

*Description of changes:*

* Adds Message type to umbrella header for the framework.

*Check points:*

- [x] All unit tests pass
- [x] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ MOVE THIS SECTION TO THE CORRECT LOCATION
 -Features for next release
 
 - **AWSIoT**
-  - feat(AWSIoT): adds back Message type ([PR #3852](https://github.com/aws-amplify/aws-sdk-ios/pull/3852))
+  - fix(AWSIoT): adds back Message type ([PR #3852](https://github.com/aws-amplify/aws-sdk-ios/pull/3852))
 
 ## 2.26.3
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ MOVE THIS SECTION TO THE CORRECT LOCATION
 
 -Features for next release
 
+- **AWSIoT**
+  - feat(AWSIoT): adds back Message type ([PR #3852](https://github.com/aws-amplify/aws-sdk-ios/pull/3852))
```
lawmicha, line 12: shold this be a fix?

lawmicha:
`feat(IoT) adds Message as public type #3852` should have `:` colon between "feat(IoT)" and the description like
`feat(IoT): adds Message as public type #3852`. i guess it's not very important here since the CI/CD pipeline doesn't use the commits to determine the version

CHANGELOG.md
brennanMKE: I looked and did not see it breaking without this header included. The individual header is public so it is included as a public header in the framework so it has always been available since it was added. I believe this is not as much a fix as it is exposing a public type. Since the integration tests did not fail without this header I think we can call this one `feat` since it is exposing the type for the feature it supports.

brennanMKE: Just tried it from Swift. It breaks the build if you use the subscribe method which needs this type. I will change to `fix` and update CHANGELOG.md. On the Objective-C side it can build and run all tests without a problem.

lawmicha (approved):

lawmicha:
merging this in @brennanMKE let me know if you have any concerns

Issue #3871, opened 2021-11-11
Eliminate blocking code in block operations


brennanMKE:
**Describe the bug**

There are Block Operations being added to OperationQueues which use DispatchGroup to `enter()` and `leave()` with `wait()` which blocks the thread while async work is done. This work may include network activity which introduces latency while the thread is blocked. There has been evidence of thread explosion which lead to the [getAWSCredentials](https://github.com/aws-amplify/aws-sdk-ios/blob/main/AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClientExtensions.swift#L765) method. This code is making the block wait on the current thread to change this block to use serial behavior so it does not leave the block until the work is completed.

An Operation transitions between 4 states: ready, executing, finished and cancelled. A Block Operation will be set to `executing` when `start()` is called on it and when the execution leaves the block it is set to `finished` allowing the OperationQueue to start other operations. This update introduces `AsyncBlockOperation` which does not block the thread and instead provides a `done()` closure to the block which is called once the async work is completed. It then changes the state to `finished` and notifies of the state change via KVO.

Testing is showing that it reduces the number of threads which are created while running many operations. Since the update operations are done for the purposes of auth it impacts every service which uses auth as well as Amplify Plugins.

See: [Asynchronous Versus Synchronous Operations](https://developer.apple.com/documentation/foundation/operation#1661231)

> If you always plan to use queues to execute your operations, it is simpler to define them as synchronous. If you execute operations manually, though, you might want to define your operation objects as asynchronous. Defining an asynchronous operation requires more work, because you have to monitor the ongoing state of your task and report changes in that state using KVO notifications. But defining asynchronous operations is useful in cases where you want to ensure that a manually executed operation does not block the calling thread.

**To Reproduce**

See [amplify-ios#1279](https://github.com/aws-amplify/amplify-ios/issues/1279) for a customer issue and a sample app which repros the issue.

**Observed Behavior**

Due to threads being blocked there is a thread explosion when there are many operations running concurrently which makes calls into `getIdentityId` which leads to this code in AWSMobileClient.

**Expected Behavior**

Threads should not be blocked when the async work can be done without blocking.

**Stack Trace**
Please provide a stack trace if applicable e.g. a crash is occurring.

See the linked issue for logs collected in the sample project.

royjit:
This is implemented in as per this PR - https://github.com/aws-amplify/aws-sdk-ios/pull/3872

PR #3872, opened 2021-11-11
bug(auth): introduces AsyncBlockOperation to eliminate thread locking in operation queues

brennanMKE:
*Issue #, if available:*

#3871 

*Description of changes:*

Changes a couple of Block Operations into Async Block Operations.

*Check points:*

- [ ] All unit tests pass
- [ ] Updated CHANGELOG.md

### DIFF
--- a/AWSAuthSDK/Sources/AWSMobileClient/AWSAsyncBlockOperation.swift
+++ b/AWSAuthSDK/Sources/AWSMobileClient/AWSAsyncBlockOperation.swift
@@ -1,0 +1,77 @@
+import Foundation
+
+enum AsyncBlockState: String {
+    case ready = "isReady"
+    case executing = "isExecuting"
+    case finished  = "isFinished"
+    case cancelled = "isCancelled"
+}
+
+typealias AsyncBlockDoneClosure = () -> Void
+typealias AsyncBlockClosure = (@escaping AsyncBlockDoneClosure) -> Void
+
+class AsyncBlockOperation: Operation {
+    let asyncBlock: AsyncBlockClosure
+
+    var state: AsyncBlockState = .ready
+
+    // handle KVO events before changing state
+    func transition(to newState: AsyncBlockState) {
+        guard state != newState else { return }
+
+        willChangeValue(forKey: newState.rawValue)
+        willChangeValue(forKey: state.rawValue)
+
+        state = newState
+
+        didChangeValue(forKey: newState.rawValue)
+        didChangeValue(forKey: state.rawValue)
+    }
+
+    override var isReady: Bool {
+        state == .ready
+    }
+
+    override var isExecuting: Bool {
+        state == .executing
+    }
+
+    override var isFinished: Bool {
+        state == .finished
+    }
+
+    override var isCancelled: Bool {
+        state == .cancelled
+    }
+
+    init(asyncBlock: @escaping AsyncBlockClosure) {
+        self.asyncBlock = asyncBlock
+        super.init()
+    }
+
+    override var isAsynchronous: Bool { true }
+
+    override func start() {
+        guard !isCancelled && isReady else {
+            return
+        }
+        main()
+    }
+
+    override func main() {
+        guard !isCancelled else { return }
+
+        transition(to: .executing)
+
+        let done: AsyncBlockDoneClosure = { [weak self] in
+            guard let self = self else { fatalError() }
+            self.transition(to: .finished)
+        }
+
+        asyncBlock(done)
+    }
+
+    override func cancel() {
+        transition(to: .cancelled)
+    }
+}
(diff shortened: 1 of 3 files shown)


AWSAuthSDK/Sources/AWSMobileClient/AWSAsyncBlockOperation.swift
```diff
@@ -0,0 +1,70 @@
...
+    case ready = "isReady"
+    case executing = "isExecuting"
+    case finished  = "isFinished"
+    case cancelled = "isCancelled"
+}
+
+public typealias AWSAsyncBlockDoneClosure = () -> Void
+public typealias AWSAsyncBlockClosure = (@escaping AWSAsyncBlockDoneClosure) -> Void
+
+public class AWSAsyncBlockOperation: Operation {
```
lawmicha, line 13: This looks similar to https://github.com/aws-amplify/amplify-ios/blob/main/Amplify/Core/Support/AsychronousOperation.swift which was introduced in Amplify iOS for operations that require to do some async work. I'm thinking
- If the intent of the two classes are pretty much the same, can we review `AsynchronousOperation` with your learnings and identify if there are any changes we should do to the one in Amplify? Not blocking this PR, just can keep an issue open for the changes we want to make to consolate the two in terms of behavior and usability, as we want to avoid breaking changes unless it's really needed (performance improvement, incorrect usage of subclassing Operation)
- Should `AWSAsyncBlockOperation` be in AWSCore with some unit tests? I'm guessing this can be reused in other places like TransferUtility or Pinpoint SDK if the same code patterns were applied there

brennanMKE: We discussed this during a meeting recently. This type is similar but is also in AWS Mobile Client and built in Swift code. AWS Core is Objective-C and this module is a dependency of Amplify Auth. AsynchronousOperation also does a bit more than this type does which includes integration with other Amplify systems including Combine features. AsyncBlockOperation does as little as possible to integrate properly with an `OperationQueue` based on Apple's docs to eliminate the need for any blocking to free up threads. The 2 key mechanisms are the KVO notifications and the `done` block which moves to the finished state so the next `Operation` can start.

brennanMKE:
Revised to set `pendingAWSCredentialsCompletion` with a closure so that `done()` is called and the use of the `DispatchGroup` for locking was eliminated completely. It works well with the local dev app.

CHANGELOG.md
```diff
@@ -11,10 +11,17 @@ MOVE THIS SECTION TO THE CORRECT LOCATION
...
+
+- **AWSMobileClient**
+  - fix(AWSMobileClient): eliminates thread blocking in operations ([PR #3872](https://github.com/aws-amplify/aws-sdk-ios/pull/3872))
 
 ## 2.26.5
 
+- **AWSCore**
+  - Support for `ap-southeast-3` - Asia Pacific (Jakarta) (see [AWS Regional Services List](https://aws.amazon.com/about-aws/global-infrastructure/regional-product-services/) for a list of services supported in the region)
+
+- Return Swift optionals from Objective C methods that can return nil
```
lawmicha, line 23:
```suggestion
- Return Swift optionals from Objective C methods that can return nil ([PR #3912](https://github.com/aws-amplify/aws-sdk-ios/pull/3912))

```

CircleciScripts/create_xcframeworks.py
```diff
@@ -68,6 +68,14 @@ def map_framework_to_project(framework_list):
...
 # Archive all the frameworks.
 for framework in xcframeworks:
+    xcframework = f"{XCFRAMEWORK_PATH}{framework}.xcframework"
+
+    if os.path.exists(xcframework):
+        log(f"skipping {framework}...")
+        continue
+    
+    log(f"Creating archives for {framework}")
+
```
lawmicha, line 78: despite this being a small-ish change, it would be better to cut a new PR for this change to keep it isolated. unless i'm incorrect about this being unrelated.

lawmicha (changes requested):

.gitignore
```diff
@@ -1,3 +1,4 @@
+xcframeworks
```
lawmicha, line 1: nit: put under one of the headings, maybe **Build assets**

brennanMKE: I tried to remove this commit. It should not be in this branch. I will try again. GitHub is not allowing to to modify the branch and force push like I've done before. Normally I can clean up a branch.

CHANGELOG.md
brennanMKE: I've rebuilt the branch and used fixup to make it a single commit without updates to CHANGELOG.md to avoid the merge conflict. now it is just the code changes and the Xcode project file.

CircleciScripts/create_xcframeworks.py
brennanMKE: I've removed this unrelated change.

lawmicha (approved):
LGTM

sagarjoshi:
When will this be part of a release ?

PR #3913, opened 2021-11-27
fix: Return non nil credentials back from CredentialsProvider

royjit:
*Issue #, if available:* https://github.com/aws-amplify/amplify-ios/issues/1501

*Description of changes:* 
In `AWSCredentialsProvider` class `self.credentials` gets the credentials values directly from the keychain. So these value can change underneath while we try to access it. This change holds on to a copy of the credentials and returns them to use to make sure that a valid result is returned.

*Check points:*

- [x] All unit tests pass
- [ ] Updated CHANGELOG.md

### DIFF
--- a/AWSCore/Authentication/AWSCredentialsProvider.m
+++ b/AWSCore/Authentication/AWSCredentialsProvider.m
@@ -95,6 +95,15 @@ static NSString *const AWSCredentialsProviderKeychainIdentityId = @"identityId";
             self.expiration];
 }
 
+- (nonnull id)copyWithZone:(nullable NSZone *)zone {
+    AWSCredentials *object = [[AWSCredentials alloc] initWithAccessKey:self.accessKey.copy
+                                                             secretKey:self.secretKey.copy
+                                                            sessionKey:self.sessionKey.copy
+                                                            expiration:self.expiration.copy];
+    
+    return object;
+}
+
 @end
 
 @interface AWSStaticCredentialsProvider()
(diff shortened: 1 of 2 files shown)


AWSCore/Authentication/AWSCredentialsProvider.m
```diff
@@ -636,10 +647,11 @@ - (void)setUpWithRegionType:(AWSRegionType)regionType
             // 1. The cached logins are different from the one the identity provider provided.
             // 2. The cached credentials is nil.
             // 3. The credentials expire within 10 minutes.
+            credentials = self.internalCredentials.copy;
```
harsh62, line 650: What is the reason you are assigning it back? Can you give a brief description?

royjit: I am just holding on to the value of `self.internalCredentials` inside a local variable so that we will have a valid value  even if the class variable has changed.

harsh62 (dismissed):

AWSCore/Authentication/AWSCredentialsProvider.m
```diff
@@ -600,12 +609,14 @@ - (void)setUpWithRegionType:(AWSRegionType)regionType
...
+    
+    __block AWSCredentials *credentials = self.internalCredentials.copy;
     // Returns cached credentials when all of the following conditions are true:
     // 1. The cached credentials are not nil.
     // 2. The credentials do not expire within 10 minutes.
-    if (self.internalCredentials
-        && [self.internalCredentials.expiration compare:[NSDate dateWithTimeIntervalSinceNow:10 * 60]] == NSOrderedDescending) {
-        return [AWSTask taskWithResult:self.internalCredentials];
+    if (credentials
+        && [credentials.expiration compare:[NSDate dateWithTimeIntervalSinceNow:10 * 60]] == NSOrderedDescending) {
```
brennanMKE, line 618: This could be done with a computed property which can be tested and make the context more clear. First, add the property declaration to the class extension.

```objc
@interface AWSCredentials()

@property (readonly) BOOL credentialsExpired;

@end
```

Then implement the method.

```objc
- (BOOL)credentialsExpired {
    return (self.credentials &&
        [self.credentials.expiration compare:[NSDate dateWithTimeIntervalSinceNow:10 * 60]] == NSOrderedDescending);
}
```

In the tests code the property will have to be declared again to make it accessible. It appears this property could be used in more that 1 place in this source file.

royjit: Added the helper method

brennanMKE (approved):
👍

PR #3948, opened 2021-12-23
feat(AWSMobileClient): add deleteUser API

ameter:
*Description of changes:*
Add deleteUser API to AWSMobileClient to support compliance with this [Apple requirement](https://developer.apple.com/news/?id=mdkbobfo).

*Check points:*

- [x] All unit tests pass
- [x] Updated CHANGELOG.md

### DIFF
--- a/AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClientExtensions.swift
+++ b/AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClientExtensions.swift
@@ -672,7 +672,7 @@ extension AWSMobileClient {
         let _ = self.userpoolOpsHelper.currentActiveUser!.delete().continueWith { (task) -> Any? in
             if task.result != nil {
                 // If global signout is successful, we clear tokens locally and perform signout flow.
-                self.signOut()
+                //self.signOut()
                 completionHandler(nil)
             } else if let error = task.error {
                 // If there is an error signing out globally, we notify the developer.
--- a/AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClientExtensions.swift
+++ b/AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClientExtensions.swift
@@ -668,14 +668,19 @@ extension AWSMobileClient {
     ///
     /// - Parameters:
     ///   - completionHandler: completion handler for success or error callback.
-    public func deleteUser(completionHandler: @escaping ((Error?) -> Void)) {
+    public func deleteUser(signOut: Bool = true, completionHandler: @escaping ((Error?) -> Void)) {
         let _ = self.userpoolOpsHelper.currentActiveUser!.delete().continueWith { (task) -> Any? in
             if task.result != nil {
-                // If global signout is successful, we clear tokens locally and perform signout flow.
-                //self.signOut()
-                completionHandler(nil)
+                // User was successfully deleted.
+                if signOut {
+                    // If signOut is true (default), perform global signout and invalidate tokens.
+                    let signOutOptions = SignOutOptions(signOutGlobally: true, invalidateTokens: true)
+                    self.internalSignOut(options: signOutOptions, completionHandler: completionHandler)
+                } else {
+                    completionHandler(nil)
+                }
             } else if let error = task.error {
-                // If there is an error signing out globally, we notify the developer.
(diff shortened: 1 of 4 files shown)


AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClientExtensions.swift
```diff
@@ -664,6 +664,30 @@ extension AWSMobileClient {
         self.userPoolClient?.clearAll()
     }
     
+    /// Asynchronous deleteUser method which requires network activity.
+    ///
+    /// - Parameters:
+    ///   - completionHandler: completion handler for success or error callback.
+    public func deleteUser(signOut: Bool = true, completionHandler: @escaping ((Error?) -> Void)) {
```
brennanMKE, line 671: nit: I see force unwrapping done here and elsewhere so it appears to be common with this extension. It would be better to guard and call `fatalError` with a useful message if this value is ever nil.

AWSAuthSDK/Tests/AWSMobileClientTests/AWSMobileClientDeleteUserTests.swift
```diff
@@ -0,0 +1,63 @@
...
+    ///    - I invoke `deleteUser` with completion callback
+    /// - Then:
+    ///    - My `deleteUser` completion callback is invoked with a AWSMobileClientError.notSignedIn error.
+    ///
+    func testDeleteUserUnauthenticatedUser() {
+        let username = "testUser" + UUID().uuidString
+        let signoutExpectation = expectation(description: "Successfully signout for unauthenticated User")
+        signUpAndVerifyUser(username: username)
+        XCTAssertTrue(AWSMobileClient.default().isSignedIn == false, "Expected to return false for isSignedIn")
+        sleep(1)
```
brennanMKE, line 52: Is there a way to replace the delay with a completion handler instead of using `sleep`? What you could do is add an optional closure property which is internal and only set for tests which is called once `signUpAndVerifyUser` is done.

brennanMKE (commented):
added comments

AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClientExtensions.swift
ameter: This cannot be nil if there is a valid CongnitoUserPool configuration.  If there's not a valid configuration, an error will be thrown during configuration, so we will never get to this point.  With that said, I agree that it's better to avoid force unwrapping since the effects are not obvious when reading the code.  I changed this to handle it explicitly.  As you said, this is now handled differently than the rest of this extension, but I feel this is preferable, and this could serve as an example to update the other methods going forward.

AWSAuthSDK/Tests/AWSMobileClientTests/AWSMobileClientDeleteUserTests.swift
ameter: Agreed, in my opinion this should definitely be handled via a completion callback instead of a sleep.  With that said, the function this is using is shared and handled via a sleep throughout the AWSMobileClient integration tests, so if we change it, we should probably change it throughout, which is beyond the scope of this PR.  We should consider a follow-on PR to change the tests to callbacks throughout AWSMobileClient.

AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClientExtensions.swift
brennanMKE: 👍

AWSAuthSDK/Tests/AWSMobileClientTests/AWSMobileClientDeleteUserTests.swift
brennanMKE: 👍

brennanMKE (approved):
LGTM

dmiluski:
Was this also going to be implemented in the Android SDK? Or specific to iOS given the Apple requirement.

ameter:
> Was this also going to be implemented in the Android SDK? Or specific to iOS given the Apple requirement.

Apple's account deletion mandate is driving this requirement, but it will be added across platforms, including Android.

Issue #3949, opened 2021-12-28
Got some crash reports on Crashlytics from AWSMobileClient 2.26.6. 


koogawa:
**Describe the bug**

The app is on poduction and we are getting these crashes frequently from AWSMobileClient 2.26.6. 

![image](https://user-images.githubusercontent.com/893643/147545762-9cbf0249-e082-4591-a12e-a0c6065f1885.png)

**To Reproduce**
1. Launch app

**Observed Behavior**

**Expected Behavior**

**Stack Trace**

<img width="1291" alt="スクリーンショット 2021-12-23 8 44 52" src="https://user-images.githubusercontent.com/893643/147546055-df129c85-3ead-4689-8fc7-bca20b2a0050.png">

**Code Snippet**
Please provide a snippet of the code causing the issue or how you are using the service that has the bug.

```
        do {
            try Amplify.add(plugin: AWSCognitoAuthPlugin())
            try Amplify.add(plugin: AWSS3StoragePlugin())
            try Amplify.configure()
        } catch {
            Logger.default().logError(error)
        }
```

**Unique Configuration**
If you are reporting an issue with a unique configuration or where configuration can make a difference in code execution (i.e. Cognito) please provide your configuration. Please make sure to obfuscate sensitive information from the configuration before posting.

**Areas of the SDK you are using (AWSMobileClient, Cognito, Pinpoint, IoT, etc)?**

**Screenshots**

**Environment(please complete the following information):**
 - SDK Version: [e.g. 2.26.6]
 - Dependency Manager: Cocoapods
 - Swift Version : 5
 - Xcode Version: 13

```
- Amplify (1.17.0):
    - Amplify/Default (= 1.17.0)
  - Amplify/Default (1.17.0)
  - AmplifyPlugins/AWSCognitoAuthPlugin (1.17.0):
    - AWSAuthCore (~> 2.26.5)
    - AWSCognitoIdentityProvider (~> 2.26.5)
    - AWSCognitoIdentityProviderASF (~> 2.26.5)
    - AWSCore (~> 2.26.5)
    - AWSMobileClient (~> 2.26.5)
    - AWSPluginsCore (= 1.17.0)
  - AmplifyPlugins/AWSS3StoragePlugin (1.17.0):
    - AWSCore (~> 2.26.5)
    - AWSPluginsCore (= 1.17.0)
    - AWSS3 (~> 2.26.5)
  - AppSyncRealTimeClient (1.6.0):
[...]

PR #3951, opened 2021-12-28
fix(AWSMobileClient): crash fix for call to leave DispatchGroup (#3949)

brennanMKE:
*Issue #, if available:*

#3949

*Description of changes:*

Fix for crash bug related to recent async update.

*Check points:*

- [x] All unit tests pass
- [x] Updated CHANGELOG.md

### DIFF
--- a/AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClient.swift
+++ b/AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClient.swift
@@ -83,8 +83,7 @@ final public class AWSMobileClient: _AWSMobileClient {
         queue.maxConcurrentOperationCount = 1
         return queue
     }()
-    internal let tokenFetchLock = DispatchGroup()
-    
+
     /// Operation Queue to make sure there is only 1 active API call which is fetching/ waiting for AWS Credentials.
     internal let credentialsFetchOperationQueue: OperationQueue = {
         var queue = OperationQueue()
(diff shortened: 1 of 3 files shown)


AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClient.swift
```diff
@@ -621,9 +620,6 @@ final public class AWSMobileClient: _AWSMobileClient {
         self.performHostedUISuccessfulSignInTasks(disableFederation: hostedUIOptions.disableFederation, session: session, federationToken: federationToken!, federationProviderIdentifier: federationProviderIdentifier, signInInfo: &signInInfo)
         self.mobileClientStatusChanged(userState: .signedIn, additionalInfo: signInInfo)
         completionHandler(.signedIn, nil)
-        if self.pendingGetTokensCompletion != nil {
-            self.tokenFetchLock.leave()
-        }
```
lawmicha, line 626: is there a change in behavior here? we used to enter the lock on getTokens, then check that the refresh token expired, and then set pendingGetTokensCompletion. developers listen to the state for `signedOutUserPoolsTokenInvalid` and will prompt the user to sign back in. When the user signs back in, this method is called. If there is a `pendingGetTokensCompletion` what should we do that is the equivalent to before?  should it be
```
if self.pendingGetTokensCompletion != nil {
            self.pendingGetTokensCompletion("??", nil) // so that the `done()` is called in AWSMobileClientExtensions.swift line 856

```
Amplify's auth plugin will set up the listeners and call releaseSignIn immediately, which will call pendingGetTokensCompletion with an error and set it to `nil`.

lawmicha: oh, it's calling it in the next line

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 - **AWSMobileClient**
   - Reset completion sources in AWSMobileClient (See [PR #3732](https://github.com/aws-amplify/aws-sdk-ios/pull/3732))
+  - Crash fix in AWSMobileClient for call to leave DispatchGroup (See [PR #3732](https://github.com/aws-amplify/aws-sdk-ios/pull/3949))
```
royjit, line 11:
```suggestion
  - Crash fix in AWSMobileClient for call to leave DispatchGroup (See [PR #3951](https://github.com/aws-amplify/aws-sdk-ios/pull/3951))
```

royjit (approved):
LGTM

lawmicha (dismissed):

lawmicha (approved):

Issue #4104, opened 2022-04-26
Change Pinpoint to allow for push events for received and opened


brennanMKE:
The `UNUserNotificationCenterDelegate` supports `willPresent` and `didReceive` methods and `willPresent` is only called when the app is in the foreground. It should report the `received` event for pushes. The `didReceive` method should report the `opened` event regardless of the app being in the foreground or not.

The distinction is not currently handled in the Pinpoint SDK.

harsh62:
The issue has been released.

PR #4105, opened 2022-04-26
fix(pinpoint): adds push event to indicate if a push is received or opened

brennanMKE:
*Issue #, if available:*

#4104

*Description of changes:*

Defines `AWSPinpointPushEvent` and uses it in the intercept methods for the Pinpoint SDK so that the proper events can be reported.

*Check points:*

- [x] All unit tests pass
- [x] Updated CHANGELOG.md

### DIFF
--- a/AWSPinpoint/AWSPinpointNotificationManager.h
+++ b/AWSPinpoint/AWSPinpointNotificationManager.h
@@ -41,6 +41,11 @@ typedef NS_ENUM(NSInteger, AWSPinpointPushEventSourceType) {
     AWSPinpointPushEventSourceTypeUnknown
 };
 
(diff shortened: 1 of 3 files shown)


brennanMKE:
> It looks like the change makes `received` and `opened` mutually exclusive in the opt-in behavior, is that correct?
> My understanding of the [documented behavior](https://docs.aws.amazon.com/pinpoint/latest/developerguide/event-streams-data-campaign.html) is that it can be both.

We discussed during standup. See the linked issue (#4104) for an explanation of how this update works. Basically there are 2 delegate functions and `.received` is meant for one and `.opened` is meant for the other. When the app is in the foreground and a notification alert is displayed the `.opened` event is reported. When the user taps the alert the `.opened` event is reported. This is the intended behavior.

Issue #4128, opened 2022-05-06
Large uploads use excessive memory


brennanMKE:
**Describe the bug**

Uploading a large file causes more memory to be needed than is possible which causes runtime exceptions.

See: https://github.com/aws-amplify/amplify-ios/issues/1775

PR #4129, opened 2022-05-06
fix(s3): adds autorelease pool around file creation for multipart uploads #4128

brennanMKE:
Found the memory spike happens just before uploading when partial files are created. It appears to hold onto the memory during the run loop so an `@autoreleasepool` was added to clear the memory in the loop so it does not add up.

*Issue #, if available:*

#4128

*Description of changes:*

*Check points:*

- [ ] All unit tests pass
- [ ] Updated CHANGELOG.md


ameter (dismissed):
LGTM

ameter (approved):

PR #4136, opened 2022-05-12
bump version

harsh62:
release commit


atierian:
What's the reason this is a minor bump rather than patch?

ameter (approved):

royjit:
Why is this a major version change?

ameter:
change log says breaking changes in location model

Issue #4356, opened 2022-10-20
Warnings in Xcode 13 during event recording


frankiesimon:
### Describe the bug

When recording events I see a lot of these:

```
[general] *** -[NSKeyedUnarchiver validateAllowedClass:forKey:] allowed unarchiving safe plist type ''NSNumber' (0x117ef36e8) [/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Library/Developer/CoreSimulator/Profiles/Runtimes/iOS.simruntime/Contents/Resources/RuntimeRoot/System/Library/Frameworks/Foundation.framework]' for key 'NS.objects', even though it was not explicitly included in the client allowed classes set: '{(
    "'NSMutableString' (0x117eefae8) [/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Library/Developer/CoreSimulator/Profiles/Runtimes/iOS.simruntime/Contents/Resources/RuntimeRoot/System/Library/Frameworks/Foundation.framework]",
    "'NSDictionary' (0x11bdcba58) [/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Library/Developer/CoreSimulator/Profiles/Runtimes/iOS.simruntime/Contents/Resources/RuntimeRoot/System/Library/Frameworks/CoreFoundation.framework]"
)}'. This will be disallowed in the future.

[general] *** -[NSKeyedUnarchiver validateAllowedClass:forKey:] allowed unarchiving safe plist type ''NSString' (0x117ef2950) [/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Library/Developer/CoreSimulator/Profiles/Runtimes/iOS.simruntime/Contents/Resources/RuntimeRoot/System/Library/Frameworks/Foundation.framework]' for key 'NS.keys', even though it was not explicitly included in the client allowed classes set: '{(
    "'NSMutableString' (0x117eefae8) [/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Library/Developer/CoreSimulator/Profiles/Runtimes/iOS.simruntime/Contents/Resources/RuntimeRoot/System/Library/Frameworks/Foundation.framework]",
[...]

PR #4357, opened 2022-10-20
fix(AWSPinpoint): Connectivity errors should be retried indefinitely

ruisebas:
*Issue #, if available:*
- https://github.com/aws-amplify/aws-sdk-ios/issues/4356
- https://github.com/aws-amplify/amplify-swift/issues/2401
- https://github.com/aws-amplify/amplify-swift/issues/2471

*Description of changes:*
  - Checking for attributes and metrics limits independently from each other, as Pinpoint does allow to have up to 40 metrics and 40 attributes.
  - Events that failed to submit due to connectivity errors are now considered to be retryable indefinitely, i.e. their `retryCount` won't be increased. 
  - Removed warnings with `NSKeyedUnarchiver` due to missing allowed classes.

*Check points:*

- [X] All unit tests pass
- [ ] ~All integration tests pass~
- [X] Updated CHANGELOG.md

### DIFF
--- a/AWSPinpoint/AWSPinpointEvent.m
+++ b/AWSPinpoint/AWSPinpointEvent.m
@@ -31,7 +31,6 @@ NSString *const AWSPinpointEventErrorDomain = @"com.amazonaws.AWSPinpointEventEr
 @property (nonatomic, readwrite) AWSPinpointSession *session;
 @property (nonatomic, readwrite) NSMutableDictionary *attributes;
 @property (nonatomic, readwrite) NSMutableDictionary *metrics;
-@property (atomic, readonly) int currentNumOfAttributesAndMetrics;
 
 @end
 
@@ -67,10 +66,6 @@ NSString *const AWSPinpointEventErrorDomain = @"com.amazonaws.AWSPinpointEventEr
         _metrics = metrics;
     }
     return self;
-}
-
-- (int) currentNumOfAttributesAndMetrics {
-    return (int)self.attributes.count + (int)self.metrics.count;
 }
 
 - (NSMutableDictionary*) attributes {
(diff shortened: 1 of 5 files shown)


phantumcode (dismissed):

frankiesimon:
@ruisebas Sorry for the flood of comments. Just making sure it's clear. There IS a quota of also testing the combined number of metrics and attributes as I wrote in my last comment it's confirmed by AWS support. So you should probably not remove the check as the PR currently does but rather adjust it to what PinPoint would actually check. (And again, it's best if you reproduce on your end to confirm the behaviors when surpassing the separate/mutual quotas).

I can confirm the scenarios I tested but I didn't try to send 40 attributes and 40 metrics. The issue discussion has separate parts based on what I know vs what I was guessing. Last comment mostly relevant for this PR - adjusting the condition to be <=.

https://github.com/aws-amplify/amplify-swift/issues/2471

AWSPinpoint/AWSPinpointEvent.m
```diff
@@ -115,12 +110,12 @@ - (void)addAttribute:(NSString *)theValue forKey:(NSString *)theKey {
     if(!theKey) return;
     
     @synchronized(self.attributes) {
-        if(self.currentNumOfAttributesAndMetrics < MAX_NUM_OF_METRICS_AND_ATTRIBUTES) {
+        if(self.attributes.count < MAX_NUM_OF_METRICS_AND_ATTRIBUTES) {
```
frankiesimon, line 113: I think the correct quota check would be the same as before BUT the comparison should be <=. Also the same warning applies.

AWSPinpoint/AWSPinpointEvent.m
```diff
@@ -160,11 +155,11 @@ - (void)addMetric:(NSNumber *)theValue forKey:(NSString *)theKey {
     }
     
     @synchronized(self.metrics) {
-        if(self.currentNumOfAttributesAndMetrics < MAX_NUM_OF_METRICS_AND_ATTRIBUTES) {
+        if(self.metrics.count < MAX_NUM_OF_METRICS_AND_ATTRIBUTES) {
```
frankiesimon, line 158: I think the correct quota check would be the same as before BUT the comparison should be <=. Also the same warning applies.

AWSPinpoint/AWSPinpointEvent.m
frankiesimon: Thanks for the clarification I understand now why the condition is < and not <=.

phantumcode (approved):

Issue #4486, opened 2022-12-30
crash in AWSMobileClient.getTokens


paulz:
**Describe the bug**

A clear and concise description of what the bug is. Please be as detailed as possible and help us help you.

Crashlytics reported crash on iOS device

**Stack Trace**
```
Fatal Exception: NSInternalInconsistencyException
*** -[NSHashTable NSHashTable { [0] <AWSMobileClientXCF.FetchUserPoolTokensOperation: 0x103c09570> } ] count underflow

Fatal Exception: NSInternalInconsistencyException
0  CoreFoundation                 0x9e88 __exceptionPreprocess
1  libobjc.A.dylib                0x178d8 objc_exception_throw
2  Foundation                     0x555d60 -[NSConcreteHashTable rehash]
3  Foundation                     0xc680 hashProbe
4  Foundation                     0x129f4 -[NSConcreteHashTable addObject:]
5  AWSMobileClientXCF             0x3ae68 specialized AWSMobileClient.getTokens(_:)
6  AWSMobileClientXCF             0x20538 AWSMobileClient.logins() (<compiler-generated>)
7  AWSMobileClientXCF             0x20ce4 @objc AWSMobileClient.logins() (<compiler-generated>)
8  AWSCore                        0x3192c -[AWSCognitoCredentialsProviderHelper getIdentityId] + 301 (AWSIdentityProvider.m:301)
9  AWSCore                        0x31558 -[AWSCognitoCredentialsProviderHelper logins] + 274 (AWSIdentityProvider.m:274)
10 AWSCore                        0x55fb4 -[AWSCognitoCredentialsProvider credentialsWithCancellationToken:] + 630 (AWSCredentialsProvider.m:630)
11 AWSMobileClientXCF             0x252a8 closure #1 in AWSMobileClient.getAWSCredentials(_:) + 30 (AWSMobileClient+Credentials.swift:30)
12 AWSMobileClientXCF             0x1c0a4 @objc AWSAsyncBlockOperation.main() (<compiler-generated>)
13 AWSMobileClientXCF             0x489c0 @objc AWSAsyncOperation.start() (<compiler-generated>)
14 Foundation                     0x3c50c __NSOPERATIONQUEUE_IS_STARTING_AN_OPERATION__
15 Foundation                     0x41c20 __NSOQSchedule_f
16 libdispatch.dylib              0x13114 _dispatch_block_async_invoke2
[...]

PR #4563, opened 2023-02-17
fix(AWSMobileClient): Fixing a crash when getTokens is called while tokenOperations is being iterated.

ruisebas:
*Issue #, if available:*
https://github.com/aws-amplify/aws-sdk-ios/issues/4486

*Description of changes:*

Calling `AWSMobileClient.getTokens(_:)` at the same time a user is signing out/in could result in a crash, since the `tokenOperations` is Iterated when the auth state changes. 

This happens because `NSHashTable`'s operations are not thread safe, so I'm creating a `tokenOperationsQueue` queue that we can use to read/write from the collection. I'm setting its `.concurrent` attribute so reads can be done concurrently, but setting the `.barrier` flag when actually inserting values.

I've also wrapped this implementation in a new class called `WeakHashTable`, so it reduces the change of misuse in the future (e.g. if we add another read/write from the collection in a different place without using the queue)
 
*Check points:*

- [X] All unit tests pass
- [X] Updated CHANGELOG.md


AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClient.swift
```diff
@@ -88,6 +88,8 @@ final public class AWSMobileClient: _AWSMobileClient {
     var userPassword: String? = nil
 
     var tokenOperations:NSHashTable<FetchUserPoolTokensOperation> = NSHashTable.weakObjects()
+    let tokenOperationsQueue: DispatchQueue = DispatchQueue(label:"AWSMobileClient.tokenOperationsQueue",
```
royjit, line 91: There is similarly named queue `tokenFetchOperationQueue` and can be confusing. Should we rename this `queueToAccessTokenOperations` or some other name?

royjit (commented):
We need to test these scenarios in AWSMobileClient and Amplify v1.
- Successful fetching of tokens
- Successful fetching of tokens after it is expired, AWSMobileClient should refresh the tokens automatically if refresh token is valid
- Receive token expired notification in case refresh token is invalid

AWSAuthSDK/Sources/AWSMobileClient/AWSMobileClient.swift
ruisebas: Agreed, it becomes confusing.
So I instead decided to encapsulate this behaviour inside a new `WeakHashTable` class

royjit (approved):

Issue #4721, opened 2023-05-06
EXC_BAD_ACCESS AWSPinpointEventRecorder crash when device is locked


mmathewsTableau:
**Describe the bug**
Our app crashes when AWSPinpointEventRecorder tries to update the sqlite database while the device is locked. The crash occurs on this [line](https://github.com/aws-amplify/aws-sdk-ios/blob/2.30.4/AWSPinpoint/AWSPinpointEventRecorder.m#L1120).

We end up in this block because `db executeUpdate` fails. This makes sense because our app is configured with NSFileProtectionComplete, which means the app does not have access to the underlying filesystem when the device is locked. What I have observed is the first call to `db.lastError` in this code block is successful but the second call (the line referenced above) fails with EXC_BAD_ACCESS. I believe the action of getting the last error may close the database(?). 

I have a local fix that gets rid of this crash from happening in AWSPinpointEventRecorder. I plan to make a PR with this fix. However, I will say that I've occasionally seen EXC_BAD_ACCESS errors happening inside AWSFMDatabase using the same repro steps (it usually requires many more repro attempts to get this)

**To Reproduce**

Prereq (use an app that has NSFileProtectionComplete). I am also using a real device (not 100% sure if that's required)
1. Attach debugger to the app
2. Put breakpoint [here](https://github.com/aws-amplify/aws-sdk-ios/blob/2.30.4/AWSPinpoint/AWSPinpointEventRecorder.m#L1114). Do something to make sure the app hits it. 
3. Once the app hits the breakpoint, lock the device
4. Wait at least 5 seconds.. maybe more..
5. Step over the rest of the logic. `result` should be `no` and once you enter the  [error logging](https://github.com/aws-amplify/aws-sdk-ios/blob/2.30.4/AWSPinpoint/AWSPinpointEventRecorder.m#L1119) code block , the `*error = db.lastError;` line should cause EXC_BAD_ACCESS and a crash. 

**Observed Behavior**

App crashes

**Expected Behavior**

App doesn't crash. I think the app not being able to update the database is expected, however, I would expect the library to handle this scenario more gracefully (ie not crash) 

**Stack Trace**

Pods/AWSPinpoint/AWSPinpoint/AWSPinpointEventRecorder.m:1120:36 __60-[AWSPinpointEventRecorder putEvents:error:endpointProfile:]_block_invoke_3.234
[...]

PR #4728, opened 2023-05-08
Trying to fix EXC_BAD_ACCESS in AWSPinpointEventRecorder

mmathewsTableau:
*Issue #, if available: 4721

https://github.com/aws-amplify/aws-sdk-ios/issues/4721

*Description of changes:

You can read the bug for more details but TLDR, if the app is locked and the app is configured to protect data (ie NSFileProtectionComplete), then updating the database fails (as expected). Calling db.lastError fails when called twice consecutively.

Although I couldn't fully understand why we cannot call db.lastError twice in a row, I was able to make changes to my local pod instance to prevent the error from happening by making adjustments to the code so we are only calling db.lastError once. 

Initially, I didn't have logic to copy the error but that somehow still ended up with EXC_BAD_ACCESS occurring so that is why I added the logic to copy.

Apologies if there's better way to fix, objective c is not a language I work with very much. 

*Check points:*

- [ ] All unit tests pass
- [ ] Updated CHANGELOG.md

### DIFF
--- a/AWSPinpoint/AWSPinpointEventRecorder.m
+++ b/AWSPinpoint/AWSPinpointEventRecorder.m
@@ -1063,8 +1063,8 @@ NSString *const FAILURE_REASON = @"NSLocalizedFailureReason";
                                                               @"id" : eventID
                                                               }];
                             if (!result) {
-                                AWSDDLogError(@"SQLite error. [%@]", db.lastError);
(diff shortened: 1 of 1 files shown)


ruisebas (approved):
Thanks @mmathewsTableau !

PR #4906, opened 2023-08-16
test: fix flaky KMS test cases

5d:
*Issue #, if available:*

*Description of changes:*

- update the test case to use expecations rather than NSLock to wait for async result. As NSLock is only blocking the execution thread and the AWSTask is dispatched on the global queue, which may not really blocking the test executing thread.

*Check points:*

- [x] All unit tests pass
- [ ] Updated CHANGELOG.md


atierian (approved):
LGTM. Just one request - please add a quick explanation in the PR about how this change fixes the flaky tests. It'll be good to have a reference as it looks like this pattern is used extensively throughout other test suites in the SDK.

Issue #1576, opened 2019-05-28
Cache files left when AWSS3TransferUtilityMultiPartUploadTask canceled


SGA-ou-enhaku:
**Describe the bug**
After [AWSS3TransferUtilityMultiPartUploadTask cancel] calling, there are cache files left in /Library/Caches/S3TransferUtility.
If [S3TransferUtility uploadFileUsingMultiPart:...] -> [task cancel] all the time, there would be bad for iPhone' storage.

**To Reproduce**
1. Call [S3TransferUtility uploadFileUsingMultiPart:...]
2. Call [AWSS3TransferUtilityMultiPartUploadTask cancel]
3. Go to (App)/Library/Caches/S3TransferUtility
4. See the left cache files

**Which AWS service(s) are affected?**
S3

**Expected behavior**
Cache files should be deleted.

**Environment(please complete the following information):**
 - SDK Version: [e.g. 2.9.8 and 2.8.2]
 - Dependency Manager: [e.g. Cocoapods]
 - Objective-C

**Device Information (please complete the following information):**
 - Device: [iPhoneXR, Simulator]
 - iOS Version: [e.g. iOS 12.1]

minbi:
Hi @SGA-ou-enhaku ,

Thanks for bringing this to our attention. This would be an agreeable improvement to the current behavior. We will look into prioritizing this item. In the meantime, we are open to pull requests if you are interested.

pablogeek:
I'm also having this. Please how could we remove the cache?

pablogeek:
this code solved my issue, I run this everytime something is correctly uploaded

```
    func cleanCache(){
        if let url = FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask).first {
            let urls = try?  FileManager.default.contentsOfDirectory(at: url, includingPropertiesForKeys: nil, options: .skipsSubdirectoryDescendants)
            urls?.filter { $0.lastPathComponent.contains("S3") }.forEach { try? FileManager.default.removeItem(at: $0) }
        }
    }
```

Guolanlan:
yes,you can call suspend and it will pause,no remove cache

PR #5128, opened 2023-12-15
fix(s3): cleanup file cache after mpu task is cancelled

atierian:
*Issue #, if available:*
- https://github.com/aws-amplify/aws-sdk-ios/issues/1576

*Description of changes:*

If a `AWSS3TransferUtilityMultiPartUploadTask` is cancelled by the caller, there's no use keeping the cached file around.
According to the [File System Programming Guide](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html#//apple_ref/doc/uid/TP40010672-CH2-SW1), the /Library/Caches directory is rarely purged by the system:
> Your app is generally responsible for managing the contents of this directory and for adding and deleting files as needed. 
> ...
> In iOS 5.0 and later, the system may delete the Caches directory on rare occasions when the system is very low on disk space. This will never occur while an app is running. However, be aware that restoring from backup is not necessarily the only condition under which the Caches directory can be erased.

*Check points:*

- [x] All unit tests pass
- [ ] ~Updated CHANGELOG.md~

### DIFF
--- a/AWSS3/AWSS3TransferUtilityTasks.m
+++ b/AWSS3/AWSS3TransferUtilityTasks.m
@@ -153,6 +153,15 @@
     }
 
     [AWSS3TransferUtilityDatabaseHelper deleteTransferRequestFromDB:self.transferID databaseQueue:self.databaseQueue];
+
+    NSString *path = self.cachedFileURL.path;
+    if (path && [[NSFileManager defaultManager] fileExistsAtPath:path]) {
+        NSError *error = nil;
+        [[NSFileManager defaultManager] removeItemAtPath:path error:&error];
+        if (error) {
+            AWSDDLogError(@"Error deleting file[%@]: [%@]", path, error);
+        }
+    }
 }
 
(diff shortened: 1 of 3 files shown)


AWSS3/AWSS3TransferUtilityTasks.m
```diff
@@ -153,6 +153,15 @@ - (void)cancel {
     }
 
     [AWSS3TransferUtilityDatabaseHelper deleteTransferRequestFromDB:self.transferID databaseQueue:self.databaseQueue];
+
+    NSString *path = self.cachedFileURL.path;
```
phantumcode, line 157: can the additional have a unit test?

zamzamfp:
Hi @atierian, I am having issues with the cache files, in my case it is really bad since file sizes are huge.  is there a reason why this branch is not merged?

Issue #1159, opened 2019-01-03
AWSIoTDataManager.getShadow() timeout [race condition with AWSIoTDataManager.register()?]


armandr:
**State your question**

How does one know AWSIoTDataManager.register(withShadow) has **completed the registration**(finished subscribing)?

In my IoS app I'm calling AWSIoTDataManager.register() with QoS:1 and on success immediately calling AWSIoTDataManager.getShadow().  This often, but not always leads to a AWSIoTShadowOperationStatusTypeTimeout.

It seems possible for getShadow() to be called **before** register() task has completed subscribing to shadow topics therefore resulting in a timeout.

However, adding a slight delay (as low as 100ms, but more than 10us) in between the register() and getshadow() makes the result more deterministic and successful (at least on my machine and simulator)

 The BOOL return value of AWSIoTDataManager.register() is based on the input parameters and does not provide an ACK Callback.

How does one determine that the register is successful and complete so shadow get/update operations can begin?

**Which AWS Services are you utilizing?**
AWSIoT, AWSCognito

**Provide code snippets (if applicable)**
 ```
let res = self.iotDataManager.register(withShadow: "DATHING", options: ["shadowOperationTimeoutSeconds": "QoS": 1, ": true],
                        eventCallback: { (string1, optype, status, string2, data) in
                       // handle callback
                    })
                    
DispatchQueue.main.asyncAfter(deadline: .now()+0.5 ) {                        
      self.iotDataManager.getShadow("HUMTHING")
}

```
**Environment(please complete the following information):**
 - SDK Version: latest
 - Dependency Manager: Cocoapods
 - Swift Version : 4.2

**Device Information (please complete the following information):**
 - Device: Simulator
 - iOS Version: iOS latest
 - Specific to simulators:

minbi:
Hi @armandr ,

I took a look at the code and it looks like we don't currently support passing this information back. The subscribe code returns void [on this line](https://github.com/aws-amplify/aws-sdk-ios/blob/master/AWSIoT/Internal/AWSIoTMQTTClient.m#L775). I'm marking this as a feature request and also assigning @cbommas who might have more/better information.

AndreaPetrelli:
I'm in the same situation, there is any news?

dholdren:
I'm also running into this issue. seems like a common use-case, register for updates and immediately get the device shadow

kshrikant:
[...]

PR #5192, opened 2024-02-09
fix(IoT): Adding completion callbacks for `registerWithShadow` and `unregisterFromShadow` methods

ruisebas:
**Issue #, if available:**
- https://github.com/aws-amplify/aws-sdk-ios/issues/1159

**Description of changes:**

This PR adds the following methods to `AWSIoTDataManager`:
  - `registerWithShadow:options:eventCallback:completionCallback:`
  - `unregisterFromShadow:completionCallback:`

They are basically just overloading the existing ones with the addition of the `completionCallback` argument, which is invoked when the subscription to all topics is done.

The reasoning behind this is to provide a way for consumers to know when the subscription is actually done and they can get the shadow (see the referenced issue for more information).

*Check points:*

- [ ] All unit tests pass
- [ ] Updated CHANGELOG.md

### DIFF
--- a/AWSIoT/AWSIoTDataManager.m
+++ b/AWSIoT/AWSIoTDataManager.m
@@ -912,22 +912,39 @@ static NSString * const AWSIoTShadowOperationStatusTypeStrings[] = {
  */
 - (BOOL)handleSubscriptionsForShadow:(NSString *)name
                             callback:(AWSIoTMQTTExtendedNewMessageBlock)callback {
+    return [self handleSubscriptionsForShadow:name callback:callback completionHandler:nil];
+}
+
+- (BOOL)handleSubscriptionsForShadow:(NSString *)name
+                            callback:(AWSIoTMQTTExtendedNewMessageBlock)callback
+                  completionHandler:(void(^)(void))completionHandler {
     BOOL rc = NO;
     AWSIoTDataShadow *shadow = (AWSIoTDataShadow *)[self.shadows objectForKey:name];
     
     if (shadow != nil) {
+        __block NSUInteger remainingTopics = [shadow.topics count];
+        AWSIoTMQTTAckBlock ackCallback = NULL;
+        if (completionHandler) {
+            ackCallback = ^void {
+                remainingTopics--;
+                if (remainingTopics == 0) {
+                    completionHandler();
+                }
+            };
+        }
+
         for (int i = 0; i < [shadow.topics count]; i++) {
             if (callback != nil) {
                 if (shadow.enableDebugging == YES) {
                     AWSDDLogInfo(@"subscribing on %@", (NSString *)shadow.topics[i]);
                 }
-                [self subscribeToTopic:shadow.topics[i] QoS:shadow.qos extendedCallback:callback];
+                [self subscribeToTopic:shadow.topics[i] QoS:shadow.qos extendedCallback:callback ackCallback:ackCallback];
             }
             else {
                 if (shadow.enableDebugging == YES) {
                     AWSDDLogInfo(@"unsubscribing from %@", (NSString *)shadow.topics[i]);
                 }
-                [self unsubscribeTopic:shadow.topics[i]];
+                [self unsubscribeTopic:shadow.topics[i] ackCallback:ackCallback];
             }
         }
         rc = YES;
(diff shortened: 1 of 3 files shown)


AWSIoT/AWSIoTDataManager.m
```diff
@@ -1208,6 +1225,20 @@ - (BOOL) operationWithShadow:(NSString *)name
...
+                    options:(NSDictionary *)options
+              eventCallback:(void(^)(NSString *name, AWSIoTShadowOperationType operation, AWSIoTShadowOperationStatusType status, NSString *clientToken, NSData *payload))callback
+         completionCallback:(void(^)(void))completionCallback {
+    return [self registerWithShadow:name options:options eventCallback:callback completionHandler:completionCallback];
+}
+
+- (BOOL) registerWithShadow:(NSString *)name
+                    options:(NSDictionary *)options
+              eventCallback:(void(^)(NSString *name, AWSIoTShadowOperationType operation, AWSIoTShadowOperationStatusType status, NSString *clientToken, NSData *payload))callback
+          completionHandler:(nullable void(^)(void))completionHandler {
```
ruisebas, line 1241: Having the `nullable` annotation in the interface method resulted in weird stuff when the method was translated to Swift and autocompletion was attempted.

While not an error itself, I rather just do this to avoid confusing consumers, i.e. having the `completionCallback` declared as `nonnull` and having an implementation-only method that can take `nil`.

The same applies for the `unregisterFromShadow` method

atierian:
Noting API BR approval for API addition -- code review to be completed independently.

tylerjroach:
Approving as API BR

AWSIoT/AWSIoTDataManager.m
```diff
@@ -912,22 +912,39 @@ - (void)createSubscriptionsForShadow:(AWSIoTDataShadow *)shadow
...
+}
+
+- (BOOL)handleSubscriptionsForShadow:(NSString *)name
+                            callback:(AWSIoTMQTTExtendedNewMessageBlock)callback
+                  completionHandler:(void(^)(void))completionHandler {
     BOOL rc = NO;
     AWSIoTDataShadow *shadow = (AWSIoTDataShadow *)[self.shadows objectForKey:name];
     
     if (shadow != nil) {
+        __block NSUInteger remainingTopics = [shadow.topics count];
```
thisisabhash, line 925: is `__block` used so that the changes to `remainingTopics` are visible inside this block when `ackCallback` is called?

AWSIoT/AWSIoTDataManager.m
```diff
@@ -912,22 +912,39 @@ - (void)createSubscriptionsForShadow:(AWSIoTDataShadow *)shadow
  */
 - (BOOL)handleSubscriptionsForShadow:(NSString *)name
                             callback:(AWSIoTMQTTExtendedNewMessageBlock)callback {
+    return [self handleSubscriptionsForShadow:name callback:callback completionHandler:nil];
+}
+
+- (BOOL)handleSubscriptionsForShadow:(NSString *)name
```
thisisabhash, line 918: nit: can we have a doc comment like the above function?

AWSIoT/AWSIoTDataManager.m
thisisabhash: nit: can we rename the parameter to something with *callback to maintain consistency? something like `completionCallbackNullable`?

AWSIoT/AWSIoTDataManager.m
ruisebas: will add!

AWSIoT/AWSIoTDataManager.m
ruisebas: Exactly :) Just Objective-C stuff

AWSIoT/AWSIoTDataManager.m
ruisebas: Umm I think that's a very long and over-descriptive name. The `-Nullable` suffix is redundant because the parameter is already marked as `nullable`.

Since this is basically a private method, and the public API does name the parameter just `completionCallback`, I'd prefer to keep the `completionHandler` name even if it breaks consistency 😅

thisisabhash (approved):

